Issue #1, opened 2020-05-11
Question about reliance on S3


wlandau:
In the [requirements](https://github.com/Rconsortium/oop-wg/wiki/Requirements) document, the compatibility section states

> Ideally the new system will be an extension of S3, because S3 is already at the bottom of the stack and many of the other systems have some compatibility with S3.

And the syntax section states

> Ideally the entire API could be free of side effects and odd conventions, like making the . character significant in method names. Direct manipulation of class and generic objects should enable this.

In what way would we build on S3 without S3 syntax? Do we build a layer of sugar on top and ignore the dot-based dispatch underneath? Maybe hide the S3 mechanics inside the class contract?

lionel-:
The S3 dot syntax is based on lexical dispatch (finding methods in the current environment), which doesn't work well in general. R has been moving away from lexical dispatch for several versions now. Methods should now be registered with NAMESPACE directives or with `.S3method()`.

The API can't really be free of side effects since registration into a generic necessarily has global effect (all users of the generic will see the change). @hadley was wondering about using the replacement/assignment syntax as a cue of the side effect:

```r
method(generic, class) <- function(...) {}
```

I find this quite idiomatic.

ltierney:
I would not be a good idea to use the name `setMethod` with the semantics you describe as the name suggest mutating a binding of some sort.

I'll probably find myself saying this a lot in these discussions: These sorts of issues have been addressed both in CLOS (Common Lisp Object System) and in Dylan. These are two of very few generic-function-based OOP frameworks and definitely worth looking at .

lawremi:
Thanks for the reminder about CLOS and Dylan. They clearly inspired the methods package. For example, `(defgeneric)` in CLOS maps to `setGeneric()`. It's great that @hadley and @lionel- are interested in making the side effects more explicit.

wlandau:
Should topics in this thread go in the spec? Do we need explicit policies around lexical dispatch or side effects? I certainly learned from the discussion, but I wonder how far into implementation territory the spec should go.

lawremi:
We could state that we will develop a new syntax that will make method registration more explicit, compared to the implicit registration by symbol name from S3.

wlandau:
Closing because method registration is explicit in the [design](https://github.com/RConsortium/OOP-WG/blob/master/spec/design.md).

PR #11, opened 2020-06-14
Mention explicit method registration in the spec

wlandau:
Implements @lawremi's suggestion from https://github.com/RConsortium/OOP-WG/issues/1#issuecomment-643441572. I feel this wraps up #1 if we want to keep the spec as general as it is now. (Not a strong opinion though, confirmation from others would be nice.)

Also cc @lionel-, I could not add you as a reviewer.


ltierney (approved):

hadley:
Merging since all reviews complete.

Issue #7, opened 2020-05-11
Interface evolution?


hadley:
Should we consider the challenge of changing the interface of a generic (i.e. adding or removing arguments, or moving a generic between packages). These are all things that have come up for tidyverse packages, and I don't feel like we have a good solution for them.

Similarly, what about adding attributes to a class? What if, unbeknownst to you, a subclass is already using the attribute you add?

lawremi:
Evolution of generic formals hinges on whether every method needs to have the same formals as its generic (with the "..." affording some flexibility). We should probably immediately state that as a requirement, because the alternative (like Julia, or like overloading in other frameworks) would be really complicated and confusing to users. In Bioconductor, we tend to make our generics extremely... generic... like `function(x, ...)` to maximize flexibility, but of course, that strongly limits how much a client can assume about a generic. So no great solutions here either. Moving generics is usually easier to handle, especially if moving up the package hierarchy, and downstream packages use `import()` instead of `importFrom()` (which is often not a great idea in many cases).

Adding attributes to a class is also problematic from the serialization perspective (there have been proposals around supporting deserialization hooks). Field name clashes are infrequent in my experience, but I guess they could be handled by (virtual) grouping fields by class. This type of conflict is called "field hiding" in Java and is handed by qualifying with the `super` keyword. In C++, field references are qualified by the class name, since there is multiple inheritance.

hadley:
Interface evolution is related to #6; if a mismatch only causes a run-time warning (not an R CMD check problem), it's possible to work around any mismatches within the method itself.

lawremi:
Please clarify what you mean. How would we call the method if the arguments do not match? Drop the new arguments? Expect the method implementation to check every argument with `missing()`?

hadley:
Here's the scenario: I've mistakenly created generic `pkgA::foofy(x, y)` and pkgB has implemented `foofy.character(x, y)`. Now in pkgA 2.0.0, I change the signature of the generic to `foofy(x, y, z, ...)` (i.e. adding an argument, and allowing methods to define their own additional arguments). I email the author of pkgB telling them about the problem. Ideally they'd be able to release a version of pkgB that works with pkgA 1.0.0 and 2.0.0, but if they define `foofy.character(x, y, z, ...)` they'll get an error when R CMD check is run with the released version of pkgA. (A rarer problem is that I've defined `barfy(x, y, ...)` and later decided that `y` is unneeded, and I want to remove it.)

I think there are two things that would help:

[...]

PR #12, opened 2020-07-17
Requirements related to #7

lawremi:
Roughly addresses the concerns raised in #7. Interface evolution continues to be a challenge to nail down.


spec/requirements.md
```diff
@@ -43,3 +47,10 @@ Ideally the entire API could be free of side effects and odd conventions, like m
...
 
+## Formal arguments
+
+The formal arguments of a generic must be present in every method for that generic. This is in contrast to Julia, where methods can have completely different sets of arguments. We favor a fixed set of formal arguments for the sake of consistency and to enable calling code to depend on a minimal set of arguments that are always present. If the generic formal argument list includes `...`, then methods can add their own arguments. The extra arguments are useful for controlling specialized behaviors, as long as the calling code can assume that calling the generic will always dispatch to a method that supports them. In accordance with the LSP, we could explore enforcing that a method only adds arguments to those of a method dispatching on a parent class. This is easiest to conceptualize and would be most useful in the single dispatch case, but we should also be able to develop a set of rules for nested multiple dispatch.
```
mmaechler, line 52: "LSP" = ? (spell out in paren).

Not quite clear what in ", as long as the calling code can assume that calling the generic will always dispatch to a method that supports them."    'support' really means. If I think of S3 examples, "support" would rather mean "tolerate", i.e., silently disregard in case the extra arguments used in the call are not part of that method's arguments .. for which then, that method (and many others?) may need to also get the `...`  as Hadley mentioned somewhere earlier in the #7

ltierney (approved):

hadley:
I resolved the merge conflict and squash.

PR #15, opened 2020-07-20
Class namespacing

### LINKED ISSUE #4: Class namespacing
Should we explicitly include the class namespacing problem? i.e. with S3, both pkgA and pkgB can define a class `foo`, and there's no distinction at the dispatch level. In [adv-r](https://adv-r.hadley.nz/s3.html#s3-classes) I recommend using the package name as a class prefix to avoid this problem).

hadley:
Fixes #4

### DIFF
--- a/spec/requirements.md
+++ b/spec/requirements.md
@@ -19,6 +19,8 @@ Ideally the new system will be an extension of S3, because S3 is already at the 
 
 It is important for classes to be defined extrinsically from instances, because it makes the data contract more obvious to both developers (reading the code) and users (interacting with objects). S4 represents classes as proper objects; however, typically developers will only refer to the classes by name (string literal) when interacting with the S4 API. Interacting directly with objects instead would likely simplify the API (syntax) and its implementation.
 
+Class objects need to be namespaced so that it's method dispatch still occurs correctly even when multiple packages implement a class with the same (string) name.
+
 ## Generics as extended function objects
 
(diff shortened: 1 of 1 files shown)


spec/requirements.md
```diff
@@ -19,6 +19,8 @@ Ideally the new system will be an extension of S3, because S3 is already at the
 
 It is important for classes to be defined extrinsically from instances, because it makes the data contract more obvious to both developers (reading the code) and users (interacting with objects). S4 represents classes as proper objects; however, typically developers will only refer to the classes by name (string literal) when interacting with the S4 API. Interacting directly with objects instead would likely simplify the API (syntax) and its implementation.
 
+Class objects need to be namespaced so that it's method dispatch still occurs correctly even when multiple packages implement a class with the same (string) name.
```
wlandau, line 22: Minor grammatical correction: "A class object needs to be namespaced so that its method dispatch still occurs correctly even when multiple packages implement a class with the same (string) name."

ltierney:
I'm no sure this is an option if we want S3 compatibility. Also, making sure this would work in the context of serialization is likely to be tricky. I'm not sure there is a viable alternative to something like a `foo_bar` naming scheme for class `foo` in package `bar`.

hadley:
Ah good point. Let me close, and we can consider more systematically as we think more about implementation.

PR #21, opened 2020-09-19
Clarify basic interface up front

hadley:
I've moved the definition of the basic interface into the first paragraph so you get some idea of what the key terms are and what the basic syntax looks like. I was reading in between the lines a bit so let me know if I got it wrong.

(Although looking at it now I think `properties()` and `methods()` would just call `classObject()` on their input, so maybe it would be better to remove that call all together)

### DIFF
--- a/spec/design.md
+++ b/spec/design.md
@@ -2,9 +2,11 @@
 ## Object
 
 We define an object in this system as any R object with:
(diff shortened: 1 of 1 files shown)


lawremi (approved):
Thanks, great improvement to clarity, and good catch with the `classObject()` vs. `class(x)`.

PR #27, opened 2020-09-21
Tweaks to the spec

hadley:
Based on the call, plus a few things I spotted while reading it.

### DIFF
--- a/spec/design.md
+++ b/spec/design.md
@@ -7,6 +7,7 @@ We define an object in this system as any R object with:
 - A direct reference to the **class object**, retrieved with `classObject()`.
 - Additional attributes storing class-dependent **properties**, accessible
   with `@`.
+- Classes, generics, and methods are created with `newClass()`, `newGeneric()`, and `newMethod()` respectively, and only have impact when assigned.
 
 ## Class
 
@@ -150,12 +151,20 @@ gain support for nested dispatch.
  - `prop(x, name)`, `prop<-(x, name, value)`: Get and set a property value
  - `x@name` and `x@name <- value`: Shortcut to get and set property values
 
+
 ### Classes
-Calling `defineClass()` defines a new class. Its signature:
+
+Calling `newClass()` creates a new class. Its signature:
+
 ```{R}
-defineClass <- function(name, parent = object, constructor, init = structure,
-                        validity = function(object) { },
-                        properties = list())
+defineClass(
+  name, 
+  parent = object, 
+  constructor, 
+  init = structure,
+  validity = function(object) { },
+  properties = list()
+)
 ```
 
 where:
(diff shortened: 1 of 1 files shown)


spec/design.md
```diff
@@ -174,22 +180,32 @@ The return value is `constructor`, except it is an instance of class
...
+  properties = c(start = "numeric", end = "numeric")
+)
+
+Range(start = 1, end = 10)
 ```
 
 ### Properties
 
-Calling `defineProperty()` define a new property. It has the
+Calling `newProperty()` define a new property. It has the
```
hadley, line 204: I think this needs fleshing out — esp whether the name belongs to the property of the property list. Is it:

```R
defineClass("Range", ..., properties = list(newProperty("start", "numeric"))
# OR 
defineClass("Range", ..., properties = list(start = newProperty("numeric"))
```

spec/design.md
```diff
@@ -202,19 +218,22 @@ where:
...
+Calling `newGeneric()` defines a new generic. It has the signature:
+
 ```{R}
-function(name, FUN, signature) { }
+newGeneric(name, FUN, signature)
 ```
+
 The `signature` would default to all of the arguments in
-`formals(FUN)`. The body of `FUN` would resemble S3 and S4
+`formals(FUN)[1]`. The body of `FUN` would resemble S3 and S4
```
hadley, line 228: I'd suggest we use single dispatch by default

spec/design.md
```diff
@@ -202,19 +218,22 @@ where:
...
 generics. It might just call `UseMethod()`.
 
 ### Methods
 
-Calling `defineMethod()` defines a new method, with the signature:
+Calling `newMethod()` defines a new method, with the signature:
+
 ```{R}
-function(generic, signature, FUN) { }
+newMethod(generic, FUN, signature)
```
hadley, line 236: Moved signature to end for consistency with `newGeneric`. Is it inconsistent that this object doesn't contain the class name?

lawremi: I see the need for consistency. This may just be my S4 lens, but it's been useful having the signature argument before the implementation, since it's a lot easier to see. Then again, people may just use the syntactic sugar instead.

lawremi: Also, the two `signature` arguments have different meanings. For generics, it is a list of the argument names on which to dispatch. For methods, it is a vector with a class name for every argument in the generic signature. Maybe we need different names? Like `newGeneric()` could have `signatureArgs` instead of `signature`.

spec/design.md
lawremi: Makes sense. Would you please simplify this to "The `signature` would default to names(formals(FUN))[1L]"?

spec/design.md
hadley: Oh true. I was also wondering about `newGeneric()` — if we have the name of the generic, then we could actually generate the body of the function, but then what would `fun` be?

hadley: Maybe the `newMethod()` argument name could contain `class`?

lawremi: The only reason I put `FUN` there was to get the formals. Seems more natural to pass a function than a call to `alist()`. I agree the body is redundant, just like S4.

lawremi: The `class` suggestion seems reasonable. Please commit a proposal.

spec/design.md
```diff
@@ -174,22 +180,32 @@ The return value is `constructor`, except it is an instance of class
 `Class`, which defines properties that describe the rest of the class.
 
 Example:
+
 ```{R}
-Child <- defineClass("Range", Vector, function(start, end) {
+Range <- defineClass("Range", 
+  Vector, 
+  function(start, end) {
     stopifnot(is.numeric(start), is.numeric(end), end >= start)
```
wlandau-lilly, line 188: Minor detail: to me, `end >= start` may blur the line between constructor and validator a bit.

lawremi: Some redundancy is good for defensive programming and more direct error messages.

spec/design.md
hadley: I'm going to turn it back to `signature` and merge this PR. Then we can discuss the interface in more detail in #29.

spec/design.md
hadley: And I think the line between constructor and validator is fundamentally blurred 😄

PR #35, opened 2020-10-05
Update method definition

### LINKED ISSUE #29: Where do the arguments for the method specification live?
Say we want to define a `mean()` method for class `A`.  Assuming a hypothetical `method<-` function that we use to modify the global methods table, I think there are three ways we could supply the data: 

```R
method(mean, A) <- newMethod(function(x) 10)
# which could be shortened to 
method(mean, A) <- function(x) 10

method(mean) <- newMethod("A", function(x) 10)

methods() <- newMethod("mean", "A", function(x) 10)
```

(We could also do `methods(A) <- newMethod("mean", function(x) 10)` but I think that encourages the wrong mental model of the OO system)

I'm kind of partial to the first, because it means that we wouldn't need to have a user facing method constructor. But either way, I think we need to consider the design of the method/class/generic constructors in tandem with the assignment function that they use to have an impact on the state of the world.

hadley:
Fixes #29

### DIFF
--- a/spec/design.md
+++ b/spec/design.md
@@ -265,7 +265,7 @@ method(generic, signature) <- function(x, ...) {}
 
 `method<-` will perform method validation ensuring that the method is compatible with the generic (i.e. all arguments before `...` have the same names in the same order; if the generic doesn't have `...` all arguments must be same). 
 
-You should only defined a method if you "own" either the generic, or at least one of the classes in the signature. This can not be enforced programmatically (because there's no straightforward defintion of ownership), but will be recommended as best practice.
+Documentation will discuss the risks of defining a method when you don't own either the generic or the class.
 
 `method<-` is designed to work at run-time (not just package build-time) so that methods can be defined when suggested packages are loaded later:
 
--- a/spec/design.md
+++ b/spec/design.md
@@ -263,7 +263,7 @@ method(generic, signature) <- function(x, ...) {}
   character vector. If a character vector is supplied, class objects
   are searched for in the calling frame.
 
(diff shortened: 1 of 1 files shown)


spec/design.md
```diff
@@ -250,26 +250,31 @@ MyClass <- Class(slot1 := integer,
...
-```{R}
-generic[[classObject]] <- function(x) { }
-```
+`method<-` will perform method validation ensuring that the method is compatible with the generic (i.e. all arguments before `...` have the same names in the same order; if the generic doesn't have `...` all arguments must be same). 
 
-For multiple (nested) dispatch:
-```{R}
-generic[[classObject1]][[classObject2]] <- function(x) { }
-```
+You should only defined a method if you "own" either the generic, or at least one of the classes in the signature. This can not be enforced programmatically (because there's no straightforward defintion of ownership), but will be recommended as best practice.
```
lawremi, line 268: Perhaps it would be simplest to just say that the documentation will explain the risks involved when defining a method for a third party generic and class. For example, it is common practice now for generics to be defined by a "stub" package, like the MatrixGenerics package. Ideally all of the "class owners" would implement methods for them, but in practice some other community member may implement compatibility. Being too blunt in our documentation may discourage what would otherwise be a useful contribution. Ultimately, this is more of a "community norm" than a technical consideration.

lawremi (commented):
I didn't realize that `method<-` would replace the syntactic sugar, but after thinking about it, it makes sense to be explicit and consistent. Just left one comment around the documentation.

Issue #36, opened 2020-10-05
Method reflection


hadley:
Given that `method(generic, class) <- method` defines a method (assuming that #35 is merged), I propose that `method(g, s)` retrieve the method for the generic `g` for signature `s`. I think this should be relatively uncontroversial, but there are a couple of questions:

* What should happen if there is no method — should it return an error or NULL? I think `NULL` will be easier to program with since it is not common to use `tryCatch()` for control flow in R.

* Should you be able to request a specific method? (i.e. ignore inheritance)

hadley:
Also if the method reflection syntax is quite simple, is that what generics should use? e.g. adpating from the examples in the design doc would we have:

```R
plot <- function(x, ...) {
  method(plot, classObject(x))(x, ...)
}

publish <- function(x, y, ...) {
  sig <- list(classObject(x), classObject(y))
  method(publish, sig)(x, y, ...)
}
```

hadley:
If we took this approach, we'd probably need some special way to simulate a method "name" in backtraces.

hadley:
And maybe `method()` would attach `generic` and `signature` attributes to the returned object?

hadley:
Need to think about S3 compatibility.

wlandau:
> What should happen if there is no method — should it return an error or NULL? I think NULL will be easier to program with since it is not common to use tryCatch() for control flow in R.

I think I would actually appreciate an informative error. Seems like we might otherwise see a lot of "attempt to apply non-function".

lawremi:
Seems analogous to `get()` vs. `get0()`.  Maybe have an `ifnotfound=` argument?

hadley:
I really don't like having an option that changes the side-effects (error vs not) or return type (function vs function/NULL) so if we have to have a version that throws an error, I think it should be too different functions. But I think if you're doing method introspection, you can be trusted to check the result for `NULL`, if that's clearly documented.

lawremi:
Always behaving the same way, as Hadley suggests, is certainly appealing for its simplicity. It might help to think of some use cases for this situation. Programmatically calling `selectMethod()` is exceedingly rare. The much more common interactive queries would presumably go through `methods()`.

PS: Btw, `selectMethod()` uses the `optional=` argument to handle this (so Hadley wouldn't like it).

wlandau:
[...]

PR #37, opened 2020-10-05
Experiment with new method reflection

hadley:
#36


lawremi:
Happy with these changes. Looks like there is a merge conflict though due to the reorganization.

hadley:
Yeah, I figured that'd happen. I'll probably just re-do from scratch.

PR #44, opened 2020-10-06
Describe validation and how to opt-out of it

### LINKED ISSUE #25: When is validation performed and how to do you suppress it when needed?
One strawman proposal for how to suppress temporarily:

```R
eventually_valid <- function(code) {
  options(check_validity = FALSE)
  on.exit(options(check_validity = TRUE))
  
  validate(code)
}
```

Still need to consider when the OO system would call `validate()` automatically.

hadley:
Fixes #25

### DIFF
--- a/spec/design.md
+++ b/spec/design.md
@@ -108,6 +108,11 @@ move_right <- function(x, y) {
 
 This ensures that the validation will not trigger if `x@start + y` is greater than `x@end`.
 
+The system also provides `implicitlyValid()` for expert use only.
+This is similar to `eventuallyValid()` but does not check for validity at the end.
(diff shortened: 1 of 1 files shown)


lawremi (approved):

hadley:
@mmaechler @ltierney does this look ok to you? (goal is to capture discussion in the linked issue, not to declare this as the final answer)

ltierney (approved):

mmaechler (changes requested):
Really a good idea to have both  `eventuallyValid(object, *)`  and `implicitlyValid()` with the small crucial difference as proposed.

I'd prefer to call the 2nd argument `FUN` or similar as I would want to require a function there, and the word `code` is too vague for that..

PR #46, opened 2020-10-07
Define class unions

### LINKED ISSUE #31: Class unions
The methods package defines class unions as new classes, inserted above the members. Messing with the class hierarchy causes all sorts of complications, and depends on multiple inheritance.

We could instead define class unions as instances, i.e., `union(classA, classB)` returns a ClassUnion object. Using a ClassUnion in a signature could define a separate method for each member of the union. Or, this could be encoded into the dispatch logic. Using a ClassUnion as a property type would tell the validator to return valid if the value is an instance of any of the members. I've seen this in Julia and TypeScript, and it's probably in many others. It probably helps that they're compiled.

TypeScript also has the notion of a type intersection, which makes sense given that it's entirely based on structural/duck typing. Not sure it applies in our case, particularly if we preclude multiple inheritance.

hadley:
I took at stab at turning what you described into slightly more formal language — let me know if you think I missed anything important.

Fixes #31

### DIFF
--- a/spec/design.md
+++ b/spec/design.md
@@ -82,6 +82,25 @@ Steps 2 and 3 are similar to calling `structure()`, except that property values 
 
 ### Validation
 
+### Unions
+
+A class union represents a list of possible classes.
+It is used in properties to allow a property to be one of a set of classes, and in method dispatch as a convenience for defining a method for multiple classes.
+
+``` {.R}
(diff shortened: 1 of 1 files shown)


lawremi (approved):

hadley:
@mmaechler does this look ok to you?

mmaechler (approved):
looks good to me.

PR #54, opened 2020-11-17
Start on dispatch documentation

hadley:
Based on Monday's discussion plus some more re-writing after the fact. Still needs plenty of discussion, but this should give us a framework to organise the details.

### DIFF
--- a/spec/dispatch.Rmd
+++ b/spec/dispatch.Rmd
@@ -1,0 +1,262 @@
+---
+output: github_document
+---
+
+```{r, include = FALSE}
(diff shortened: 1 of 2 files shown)


lawremi (approved):

wlandau (approved):

mmaechler (approved):
I'm approving  (but not in the sense that I would have read everything careful, and would be sure to agree) .. because it's a really nice write up, listing open and half-open questions, some quite important , and it  is a good basis for discussion in cases of doubt, and a valid starting point for implementation.

hadley:
Thanks all. I think this should give us a solid framework for further discussion.

PR #74, opened 2021-10-05
Use R7 classes for base classes

jimhester:
This is a step towards maving quoting more consistent with class names for https://github.com/RConsortium/OOP-WG/issues/66.

This would allow you to use the R7 classes for base types, so rather than `"character"` you would use `chr`.

However we still need to support quoted classes for S3 support, as S3 classes don't have a class object like R7 classes do.


lawremi:
Could we just use `character` instead of `chr`? Such as by checking the identity of the primitive (`character()` in this case).

jimhester:
We could, the advantage to what I did here was that these are normal R7 class objects vs special cases for the base types. But we could go either way I guess.

hadley:
Or maybe we could special case vectors of length 0? 

```R
  if (length(name) == 0 && (is.atomic(x) || is.list(x))) {
    base_classes[[class(name)]]
  }
```

hadley:
Nope, that's confusing because then you use `parent = R7_class` and `parent = character()`.

PR #108, opened 2021-11-02
Property handling overhaul

### LINKED ISSUE #96: Prop family of functions
* `property()` -> `prop()`
* new `prop_names()`, a la `slotNames()`
* new `props()`, a la `attributes()`
* new `prop_exists()` that returns a boolean

hadley:
The main change here is that `new_class()` now generates a complete list of property objects. This means it's now fast to figure out what properties an object should possess. 

I also renamed the accessors to use `prop` instead of `property`, and reworked `prop_safely()` and `prop()` to use a new `prop_get()` helper, and hence have somewhat more consistent behaviour. (And now you can set a property to `NULL` and retrieve its value without error).

Fixes #96


R/property.R
```diff
@@ -84,66 +87,110 @@ new_property <- function(name, class = NULL, getter = NULL, setter = NULL) {
...
+prop_obj <- function(object, name) {
+  class <- object_class(object)
+  attr(class, "properties")[[name]]
+}
+
+#' @rdname prop
+#' @export
+prop_names <- function(object) {
+  if (inherits(object, "R7_class")) {
+    names(attributes(object))
```
hadley, line 155: I'm not 100% sure this is correct; I just intuited my way to a solution. I think it's something to do with the way we have bootstrap a R7 class out of nothing, or possibly because there's two meanings of properties: the "formal" properties (i.e. the spec) and the actual properties.

jimhester:
Thanks, this looks good to me!

PR #112, opened 2021-11-11
Fully define constructor

### LINKED ISSUE #102: Can't construct object with class hierarchy
``` r
library(R7)
foo <- new_class("foo", properties = list("x" = "numeric"))
foo2 <- new_class("foo2", foo, properties = list("y" = "numeric"))
foo2(x = 1, y = 2)
#> Error in do.call(base::`@`, list(object, name)): trying to get slot "parent" from an object of a basic class ("character") with no slots
```

<sup>Created on 2021-10-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I feel like I'm missing something obvious here, so this might just need a better error.

### LINKED ISSUE #103: Ensure that object fully initialised
```R
library(R7)
foo <- new_class("foo", properties = list("x" = "numeric"))
f <- foo()
f@x
#> Error: Can't find property <foo>@x
```

And maybe define a default value in `new_property()`?

hadley:
The default constructor gets one argument for each property, passing on to the parent as appropriate.

Fixes #102. Fixes #103.

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -15,18 +15,18 @@ new_base_class <- function(name) {
 }
 
 base_classes <- new.env(parent = emptyenv())
-base_classes[["logical"]]    <- new_base_class("logical")
-base_classes[["integer"]]    <- new_base_class("integer")
-base_classes[["double"]]     <- new_base_class("double")
-base_classes[["numeric"]]    <- new_base_class("numeric")
-base_classes[["complex"]]    <- new_base_class("complex")
-base_classes[["character"]]  <- new_base_class("character")
-base_classes[["factor"]]     <- new_base_class("factor")
-base_classes[["raw"]]        <- new_base_class("raw")
-base_classes[["function"]]   <- new_base_class("function")
-base_classes[["list"]]       <- new_base_class("list")
+base_classes[["logical"]] <- new_base_class("logical")
+base_classes[["integer"]] <- new_base_class("integer")
+base_classes[["double"]] <- new_base_class("double")
+base_classes[["numeric"]] <- new_base_class("numeric")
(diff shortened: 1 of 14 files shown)


R/zzz.R
```diff
@@ -10,19 +10,23 @@ R7_object <- new_class(
   }
 )
 
+new_base_class <- function(name) {
+  R7_class(name = name, constructor = function(.data) new_object(.data))
```
hadley, line 14: I think the constructor argument has to be called `.data` (since that's what the property is) and they shouldn't have a default for consistency with the other R7 objects.

hadley:
@jimhester this probably needs more tests, but it would be good to get a second pair of eyes on it before I go much further. I'd suggest starting with https://github.com/RConsortium/OOP-WG/pull/112/files#diff-e1ddedf6c866fcae78a08c0656dd95f6da48dbd33c398917b9fd724685bd1e6d

tests/testthat/_snaps/object.md
```diff
@@ -4,19 +4,19 @@
       foo <- new_class("foo")
       foo(1)
     Error <simpleError>
-      All arguments to <foo> constructor must be named
+      unused argument (1)
```
jimhester, line 7: Are ok with these error messages now being less informative?

I think so since the constructor now has the properties as full arguments, but wanted to double check

jimhester (approved):

tests/testthat/_snaps/object.md
hadley: Yeah, it's a bit of a bummer (especially since I just wrote this code 😆), but I think it's a net win.

Issue #85, opened 2021-10-27
Method consistency changes


hadley:
To support discussion in #7:

``` r
library(R7)

foo1 <- new_generic("foo1", fun = function(x, ...) {
  method_call()
})
# Should be ok: ... in generic just indicate that methods can have
# additional arguments
method(foo1, "character") <- function(x) ""
#> Error: `method` must be consistent with <R7_generic> foo1.
#> - `generic` has `...`
#> - `method` does not have `...`

foo2 <- new_generic("foo2", fun = function(x) {
  method_call()
})
# Should error: generic doesn't have dots
method(foo2, "character") <- function(x, y = 1) ""

foo3 <- new_generic("foo3", fun = function(x, ..., y = TRUE) {
   method_call()
})
# Should error: no y argument
method(foo3, "foo3") <- function(x, ...) ""
```

<sup>Created on 2021-10-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
I'm confused by the `mean2` example in [`?new_generic`](https://rconsortium.github.io/OOP-WG/reference/new_generic.html), specifically "If you want to require methods implement additional arguments":

- Do arguments after the ellipsis have a special meaning?
- Do they have to exist in each method?

Interface evolution was very difficult in DBI.

hadley:
Arguments after `...` don't have a special meaning, but by convention that's where you put extra arguments. These would be required in every method.

hadley:
Notes from meeting:

Proposed principle: If a generic has `...`, the formals of the method must be the formals the of generic + extra arguments. If a generic doesn't have `...`, the formals of the method must be identical to the formals of the generic.

Would make sense for methods to supply defaults since the generic often does not have enough information to supply. For example, generalise of something like `digits` to `precision`, where the default for dates/times would be specified differently to numbers.

DavisVaughan:
In dplyr, for example, we have `summarize()` which has a `.groups` argument in the generic, but the dtplyr method doesn't have that. I guess this would be disabled?

```r
> dplyr::summarise
function (.data, ..., .groups = NULL) 
{
    UseMethod("summarise")
}
<bytecode: 0x7fb9ba5ceb88>
<environment: namespace:dplyr>

> dtplyr:::summarise.dtplyr_step
function (.data, ...) 
{
```

DavisVaughan:
[...]

PR #117, opened 2021-11-12
Use a consistent signature specification

hadley:
A signature is a character vector giving the names of the arguments to dispatch on.


hadley:
This makes the presence/absence of `...` in the generic formals a little inconsistent, but I'll clean that up as part of #85

jimhester (approved):

PR #118, opened 2021-11-12
Look up base classes from constructor functions

### LINKED ISSUE #66: The syntax is sometimes inconsistent
For example, class/function names sometimes *have* to be quoted, sometimes *can't* be quoted,
and sometimes can, but don't have to be quoted – in the latter case, I'm not sure if using it
incorrectly will not cause an error further down the road.

hadley:
Fixes #66. Closes #74

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -1,10 +1,9 @@
 #' @importFrom utils modifyList
 R7_class <- function(name, parent = R7_object, constructor = NULL, validator = function(x) NULL, properties = list()) {
-  if (is.character(parent)) {
-    parent_obj <- class_get(parent)
-    if (!is.null(parent_obj) && inherits(parent_obj, "R7_class")) {
-      parent <- parent_obj
-    }
+
+  parent_obj <- class_get(parent)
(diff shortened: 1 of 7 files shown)


jimhester (approved):

jimhester:
This seems good, as you noted we still have the problem on how to deal with S4 classes and unions and S3 classes.

Maybe we could have some sort of registration function similar to `setOldClass()`? It would be nice if we could avoid having to use strings to specify classes.

hadley:
Yeah, I was thinking maybe we could have `new_S3_class(class_vector, constructor, validator)`. You could use this explicitly or `class_get()` would generate it automatically with a warning.

lawremi:
For S3 classes, I think we could expect people to just define proper R7 classes, as long as their objects can be implicitly compatible (would they be?). For S4 classes, perhaps there is a way to just convert them to R7 class objects. Is registration really necessary, or could we simply use objects to represent classes?

hadley:
Yeah, the problem is easier for S4, since we already have explicit class objects. For S3 objects, we need to create to create some placeholder object to represent the class — the question is whether it's worth having a specialised helper instead of creating a full R7 spec.

PR #121, opened 2021-11-12
Quality of life improvements for unions

hadley:
* Add a print method
* Use class_get() to support new way of specifying base types
* Allow union of unions


jimhester (approved):
Nice, thanks for making this print method for unions, I ran into that issue but hadn't gotten around to solving it yet.

PR #124, opened 2021-11-13
Add more error checking to new_generic()

### LINKED ISSUE #115: Should `new_generic()` check that fun calls `method_call()`
To make it harder to accidentally mess up

hadley:
Fixes #115


jimhester (approved):

R/generic.R
```diff
@@ -134,3 +136,33 @@ print.R7_generic <- function(x, ...) {
...
+
+  if (identical(x[[1]], name)) {
+    return(TRUE)
+  }
+
+  if (length(x) == 1) {
+    return(FALSE)
+  }
+
+  for (i in seq(2, length(x))) {
```
jimhester, line 162: It is not possible for `body(fun)` to return length 0 right? (and then mess up this for loop)

hadley: I don't think so; a call has to be at least length 1 (the function being called)

jimhester: One case I thought of is for primitive functions, e.g.

``` r
body(sum)
#> NULL
```

<sup>Created on 2021-11-16 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Not sure if that case would ever come up for this usage, but it might be worth adding a conditional for 0 length just in case?

hadley: Oh yeah, good catch. I'll just change the `length(x) == 1` check above

hadley: Although in that case the check for `is.call()` above will have failed.

PR #125, opened 2021-11-14
Refactor new_generic()

hadley:
This is mostly refactoring to make it easier to implement the changes to `...` handling. There are two minor changes to behaviour:

* I've switched the order of the `signature` and `fun` in `new_generic()`. I think this is more natural because it forces you to spell out the name of the `signature` argument.
* `guess_signature()` no longer strips `...`; this makes it more compatible with the case where you generate the `fun` from the `signature`; this makes more consistent before I change the handling of `...` in both cases.


jimhester (approved):

tests/testthat/test-external-generic.R
```diff
@@ -0,0 +1,28 @@
+test_that("new_method works with both hard and soft dependencies", {
+  skip_on_os("windows")
+  skip_if(quick_test())
```
jimhester, line 3: In case you are annoyed about the tests taking a while, you can set `quick_test_enable()` to disable this particular test, which takes a while because it has to install a bunch of packages. It also sometimes fails until you actually install R7 depending on what you changed. I usually turn on the quick tests it when I am working on R7.

lawremi:
Does the user ever see that `...` is part of the signature, even though we are not dispatching using `...`?

hadley:
@lawremi what do you mean by signature? I think we might be using it to mean two slightly different things: the complete formals argument of the generic vs. which arguments are used for dispatch. Either way, we'll need to show both in the print method of the generic.

Issue #116, opened 2021-11-12
Use dispatch_on instead of signature?


hadley:
What do you think of making `signature` a bit more concrete and renaming it to `dispatch_on`, since it's a vector of argument names that are used for dispatch?  

@mmaechler @lawremi do you think this would be too confusing for existing S4 users?

lawremi:
This sounds like a good change; I think S4 users will quickly catch on. Would we also use that name for the method signature?

hadley:
(Also need to display in the print method for generics)

hadley:
Notes from meeting:

* No need to re-use signature, especially because there's some confusion about it. 
* Would be nice to have a noun, so it's a property of the generic, e.g. `dispatch_args`.
* Seems ok to require `dispatch_args` to be `names(formals(generic))[1:n]` in interest of simplicity and performance.

PR #126, opened 2021-11-14
Start on updating method_call_

hadley:
@jimhester can you please help me finish this off? My C is pretty rusty so I need some pointers.


src/code.c
```diff
@@ -201,6 +180,24 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
     R7_method_lookup_error(generic, signature_classes, envir);
   }
 
+  // Now add the remaining, non-dispatched, args
```
hadley, line 183: The loop above builds up the argument list for all arguments that are dispatched on. Now we need to do the same thing for the remaining arguments.

src/code.c
```diff
@@ -201,6 +180,24 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
...
+  SEXP gen_formals = FORMALS(generic);
+  R_xlen_t gen_formals_len = Rf_xlength(gen_formals);
+  if (gen_formals_len > gen_signature_len) {
+    for (R_xlen_t i = gen_signature_len; i < gen_formals_len; ++i) {
+      // Find value of that argument
+      SEXP name = Rf_install(CHAR(STRING_ELT(gen_signature_args, i)));
```
hadley, line 189: What's the best way to get the names of the remaining arguments? (I'm just using `gen_signature_args` as a placeholder)

src/code.c
```diff
@@ -201,6 +180,24 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
...
+      SEXP arg = Rf_findVar(name, envir);
+
+      // And add to args
+      SETCDR(tail, Rf_cons(arg, R_NilValue));
+      tail = CDR(tail);
```
hadley, line 194: Do we need to set the `TAG()` here too? I don't think the argument names matter for the dispatch argument (since they're always in the correct order), but I think it will matter here.

hadley:
(No rush, because I really need to work on other stuff, and I'm leaving for vacation on Wednesday morning)

hadley:
Per #116, can require that manually supplied `dispatch_args` is `names(formals(generic))[seq_along(dispatch_args)]`

PR #129, opened 2022-01-10
Rename signature to dispatch_args

### LINKED ISSUE #116: Use dispatch_on instead of signature?

hadley:
Note while generics no longer have a signature, methods still do (it's one class for each `dispatch_arg`).

I still prefer `dispatch_on` because it reads more naturally in `new_generic(foo, dispatch_on = "x")`, but I've implemented `dispatch_args` here.

Fixes #116


R/generic.R
```diff
@@ -7,7 +7,7 @@
 #'
 #' @param name The name of the generic. This should be the same as the object
 #'   that you assign it to.
-#' @param signature A character vector providing the names of arguments to
+#' @param dispatch_args A character vector providing the names of arguments to
 #'   dispatch on. If omitted, defaults to the required arguments of `fun`.
```
DavisVaughan, line 11: Does everyone know that "the required arguments of `fun`" `==` "the arguments of `fun` that don't have a default value"?

hadley: This gets changed again in #131, so I don't think we need to worry about it here.

lawremi (approved):
Looks good. `dispatch_on` seems more fluent, while `dispatch_args` is more descriptive. Interesting tradeoff.

Issue #110, opened 2021-11-08
Can we eliminate `next_method()`?


hadley:
(Placeholder issue based on discussion in meeting today).

Or at least reduce the dynamism as much as possible? How does Julia get around it?

hadley:
Maybe we could make `method_call(class)` call the method for the specified class? 

```R
# S3 method
`[.my_class` <- function(x, i) {
  new_my_class(NextMethod(), x = attr(x, "x"), y = attr(x, "y"))
}

# Equivalent R7 method
method(`[`, my_class) <- function(x, i) {
  new_my_class(method_call(R7_object), x = x@x, y = x@y)
}
```

This has the advantage of being crystal clear about which method you want to call, and I think adds relatively little work since when you write `NextMethod()` you presumably already have some idea of what method you want it to call.

Would need to verify that this also solves #119

lawremi:
I think what Jan (or a compiler) would like is to know at the call site which method is being called, so that would be both the generic and the signature. Something like `Class::method()` in C++ but of course we have to handle multiple dispatch. 

Maybe if we had some syntax for up-casting? For example, in S4 one can up-cast (strip off the subclass parts) with `new("SuperClass", instance_of_subclass)`. What would be the R7 equivalent? f we are going to come up with a formal construction syntax anyway, perhaps it could support this type of casting, as well?

hadley:
@lawremi my concern with replacing `NextMethod()` with upcasting + re-calling the generic is that it's still pretty painful since you have to explicitly pass along all of the extra arguments:

```R
method(`[`, my_class) <- function(x, i, ..., a = 1, b = 2, c = 3) {
  new_my_class(method_call(R7_object), x = x@x, y = x@y)
}

# vs
method(`[`, my_class) <- function(x, i, ..., a = 1, b = 2, c = 3) {
  parent <- upcast(x, R7_object)
  new_my_class(parent[x, i, ..., a = a, b = b, c = c], x = x@x, y = x@y)
}
```

(the example doesn't quite work here with `[`, but hopefully you get the idea)

In my experience, remembering to pass along all the additional arguments is a common source of error. That said, in my approach, it's not clear how you avoiding passing along arguments that only exist on this method, not the parent method, or the generic. Maybe `method_call()` would only pass along arguments that are part of the formals of the generic? But this also depends on #85, and stuff that's currently in my head and not explained well or implemented.

lawremi:
[...]

PR #130, opened 2022-01-10
Replace x@.data with r7_data(x)

### LINKED ISSUE #114: Access `.data` with a function instead of `@`?
I think that would make things more clear since it's not actually a property. (I also wonder if maybe calling it something like `parent` would be more clear)

hadley:
Fixes #114

I don't think `r7_data()` is the right name for this because no other function starts with `r7_`. I didn't try to come up with something better because it feels like this function might be transient: I think it's actually just a special case of up casting the object. So I'll take another stab at it as part of #110.


lawremi (approved):
Looks good. I agree that this is a special case of up-casting.

PR #134, opened 2022-01-17
Class specification

### LINKED ISSUE #120: Common class specification
We need a common mechanism for concerting a class specification (e.g. R7 class object, S4 class, S3 class, base type) wherever they're used (as parent, in properties, in class unions, in `method<-`).

We might want some way to be explicit about S3 classes, maybe by providing something like `new_S3_class(class_vector, constructor, validator)`. We could continue to allow a character vector for S3 classes (possibly with a warning), or require an S3 class object.

hadley:
This is a rather large set of changes but should have relatively little impact on the actual interface. The main goal is to bring consistency to the way that classes are specified as parents, in properties, in unions, and in method registration/dispatch. 

The main change to functionality is that specifying a class with a string (e.g. `method(foo, "classname")`), can now only refer to a base type. It could previously mean one of three things: the name of an R7 class defined in the calling environment, a S3 class, or a base type. This ambiguity seemed dangerous to me, so I reduced it by eliminating the lookup in the calling environment and introduced a new `s3_class()` helper for registering S3 classes.

Fixes #120


R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+#'
+#' @param x A class specification.
+#'    * An R7 class object or class union.
+#'    * An S3 class object, created by `s3_class()`.
+#'    * An S4 class object.
+#'    * A base type specified either with its constructor (`logical`, `integer`,
+#'      `double` etc) or its name (`"logical"`, `"integer"`, "`double`" etc).
+#' @param arg_name Argument name used when generating errors.
+#' @export
+as_class <- function(x, arg_name = "as_class()") {
```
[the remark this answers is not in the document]
DavisVaughan: Oh I see how you use this down in `as_signature()`, so never mind

R/method.R
```diff
@@ -36,11 +36,12 @@
...
 #' # inspect its implementation
-#' method(bizarro, list("numeric"))
-#' method(bizarro, list("factor"))
+#' method(bizarro, "numeric")
+#' method(bizarro, s3_class("factor"))
 method <- function(generic, signature) {
-  signature <- as_signature(signature)
+  # TODO: check that signature doesn't contain any unions
 
+  signature <- as_signature(signature)
```
DavisVaughan, line 44: It may be worth checking that at least 1 dispatch arg is supplied in the signature, because currently you can do:

``` r
library(R7)

bizarro <- new_generic("bizarro", dispatch_args = "x")

method(bizarro, list())
#> Error in signature[[1]]: subscript out of bounds
```

<sup>Created on 2022-02-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan: Or maybe it is a bug in `method_impl()` because that is where `signature[[1]]` is called

R/method.R
```diff
@@ -60,46 +61,40 @@ as_signature <- function(signature) {
...
 
-    out <- getS3method(generic, signature[[1]][[1]], optional = TRUE)
+  # If no R7 method is found, see if there are any S3 methods registered
+  if (inherits(generic, "R7_generic")) {
+    args <- generic@dispatch_args
+    generic <- generic@name
+  } else {
+    args <- names(formals(generic))
+    generic <- find_function_name(generic, topenv(environment(generic)))
+  }
```
DavisVaughan, line 84: There was a `args <- setdiff(args, "...")` here before, was removing that intentional?

R/method.R
```diff
@@ -214,21 +209,27 @@ new_method <- function(generic, signature, method, package = NULL) {
 
 
   for (i in seq_along(signature)) {
+    type <- class_type(signature[[i]])
```
DavisVaughan, line 212: This looks unused

R/method.R
```diff
@@ -237,6 +238,26 @@ new_method <- function(generic, signature, method, package = NULL) {
...
+  switch(class_type(x),
+   s3 = class(x),
+   s4 = class(x),
+   r7 = x@name,
+   r7_base = typeof(x),
+   stop("Unsupported")
+  )
+}
+# Class name when registering an R7 method
+r7_class <- function(x) {
```
DavisVaughan, line 252: Having both this and `R7_class()` really threw me off 😬

_pkgdown.yml
```diff
@@ -12,7 +12,7 @@ reference:
 
 - title: Introspection
   contents:
-  - class_get
+  - as_class
   - class_names
   - object_class
   - method
```
DavisVaughan, line 18: Entry for `s3_class`?

R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+    }
+    base_classes[[names(candidate)[[1]]]]
+  } else if (is.character(x) && length(x) == 1) {
+    if (x %in% names(base_classes)) {
+      base_classes[[x]]
+    } else {
+      stop(sprintf("%s: Can't find base class called '%s'", arg_name, x), call. = FALSE)
+    }
+  } else {
+    stop(sprintf("%s: class specification must by a R7 class object, result of s3_class(), a S4 class object, or a base constructor function, not a %s.", arg_name, obj_desc(x)), call. = FALSE)
```
DavisVaughan, line 41:
```suggestion
    stop(sprintf("%s: class specification must be an R7 class object, the result of `s3_class()`, an S4 class object, or a base constructor function, not a %s.", arg_name, obj_desc(x)), call. = FALSE)
```

+ regenerate snapshots

tests/testthat/_snaps/property.md
```diff
@@ -17,3 +17,26 @@
...
+    Error <simpleError>
+      <my_class>@s4 must be of class <class_s4>, not <character>
+    Code
+      my_obj@r7 <- "x"
+    Error <simpleError>
+      <my_class>@r7 must be of class <class_r7>, not <character>
+    Code
+      my_obj@r7_union <- "x"
+    Error <simpleError>
+      <my_class>@r7_union must be of class <integer> u <logical>, not <character>
```
DavisVaughan, line 41: How do you feel about a capital U? `<integer> U <logical>`. I don't really mind either way but I think it looks slightly more like the typical union symbol

tests/testthat/test-class-spec.R
```diff
@@ -0,0 +1,141 @@
...
+  klass <- new_union(text, number)
+  expect_equal(as_class(klass), klass)
+
+  expect_equal(class_type(klass), "r7_union")
+  expect_equal(class_desc(klass), "<text> u <number>")
+  expect_equal(class_deparse(klass), "new_union(text, number)")
+
+  # Can't have an instance of a union
+  expect_equal(class_inherits(text("x"), klass), TRUE)
+  expect_equal(class_inherits(number(1), klass), TRUE)
```
DavisVaughan, line 25: What does this comment mean?

R/class-spec.R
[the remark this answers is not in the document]
hadley: It's supposed to match what we do in rlang. Should it have a different name?

tests/testthat/_snaps/property.md
hadley: I was actually thinking about switching to `|` . What do you think about that?

R/class-spec.R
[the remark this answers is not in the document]
DavisVaughan: I think what struck me as odd is that `arg_name` can refer to either the function name or the object depending on the context. Given that it is like rlang, I think I would have expected:

``` r
as_class <- function(x, arg = "x") {
  base <- sprintf("Can't convert `%s` to a valid class.", arg)
  
  if (is.function(x)) {
    stop(paste0(base, "\n", "- Can't find base class corresponding to supplied constructor function."), call. = FALSE)
  } else {
    stop(paste0(base, "\n", sprintf("- Can't find base class called '%s'.", x)), call. = FALSE)
  }
}

as_class(identity)
#> Error: Can't convert `x` to a valid class.
#> - Can't find base class corresponding to supplied constructor function.

as_class("chr", arg = "arg[[2]]")
#> Error: Can't convert `arg[[2]]` to a valid class.
#> - Can't find base class called 'chr'.
```

<sup>Created on 2022-02-04 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/_snaps/property.md
DavisVaughan: I also don't mind `<union<logical, integer>>`, which makes sense if you consider that union is just another "type" so it should get wrapped in `< >`.

I think something that is currently missing from the current approach with either `U` or `|` is some kind of signal for the start / end of the union - which would be the `<union` and then the final `>` in the other approach I suggested.

I don't mind the `|` that much though, if you want to go with that

R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+    r7_union = {
+      classes <- vcapply(x@classes, class_deparse)
+      paste0("new_union(", paste(classes, collapse = ", "), ")")
+    }
+  )
+}
+
+class_inherits <- function(x, what) {
+  switch(class_type(what),
+    NULL = TRUE,
```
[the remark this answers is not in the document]
hadley: Yeah, this is used in properties to mean "everything".

R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+  } else if (isS4(x)) {
+    "s4"
+  } else {
+    stop("`x` is not standard R7 class")
+  }
+}
+
+class_desc <- function(x) {
+  switch(class_type(x),
+    NULL = "",
```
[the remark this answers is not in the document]
hadley: Maybe this would be better as `<anything>`? I'll try it out.

tests/testthat/_snaps/property.md
hadley: I was thinking it doesn't need a signal here because unions can't be nested (they're always automatically flattened), so it's just enumerating the set of possible values here.

Maybe make it `<my_class>@r7_union must be of class <integer> or <logical>, not <character>`

tests/testthat/test-class-spec.R
hadley: It's a placeholder for the missing pattern from the other tests. I'll clarify.

R/class-spec.R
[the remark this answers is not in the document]
hadley: Sounds good, although I'll leave the errors unbulleted to match base R style.

tests/testthat/_snaps/property.md
DavisVaughan: Okay, I think enumerating is fine, I like the `or` approach. Possibly with support for >=3 like `<character>, <double> or <integer>` depending on your Oxford comma preference

R/class-spec.R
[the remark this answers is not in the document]
hadley: `<ANY>` seems to look good, at least for now

R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+      classes <- vcapply(x@classes, class_deparse)
+      paste0("new_union(", paste(classes, collapse = ", "), ")")
+    }
+  )
+}
+
+class_inherits <- function(x, what) {
+  switch(class_type(what),
+    NULL = TRUE,
+    s3 = !isS4(x) && inherits(x, what[[1]]),
```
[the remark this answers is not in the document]
hadley: `what` is the output of `class_as()`, which in the case of S3 is a character vector. This isn't 100% correct, but I'm not sure how you're supposed to check this correctly. Maybe something like `is_prefix(what, class(x))`?

R/class-spec.R
```diff
@@ -0,0 +1,150 @@
...
+    s4 = isS4(x) && methods::is(x, what),
+    r7 = inherits(x, "R7_object") && inherits(x, what@name),
```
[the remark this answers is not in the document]
hadley: You forgot to make an instance of the class: `class_inherits(foo(), foo)`

R/method.R
hadley: Now at https://github.com/RConsortium/OOP-WG/issues/146

R/method.R
hadley: Good catch, thanks.

PR #135, opened 2022-01-17
Make base types more consistent

### LINKED ISSUE #77: Should vectors extend common vector type?
So that (e.g.) you could define a method that worked on all vectors.

### LINKED ISSUE #94: Integer should inherit from numeric?
``` r
library(R7)

method(foo, "numeric") <- function(x, ...) {
  10
}
foo(1:10)
#> Error: Can't find method for generic `foo()` with classes:
#> - x: <integer>
```

<sup>Created on 2021-10-28 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hadley:
Changes to base types:
* make numeric a union
* add new atomic and vector unions
* remove factor and data.frame S3 classes
* add expression base type

And includes a few quality of life improvements:

* Method dispatch failure now reports correct classes
* Methods created from union now have own names
* More explicit error when parent is a union

Fixes #77. Fixes #94. Built on top of #134.

### DIFF
--- a/R/class-spec.R
+++ b/R/class-spec.R
@@ -94,7 +94,7 @@ class_inherits <- function(x, what) {
     s3 = !isS4(x) && inherits(x, what[[1]]),
     s4 = isS4(x) && methods::is(x, what),
     r7 = inherits(x, "R7_object") && inherits(x, what@name),
-    r7_base = inherits(x, what@name),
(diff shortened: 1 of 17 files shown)


hadley:
Replaced by #142

PR #147, opened 2022-02-04
Unions for base types

### LINKED ISSUE #77: Should vectors extend common vector type?

### LINKED ISSUE #94: Integer should inherit from numeric?

hadley:
* Adds numeric, atomic, and vector unions
* Adds expression base type

Fixes #77. Fixes #94.


hadley:
@davisvaughan can you take a look at this PR too? The goal is to treat "numeric", "atomic", and "vector" like unions; i.e. you can't extend from them, but you can dispatch and use in properties.

R/zzz.R
```diff
@@ -72,6 +73,20 @@ R7_union <- new_class(
...
 
+#' @export
+str.R7_union <- function(object, ..., nest.lev = 0) {
+  cat(if (nest.lev > 0) " ")
+  cat("<R7_union>: ", class_desc(object), sep = "")
+  cat("\n")
+
+  if (nest.lev == 0) {
+    props <- props(object)
+    props$srcref <- NULL
```
DavisVaughan, line 84: Where does this come from?

R/zzz.R
```diff
@@ -107,3 +122,10 @@ global_variables(c("name", "parent", "properties", "constructor", "validator"))
   env <- as.environment(paste0("package:", pkgname))
   env[[".conflicts.OK"]] <- TRUE
 }
+
+.onLoad <- function(...) {
+  base_classes$`NULL` <- new_base_class("NULL")
```
DavisVaughan, line 127:
```suggestion
  base_classes$`NULL` <<- new_base_class("NULL")
```

? just guessing based on the others

src/dispatch.c
```diff
@@ -51,7 +51,7 @@ SEXP method_internal(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ig
 SEXP get_class(SEXP object, SEXP envir) {
     static SEXP fun = NULL;
     if (fun == NULL) {
-      fun = Rf_findVarInFrame(R_BaseEnv, Rf_install("class"));
+      fun = Rf_findVarInFrame(R_BaseEnv, Rf_install(".class2"));
```
DavisVaughan, line 54: This is kind of nice for matrices/arrays too, even though I think we haven't fully thought about those yet

> .class2(matrix(1))
[1] "matrix"  "array"   "double"  "numeric"
> class(matrix(1))
[1] "matrix" "array"

DavisVaughan (approved):
Looks nice!

src/dispatch.c
hadley: Yeah, we could always back port, if needed.

R/zzz.R
hadley: No idea; and it's not needed. I'll remove.

tests/testthat/helper.R
```diff
@@ -1,5 +1,5 @@
 text <- new_class("text", parent = "character", constructor = function(text = character()) new_object(.data = text))
-number <- new_class("number", parent = "numeric", constructor = function(x) new_object(.data = x))
+number <- new_class("number", parent = "double", constructor = function(x) new_object(.data = x))
```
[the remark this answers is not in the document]
hadley: Yeah, `parent` can't be a union because it has to be a concrete type. There's already a test for it.

tests/testthat/test-class.R
```diff
@@ -33,18 +33,14 @@ describe("R7_class", {
 })
 
 test_that("classes can inherit from base types", {
-  types <- c("logical", "integer", "complex", "character", "raw", "list")
+  types <- c("logical", "integer", "double", "complex", "character", "raw", "list")
```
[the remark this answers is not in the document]
hadley: I'll do that in a separate PR since it's unrelated to this one.

tests/testthat/test-class.R
```diff
@@ -33,18 +33,14 @@ describe("R7_class", {
...
   for (type in types) {
     f <- match.fun(type)
     foo <- new_class("foo", parent = type, constructor = function(x = f()) new_object(x))
     obj <- foo()
     expect_equal(typeof(r7_data(obj)), type)
   }
 
-  foo <- new_class("foo", parent = "numeric", constructor = function(x = numeric()) new_object(x))
-  obj <- foo()
-  expect_equal(typeof(r7_data(obj)), "double")
```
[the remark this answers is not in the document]
hadley: See the next test 😉

PR #154, opened 2022-02-09
Refactor method() and method<-()

### LINKED ISSUE #137: Unify method dispatch and method introspection
`method()` and `method_call()` need to use identical code paths. Need to think about how S4 classes should work, and generally how a class becomes a string that can index an environment. Should we use a single environment instead of nesting?

hadley:
* separate tests and documentation for method registration vs. introspection
* `method()` now lives in `dispatch.R`
* `new_method()` is now `register_method()`
* Test all existing forms of method registration
* Add more information to `check_method()` failures

Fixes #137


R/generic.R
```diff
@@ -172,3 +175,37 @@ has_call <- function(x, name) {
...
+  if (!is.environment(x)) {
+    return(x)
+  }
+
+  # Recursively collapse environments to a list
+  methods <- lapply(names(x), function(class) methods_rec(x[[class]], c(signature, class)))
+  unlist(methods, recursive = FALSE)
+}
+
+generic_add_method <- function(generic, signature, method) {
```
DavisVaughan, line 193: It would be helpful for me if we could talk about this, because I don't quite understand the iterated dispatch bit

R/method.R
```diff
@@ -1,145 +1,163 @@
-#' Retrieve or register an R7 method for a generic
+#' Register a R7 method for a generic
```
DavisVaughan, line 1:
```suggestion
#' Register an R7 method for a generic
```

R/method.R
```diff
@@ -1,145 +1,163 @@
...
 #'
 #' @description
-#' Generics partition a function into interface (a generic) and implementation
-#' (many methods). `method<-` allows you to register a method, an
-#' implementation for a specified class signature, with a generic.
+#' A generic defines the interface of a function. Once you have created a
+#' generic with [new_generic()], you provide implementations for specific
+#' signatures by registering methods with `method<-`
```
DavisVaughan, line 6:
```suggestion
#' signatures by registering methods with `method<-`.
```

R/method.R
```diff
@@ -1,145 +1,163 @@
...
-#' `method()` retrieves a method for a given signature. You typically shouldn't
-#' need this function while programming, because calling the generic will
-#' automatically dispatch to the correct method, but it's often useful
-#' interactively in order to see the implementation of a specific method.
+#' The goal is for `method<-` to be the single function you need when working
+#' with R7 generics or R7 classes. This means that as well as registering
+#' methods for R7 classes on R7 generics, you can also register methods for
+#' R7 classes on S3 or S4 generics, and S3 or S4 classes for R7 generics.
+#' But this is not a general method registration function: at least one of
+#' `generic` and `signature` needs to be from R7.
```
DavisVaughan, line 13:
```suggestion
#' But this is not a general method registration function: at least one of
#' `generic` or `signature` needs to be from R7.
```

R/method.R
```diff
@@ -1,145 +1,163 @@
...
-  if(!is.environment(x)) {
-    return(x)
+  if (is_external_generic(generic)) {
+    if (!is.null(package)) {
+      # method registration within package, so add to lazy registry
+      external_methods_add(package, generic, signature, method)
+    } else {
+      # otherwise find the generic and register
+      generic <- getFromNamespace(generic$name, asNamespace(generic$package))
+      register_method(generic, signature, method)
```
DavisVaughan, line 63: I think the fact that `package` was always coming out as `"R7"` (mentioned over Slack) implies that we don't have a test for this branch?

R/method.R
```diff
@@ -1,145 +1,163 @@
...
+    }
+  } else if (is_s3_generic(generic)) {
+    generic_name <- attr(generic, "name")
+
+    if (length(signature) != 1 || class_type(signature[[1]]) != "r7") {
+      msg <- sprintf(
+        "When registering methods for S3 generic %s(), signature be a single R7 class",
```
DavisVaughan, line 70:
```suggestion
        "When registering methods for S3 generic %s(), `signature` must be a single R7 class",
```

R/method.R
```diff
@@ -1,145 +1,163 @@
...
+        generic_name
+      )
+      stop(msg, call. = FALSE)
+    }
+    class <- signature[[1]]@name
+    registerS3method(generic_name, class, method, envir = parent.frame())
```
DavisVaughan, line 76: I am not sure if the use of `parent.frame()` here will always be right or not? Should this function have an `envir` argument so we can always pass an environment through as needed?

R/method.R
```diff
@@ -1,145 +1,163 @@
...
-  if (inherits(generic, "R7_generic")) {
-    args <- generic@dispatch_args
-    generic <- generic@name
-  } else {
-    args <- setdiff(names(formals(generic)), "...")
-    generic <- find_function_name(generic, topenv(environment(generic)))
-  }
+  # Create grid of indices
+  indx <- lapply(signature, seq_along)
+  comb <- as.matrix(rev(do.call("expand.grid", rev(indx))))
```
DavisVaughan, line 106: Inlining a base R version of `tidyr::expand_grid()` to use on `indx` would also be pretty straightforward for this.

If you keep it this way, it might help to add a comment stating that the point is to vary the first column the slowest, unlike `expand.grid()`

R/method.R
```diff
@@ -1,145 +1,163 @@
...
-    out <- getS3method(generic, s3_class_name(signature[[1]]), optional = TRUE)
-    if (!is.null(out)) {
-      return(out)
-    }
+  rows <- lapply(1:nrow(comb), function(i) comb[i, ])
+  lapply(rows, function(row) Map("[[", signature, row))
+}
+
+as_generic <- function(x) {
+  if (inherits(x, "R7_generic") || is_external_generic(x)) {
```
DavisVaughan, line 114: Seems like `is_generic(x) inherits(x, "R7_generic")` would probably be worth it

R/method.R
```diff
@@ -1,145 +1,163 @@
...
-    return(out)
+  if (!is.function(x)) {
+    msg <- sprintf("`generic` must be a function, not a %s", obj_desc(x))
+    stop(msg, call. = FALSE)
   }
 
-  method_lookup_error(generic, args, signature)
+  # For now, assume that it's an S3 generic
+  attr(x, "name") <- find_generic_name(x)
+  class(x) <- "R7_S3_generic"
```
DavisVaughan, line 125: I feel like `structure(list(generic = x, name = name), class = "R7_S3_generic")` may be a little more structured? It feels a little wrong to me to add attributes and classes directly to the S3 generic

src/dispatch.c
```diff
@@ -14,7 +14,8 @@ Rboolean should_ignore(SEXP value, SEXP ignore) {
   return FALSE;
 }
 
-SEXP method_internal(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ignore) {
+// Recursively walk through method table to perform iterated dispatch
+SEXP method_rec(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ignore) {
```
DavisVaughan, line 18: I feel like I will always forget that `rec` is short for recurse, could it just be `method_recurse()`? Same comment about the R function that has a similar name

src/dispatch.c
```diff
@@ -48,6 +49,35 @@ SEXP method_internal(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ig
...
+  SEXP ns = Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("R7"));
+
+  if (R7_method_lookup_error_fun == NULL) {
+    R7_method_lookup_error_fun = Rf_findVarInFrame(ns, Rf_install("method_lookup_error"));
+  }
+  SEXP name = Rf_getAttrib(generic, Rf_install("name"));
+  SEXP args = Rf_getAttrib(generic, Rf_install("dispatch_args"));
+  SEXP R7_method_lookup_error_call = PROTECT(Rf_lang4(R7_method_lookup_error_fun, name, args, signature));
+  Rf_eval(R7_method_lookup_error_call, ns);
+}
```
DavisVaughan, line 63: I'm not sure if this will yell at you or not if you run it through rchk. I know we do similar things in rlang/vctrs but I think we also tag those functions with `__attribute__ ((noreturn))`

https://github.com/r-lib/rlang/blob/6e9ef5ee1aaab20dc9d2a04b7a3be043e869f4fc/src/rlang/rlang-types.h#L21
https://github.com/r-lib/rlang/blob/570dde18f2aa3a5817206acecbe9c56d4dddbea0/src/rlang/cnd.c#L73

src/dispatch.c
```diff
@@ -48,6 +49,35 @@ SEXP method_internal(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ig
...
+
+/* TODO: handle errors when method is not found */
```
DavisVaughan, line 65: I think you do this now?

DavisVaughan (approved):

R/dispatch.R
```diff
@@ -1,12 +1,77 @@
...
+method <- function(generic, signature) {
+  if (!inherits(generic, "R7_generic")) {
+    stop("`generic` must be an <R7_generic>")
+  }
+
+  signature <- as_signature(signature)
+  is_union <- vlapply(signature, is_union)
+  if (any(is_union)) {
+    stop("Can't dispatch on unions; must be a concrete type")
+  }
```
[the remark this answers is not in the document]
hadley: I tried that much earlier; the problem is that (e.g.) method dispatch and subclassing have different constraints, so `as_class()` starts to accumulate a bunch of different arguments. So I think it's simpler to constrain `as_class()` results as needed.

R/external-generic.R
```diff
@@ -16,43 +16,90 @@
...
+  invisible(x)
+}
+
+is_external_generic <- function(x) {
+  inherits(x, "R7_external_generic")
+}
+
 #' @importFrom utils getFromNamespace packageName
 #' @rdname new_external_generic
+#' @param package Package name. Advanced use only.
```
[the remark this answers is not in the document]
hadley: Oops, remnant of earlier attempt to fix the build failure.

R/external-generic.R
```diff
@@ -16,43 +16,90 @@
...
 #' @export
 method_register <- function() {
```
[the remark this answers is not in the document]
hadley: Yeah, and it's super confusing to have `register_method()` and `method_register()`. 

I'm not sure `external_methods_register()` is quite the right name because it feels like it's the generic that's external, not the methods, but it's definitely an improvement and we can iterate again in the future.

PR #161, opened 2022-02-12
Add OO type to class/object description

### LINKED ISSUE #133: How to display class names?
It would be good to have a consistent way to display class names in user facing messages. I think there are two questions:

* What symbol do we want to use to make it clear we're referring to a class? So far I've been using `<class>`
* Do we also want to show the system name? e.g. `S3<data.frame>`, `S4<range>`? If so, how?

hadley:
S3 and S4 are always included. R7 is included for object descriptions, but not class descriptions, since the R7 is, I think, reasonablly obviously implied. Base types don't a prefix because they should be easily recognised.

Fixes #133


hadley:
From discussion at January meeting, don't need to show R7, since it's typically implied.  Should also remove duplication from properties error.

PR #166, opened 2022-02-12
Make dispatch_args second argument of new_generic()

### LINKED ISSUE #164: Switch argument order of new_generic() again
Since it should now be fairly rare that you have to create your own generic body.

hadley:
Since most of the time, it's all you need. Fixes #164


lawremi (approved):
Seems pragmatic, and probably a welcome change from S4.

Issue #163, opened 2022-02-12
Review `object_class()` / `r7_class_name()`


hadley:
These are a pair of functions responsible for generating the character vectors used to index into the method table, for method registration (single class) and method dispatch (all classes). I think they need to:

* Add a prefix for OO type (to disambiguate between S3, S4, and R7 classes with the same name)
* Include the package name for S4 classes, where available

hadley:
Probably rename to:

* `object_class()`+ part of `method_rec()` -> `class_dispatch()`
* `r7_class_name()` -> `class_register()`

PR #172, opened 2022-02-15
Refactor class "names"

hadley:
Make it more clear how we compute the character vectors of class names that are used to register and dispatch methods.

* `class_register()` provides the single class name used when registering a method
* `obj_dispatch()` provides the complete vector of class names from an object. This is iterated over to find the matching method.
* `class_dispatch()` provides the complete vector of class names from an _class_. This is only used for method introspection.

There's no `obj_register()` because method registration always works with classes.

Part of #163.


src/dispatch.c
```diff
@@ -92,42 +69,19 @@ SEXP method_(SEXP generic, SEXP signature, SEXP ignore) {
...
-  if (Rf_inherits(object, "R7_object")) {
-    return Rf_getAttrib(object, Rf_install("object_class"));
+  static SEXP obj_dispatch_fun = NULL;
+  if (obj_dispatch_fun == NULL) {
+    obj_dispatch_fun = Rf_findVarInFrame(ns, Rf_install("obj_dispatch"));
   }
 
-  SEXP klass = get_class(object, envir);
+  SEXP methods_extends_call = PROTECT(Rf_lang2(obj_dispatch_fun, object));
+  SEXP res = Rf_eval(methods_extends_call, ns);
```
DavisVaughan, line 81:
```suggestion
  SEXP obj_dispatch_call = PROTECT(Rf_lang2(obj_dispatch_fun, object));
  SEXP res = Rf_eval(obj_dispatch_call, ns);
```

src/dispatch.c
```diff
@@ -92,42 +69,19 @@ SEXP method_(SEXP generic, SEXP signature, SEXP ignore) {
...
+  UNPROTECT(1);
```
DavisVaughan, line 82: In the future, you can probably pre-allocate `obj_dispatch_call` at package load time, and then all you'd have to do it set the `object` in the second slot of the call object and `Rf_eval()` it.

You could probably also pre-store the `ns` at package load time as well.

Probably a little premature for all that now though

src/dispatch.c
```diff
@@ -22,40 +22,14 @@ SEXP method_rec(SEXP table, SEXP signature, R_xlen_t signature_itr, SEXP ignore)
...
-      SEXP klass = Rf_install(CHAR(STRING_ELT(classes, i)));
-      SEXP val = Rf_findVarInFrame(table, klass);
-      if (TYPEOF(val) == ENVSXP) {
-        val = method_rec(val, signature, signature_itr + 1, ignore);
-      }
-      if (TYPEOF(val) == CLOSXP && (ignore == R_NilValue || !should_ignore(val, ignore))) {
-        return val;
-      }
+    if (TYPEOF(val) == CLOSXP && (ignore == R_NilValue || !should_ignore(val, ignore))) {
+      return val;
```
DavisVaughan, line 32: This was a nice cleanup

tests/testthat/test-class-spec.R
```diff
@@ -49,14 +54,58 @@ test_that("can work with S4 classes", {
...
+})
+
+test_that("can work with S4 subclasses", {
+  methods::setClass("Foo1", slots = c(start = "numeric", end = "numeric"))
+  methods::setClass("Foo2", contains = "Foo1")
+  klass <- methods::getClass("Foo2")
+
+  expect_equal(class_type(klass), "s4")
+  expect_equal(class_dispatch(klass), c("Foo2", "Foo1"))
+  expect_equal(class_register(klass), "Foo2")
```
DavisVaughan, line 77: Are there cases where `class_register(x) != class_dispatch(x)[[1]]`? Besides the `NULL` case? I was looking for a test where that was the case.

tests/testthat/test-class-spec.R
```diff
@@ -81,34 +130,38 @@ test_that("can work with simple S3 classes", {
...
     class = c("ordered", "factor"),
     constructor = function(.data = numeric(), levels) ordered(.data, levels)
   )
   expect_equal(as_class(klass), klass)
 
-  expect_equal(class_type(klass), "s3")
-  expect_equal(class_names(klass), c("R7_object", "ordered", "factor"))
+  expect_equal(class_type(klass), "r7_s3")
+  expect_equal(class_dispatch(klass), c("R7_object", "ordered", "factor"))
+  expect_equal(class_register(klass), "ordered")
```
DavisVaughan, line 156: Oooooh interesting I think I answered my own question from above.

And if you made an R7 subclass of ordered, then this looks good too

``` r
klass <- s3_class(
  class = c("ordered", "factor"),
  constructor = function(.data = numeric(), levels) ordered(.data, levels)
)

subklass <- new_class("ordered2", parent = klass)

class_dispatch(subklass)
#> [1] "ordered2"  "R7_object" "ordered"   "factor"
class_register(subklass)
#> [1] "ordered2"

obj <- subklass(1L, "x")
obj_dispatch(obj)
#> [1] "ordered2"  "R7_object" "ordered"   "factor"
```

PR #173, opened 2022-02-15
Tweaks to S3 class

hadley:


R/S3.R
```diff
@@ -112,14 +113,14 @@ check_constructor <- function(constructor) {
   }
 }
 
-is_S3_class <- function(x) {
-  inherits(x, "R7_S3_class")
+is_new_S3_class <- function(x) {
+  inherits(x, "R7_new_S3_class")
 }
```
DavisVaughan, line 118: `is_S3_class()` should stay here too

DavisVaughan (commented):
Left some general comments mainly related to what I think are copy paste issues. Still not sure how I feel about this.

In the docs for `?S3_class` you mention two separate uses for `S3_class()`:

> Method dispatch, properties, and unions

> Extending an S3 class

Maybe the first could use a simple "stub" for an S3 class, like `S3("factor")` which is just a signaler for the type of S3 class that you want to register for / allow as a property. It would have no other arguments and only allow a single string.

And when you extend an S3 class, you need to create a proper one with `new_S3_class()`, which has the other constructor/validator arguments. The `constructor` becomes required at this point.

Mainly it just feels like these have separate responsibilities somehow, which is apparent in the docs but not as apparent in the implementation

vignettes/case_studies.Rmd
```diff
@@ -103,13 +103,11 @@ In particular the number of rows of the `rows` property matches the number of ro
...
-tbl_df <- S3_class(c("tbl_df", "tbl", "data.frame"))
-
 annotated_df <- new_class("annotated_df",
   properties = list(
-    data = tbl_df,
-    rows = tbl_df,
-    columns = tbl_df
+    data = new_S3_class("tbl_df"),
+    rows = new_S3_class("tbl_df"),
+    columns = new_S3_class("tbl_df")
```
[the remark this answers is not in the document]
hadley: Yes, because `inherits(x, "tbl_df")` would be true.

hadley:
I get where you're coming from, but adding another function (with a new naming convention) doesn't currently feel worth it to me. I'll removed the restriction that `class` has to be a string, and replaced it with a suggestion in the docs. This means that there are three ways of using `new_S3_class()`:

* `new_S3_class("POSIXct")` quick and dirty for method registration
* `new_S3_class(c("POSIXct", "POSIXt"))`, full class declaration so manual introspection also works
* `new_S3_class(c("POSIXct", "POSIXt")), construction = ...)`, full ability to extend.

That feels ok to me.

R/S3.R
```diff
@@ -58,7 +58,9 @@
 #' ```
 #'
 #' @export
-#' @param class Character vector of S3 classes
+#' @param class S3 class vector (i.e. the what `class()` return). For method
```
DavisVaughan, line 61:
```suggestion
#' @param class S3 class vector (i.e. what `class()` returns). For method
```
```suggestion
#' @param class S3 class vector (i.e. the what `class()` return). For method
```

PR #174, opened 2022-02-16
Don't use R7 for base and S3 wrappers

hadley:
I think this makes things simpler overall. Because there's less bootstrapping overall, it also makes it a bit easier to iterate since it's harder to break things to the point where the package no longer loads.


R/class.R
```diff
@@ -5,11 +5,12 @@
 #' when an object is passed to a generic.
 #'
 #' @param name The name of the class, as a string.
-#' @param parent The parent class.
+#' @param parent The parent class to inherit behavior from. All R7 objects
+#'   eventually inherit from There are three options:
```
DavisVaughan, line 9:
```suggestion
#'   eventually inherit from [R7_object]. There are three options:
```

R/class.R
```diff
@@ -5,11 +5,12 @@
...
 #'
-#'   * To inherit behaviour from an R7 class, pass the class object.
-#'   * To inherit behaviour from a base type, pass the function you'd use
-#'     to construct the object, e.g. `character`, `integer`.
+#'   * The R7 base class, [R7_object].
+#'   * A S3 class wrapped by [S3_class()].
```
DavisVaughan, line 12:
```suggestion
#'   * An S3 class wrapped by [new_S3_class()].
```

tests/testthat/_snaps/S3.md
```diff
@@ -3,7 +3,7 @@
     Code
       S3_class(c("ordered", "factor"))
     Output
-      S3 class <ordered/factor>
+      <R7_S3_class>: S3<ordered>
```
DavisVaughan, line 6: It seems to use `class_desc()` now. So we don't show all classes anymore?

DavisVaughan (approved):

tests/testthat/_snaps/S3.md
hadley: What do you think? I don't have any strong feelings either way except that it seems better to be consistent? (i.e. any change should be to `class_desc()`)

PR #175, opened 2022-02-16
Compatibility vignette

### LINKED ISSUE #53: Be precise about compatibility
I think these are the 12 possible types of compatibility.

Inheritance:

* R7 extends S3 object: yes
* R7 extends S4 object: no?
* S3 extends R7 object: yes, to support incremental replacement of base classes.
* S4 extends R7 object: yes, via S3.

Composition:

* R7 has S3 property: yes
* R7 has S4 property: no
* S3 has R7 attribute: yes, because no restrictions
* S4 has R7 slot: yes, via S3

Methods:

* R7 generic + S3 object: yes
* R7 generic + S4 object: no?
* S3 generic + R7 object: yes
* S4 generic + R7 object: yes, via S3 

---

I think we can get away with not supporting S4 in R7 as long as we assume that any S4 -> R7 conversion proceeds from the inside out (i.e. start with the base class and it propagates out over time)

hadley:
Fixes #53

### DIFF
--- a/vignettes/compatibility.Rmd
+++ b/vignettes/compatibility.Rmd
@@ -1,0 +1,70 @@
+---
+title: "Compatibility"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Compatibility}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+R7 is designed to be compatible with S3 and S4. This vignette discusses the details.
+
+```{r setup}
+library(R7)
+```
+
+## S3
+
+* R7 can register methods for:
+  * R7 class and S3 generic
+  * S3 class and R7 generic
+* R7 classes can extend S3 classes
+* S3 classes can extend R7 classes
+
+### Methods
+
+`method()` is designed to be the single tool for method registration that you need when working with R7 classes. You can also register a method for an R7 class and S3 generic without using R7, because all R7 objects have S3 classes, and S3 dispatch will operate on them normally.
+
+```{r}
+foo <- new_class("foo")
+class(foo())
+
+mean.foo <- function(x, ...) {
+  "mean of foo"
+}
+mean(foo)
+```
+
+### Classes
+
+It's also possible to extend an R7 class with S3, but I'm not sure that it's ever worth it.
+
+## S4
+
+* R7 classes can not extend S4 classes 
+* S4 classes can extend S3 classes
+* R7 can register methods for:
+  * R7 class and S4 generic
+  * S4 class and R7 generic
+
+### Unions
+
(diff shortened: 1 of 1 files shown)


vignettes/compatibility.Rmd
```diff
@@ -0,0 +1,70 @@
...
+
+* R7 classes can not extend S4 classes 
+* S4 classes can extend S3 classes
+* R7 can register methods for:
+  * R7 class and S4 generic
+  * S4 class and R7 generic
+
+### Unions
+
+S4 unions are automatically converted to R7 unions. There's an important difference in the way that class unions are handled in S4 and R7. In S4, they're handled at method dispatch time, so when you create `setUnion("u1", c("class1", "class2"))`, `class1` and `class2` now extend `u1`. In R7, unions are handled at method registration time so that registering a method for a union is just short-hand for registering a method for each of the classes.
```
lawremi, line 59: Worth clarifying that R7 unions can also be used for properties (just like S4)?

vignettes/compatibility.Rmd
```diff
@@ -0,0 +1,70 @@
...
+)
+```
+
+R7 is designed to be compatible with S3 and S4. This vignette discusses the details.
+
+```{r setup}
+library(R7)
+```
+
+## S3
```
lawremi, line 23: Probably worth describing how R7 is just an extension of S3 (objects have a class vector), so that a lot of this "just works".

vignettes/compatibility.Rmd
```diff
@@ -0,0 +1,70 @@
...
+
+It's also possible to extend an R7 class with S3, but I'm not sure that it's ever worth it.
+
+## S4
+
+* R7 classes can not extend S4 classes 
+* S4 classes can extend S3 classes
+* R7 can register methods for:
+  * R7 class and S4 generic
+  * S4 class and R7 generic
```
lawremi, line 55: What about an S4 class and an S4 generic? Towards the goal of creating a single method registration API.

vignettes/compatibility.Rmd
```diff
@@ -0,0 +1,70 @@
...
+```
+
+### Classes
+
+It's also possible to extend an R7 class with S3, but I'm not sure that it's ever worth it.
+
+## S4
+
+* R7 classes can not extend S4 classes 
+* S4 classes can extend S3 classes
```
lawremi, line 52: Does this imply that S4 classes can extend R7 classes?

vignettes/compatibility.Rmd
```diff
@@ -0,0 +1,70 @@
...
+mean(foo)
+```
+
+### Classes
+
+It's also possible to extend an R7 class with S3, but I'm not sure that it's ever worth it.
+
+## S4
+
+* R7 classes can not extend S4 classes
```
lawremi, line 51: Have we decided this is infeasible? Might be possible by creating an "old class" for the R7 class, but it's not clear if an "old class" can extend an ordinary S4 class.

hadley:
Moving into #187 since it's easier to tackle all the docs in one place.

PR #177, opened 2022-02-17
Implement missing_class and any_class

### LINKED ISSUE #67: Dispatching on "missing"
S4 lets us dispatch on "missing", i.e., an argument in the signature not being provided. This is a somewhat useful feature for multiple dispatch. Should/could we support that in R7?

hadley:
Fixes #67

You might be surprised that there are no changes to `method-dispatch.c`; that's because it was already using the MISSING placeholder.


R/class-spec.R
```diff
@@ -120,12 +140,16 @@ class_desc <- function(x) {
...
-    S4 = S4_strip_union(methods::extends(x)),
+    NULL = c("NULL", "ANY"),
+    missing = "MISSING",
+    any = "ANY",
+    S4 = c(S4_strip_union(methods::extends(x)), "ANY"),
     R7 = c(x@name, class_dispatch(x@parent)),
-    R7_base = c("R7_object", x$class),
-    R7_S3 = c("R7_object", x$class),
+    R7_base = c("R7_object", x$class, "ANY"),
+    R7_S3 = c("R7_object", x$class, "ANY"),
```
DavisVaughan, line 152: Sort of a strange sandwiching of the base/S3 classes, but I guess this makes sense for dispatch

R/class-spec.R
```diff
@@ -120,12 +140,16 @@ class_desc <- function(x) {
 
 # Vector of class names; used in method introspection
 class_dispatch <- function(x) {
+  if (identical(x, R7_object)) return(c("R7_object", "ANY"))
```
DavisVaughan, line 143: So should the `@parent` of R7_object be `any_class`? Or maybe you can't use `any_class` as a parent in general

R/class-spec.R
```diff
@@ -159,7 +187,9 @@ class_deparse <- function(x) {
 
 class_inherits <- function(x, what) {
   switch(class_type(what),
-    NULL = TRUE,
+    "NULL" = is.null(x),
+    missing = FALSE,
```
DavisVaughan, line 191: `class_inherits(missing_class, missing_class) == FALSE`?

R/object.R
```diff
@@ -36,7 +36,7 @@ new_object <- function(.data = NULL, ...) {
   }
 
   attr(object, "object_class") <- obj_cls
-  class(object) <- class_dispatch(obj_cls)
+  class(object) <- setdiff(class_dispatch(obj_cls), "ANY")
```
DavisVaughan, line 39: I see you had to do this exact pattern at least twice. Mainly just pointing that out, not sure if a different named helper is needed or not

R/special.R
```diff
@@ -0,0 +1,47 @@
+#' Special dispatch types
+#'
+#' * Use `missing_class` when you the user has not supplied an argument
```
DavisVaughan, line 3:
```suggestion
#' * Use `missing_class` when the user has not supplied an argument
```

I think you didn't want this

tests/testthat/_snaps/method-register-dispatch.md
```diff
@@ -4,19 +4,19 @@
 
 # method lookup fails with informative messages
 
-    Can't find method for generic `foo()` with classes:
-    - x: <logical>
-    - y: <MISSING>
+    Can't find method for generic `foo()` with dispatch classes:
+    - x: logical, ANY
+    - y: MISSING
```
DavisVaughan, line 9: I wonder if _always_ seeing ANY print is going to be annoying. Like, should we just assume it is implied?

tests/testthat/test-property.R
```diff
@@ -167,7 +167,8 @@ test_that("properties can be base, S3, S4, R7, or R7 union", {
 
   my_class <- new_class("my_class",
     properties = list(
-      anything = NULL,
+      anything = any_class,
+      null = NULL,
```
DavisVaughan, line 171: `missing = missing_class`? for completeness?

tests/testthat/test-property.R
```diff
@@ -178,6 +179,7 @@ test_that("properties can be base, S3, S4, R7, or R7 union", {
   expect_snapshot(my_class)
   my_obj <- my_class(
     anything = TRUE,
```
DavisVaughan, line 181: if you use `missing_arg()` as an object I guess it probably doesn't print as missing?

so maybe for this example you just wouldn't supply that argument, and it would still be captured in the snapshot?

DavisVaughan (approved):
I like the general idea, I think my only minor concern is about how aggressively we want to be showing users the ANY class in errors

R/class-spec.R
hadley: I think this is actually a sign that `class_dispatch()` needs to be refactored out into two functions, because this is the base case when recursing to get the class for an R7 object. But I don't have good name for it, and this will all get rewritten in C at some point.

R/class-spec.R
hadley: Yeah, but I don't _think_ that matters

tests/testthat/_snaps/method-register-dispatch.md
hadley: I think this is ok for now; it'll get fixed when we eventually print the original method here.

tests/testthat/test-property.R
hadley: But then there's no way to create the object 😄

PR #178, opened 2022-02-17
Add default values to properties

### LINKED ISSUE #139: Should properties have a default value?
This would be used by the default constructor when the property is not supplied. 

If properties can have a default value, should it be supplied automatically? i.e. if the property is an R7 or S4 class, it could call the default constructor. This would have the advantage that you can always easily initialise a class>

hadley:
Fixes #139


R/S3.R
```diff
@@ -66,6 +66,9 @@
 #'   have an R7 class inherit from an S3 class. It must be specified in the
 #'   same way as a R7 constructor: the first argument should be `.data`
 #'   (the base type whose attributes will be modified).
```
DavisVaughan, line 68: I know you mention the no argument case below, but it may be helpful to explicitly advise here that `.data` should have a default value

R/object.R
```diff
@@ -29,6 +29,11 @@ new_object <- function(.data = NULL, ...) {
     )
   }
 
+  missing_props <- nms[vlapply(args, is.null)]
+  for(prop in missing_props) {
+    args[[prop]] <- prop_default(obj_cls@properties[[prop]])
+  }
```
DavisVaughan, line 35: Does this limit you from setting a property to `NULL` at all? Since now `NULL` also means "go get me the default"?

DavisVaughan (approved):

vignettes/case_studies.Rmd
```diff
@@ -32,7 +32,8 @@ Person <- new_class("Person",
       setter = function(object, value) {
         object@birthdate <- as.Date(value)
         object
-      }
+      },
+      default = as.Date(NA)
```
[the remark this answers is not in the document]
hadley: If you did want that, you could always override the constructor. But I do think it should be pretty rare.

R/object.R
hadley: Oh yeah, good catch. I've switched sentinel value to `missing_class`, which while isn't exactly the same meaning seems close enough, and avoids the challenges of working with `missing()`.

PR #181, opened 2022-02-21
Implement convert() and super()

### LINKED ISSUE #110: Can we eliminate `next_method()`?

### LINKED ISSUE #136: What should coercion look like for R7?

hadley:
Fixes #110. Fixes #136


hadley:
@lawremi points out that the main feature of `up_cast()` is not performance, but that it only affects the dispatch of a single method.

hadley:
https://github.com/RConsortium/OOP-WG/pull/181/files#diff-a68deddd00e9d282550d4445b10c54406e9a17fc9d44e0968ebd44bcc7248080R48-R62 illustrates the difference between `cast()` and `cast_next()`.

hadley:
Another possible name for `cast_next()` would be `super()`

hadley:
Renamed to `super()`; I think that makes it more clear that these are related, but distinct, ideas.

_pkgdown.yml
```diff
@@ -21,6 +21,8 @@ reference:
 
 - title: Method dispatch
   contents:
+  - cast
+  - cast_next
```
DavisVaughan:
```suggestion
  - super
```

src/method-dispatch.c
```diff
@@ -5,17 +5,8 @@
 extern SEXP parent_sym;
 extern SEXP name_sym;
 
-Rboolean should_ignore(SEXP value, SEXP ignore) {
```
DavisVaughan, line 8: What did all of this do previously?

tests/testthat/test-super.R
```diff
@@ -0,0 +1,39 @@
...
+  it("overrides dispatch, matching exact class", {
+    foo1 <- new_class("foo1")
+    foo2 <- new_class("foo2", foo1)
+    foo3 <- new_class("foo3", foo2)
+
+    bar <- new_generic("bar", "x")
+    method(bar, foo1) <- function(x) 1
+    method(bar, foo3) <- function(x) 3
+
+    expect_error(bar(super(foo3())), "Can't find method")
```
DavisVaughan, line 12: I feel like this is inconsistent

``` r
foo1 <- new_class("foo1")
foo2 <- new_class("foo2", foo1)
foo3 <- new_class("foo3", foo2)

bar <- new_generic("bar", "x")
method(bar, foo1) <- function(x) 1
method(bar, foo3) <- function(x) 3

# this doesnt work
bar(super(foo3()))
#> Error: Can't find method for generic `bar()` with dispatch classes:
#> - x: foo2

# but this does
bar(foo2())
#> [1] 1
```

ltierney:
Unfortunately I haven't been able to put the time into keeping up
with things, so maybe this is all irrelevent -- feel free to
ignore if it is.

I'm worried about the idea of using casting for delegating method
handling up the disptch chain. In other seetings where I've seen
this tried it looks OK at first but runs into trouble with
slightly more complex settings. Two simple scenarios:

1) A base class A, B a subclass of A, C a subclass of B.
f has methods f.A, f.B, and f.C with

f.C <- function(x) next_method(x)  ## delegate to f.B
f.B <- function(x) next_method(x)  ## delegate to f.A

So f(<a C>) should be handled by f.A.

2) g is another generic with methods g.A and g.B. The method f.A ia
defined as

f.A <- function(x) g(x)

So in f(<a B>) the call in f.A() to g() should be handled by g.B(),
not g.A().

Are these going to be handled properly with these casting approaches?

-- 
Luke Tierney
Ralph E. Wareham Professor of Mathematical Sciences
University of Iowa                  Phone:             319-335-3386
Department of Statistics and        Fax:               319-335-3017
    Actuarial Science
241 Schaeffer Hall                  email:   ***@***.***
Iowa City, IA 52242                 WWW:  http://www.stat.uiowa.edu

tests/testthat/test-cast.R
```diff
@@ -0,0 +1,32 @@
+test_that("can register cast methods", {
+  casttest <- new_class("casttest")
+  method(cast, list(casttest, character)) <- function(from, to, ...) "c"
+  method(cast, list(casttest, integer)) <- function(from, to, ...) "i"
```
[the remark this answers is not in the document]
hadley: I don't think so; an object of class `x` already inherits from `x` so this just works. (And I added a test to verify).

hadley:
@ltierney no problems; the development has been fairly hectic and I've been down a few dead ends. I'd no longer characterise `super()` as a casting based — it creates a transient wrapper that only affects dispatch for the next generic. I think this resolves your concerns:

``` r
library(R7)
A <- new_class("A")
B <- new_class("B", A)
C <- new_class("C", B)

f <- new_generic("f", "x")
# You (now) have to supply the class you want to dispatch
# to, so there is no longer any ambiguity:
method(f, C) <- function(x) c("f-C", f(super(x, B)))
method(f, B) <- function(x) c("f-B", f(super(x, A)))
method(f, A) <- function(x) "f-A"

f(C())
#> [1] "f-C" "f-B" "f-A"

method(f, A) <- function(x) c("f-A", g(x))

# super() only affects dispatch for the next generic, so
# future generics used by the method dispatch on the original class
g <- new_generic("g", "x")
method(g, C) <- function(x) c("g-C", g(super(x, B)))
method(g, B) <- function(x) c("g-B", g(super(x, A)))
method(g, A) <- function(x) "g-A"

f(B())
#> [1] "f-B" "f-A" "g-B" "g-A"
```

<sup>Created on 2022-02-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

tests/testthat/test-super.R
DavisVaughan: Do you think we could talk about this one in the next OOP working group meeting?

To me it feels like `super(foo3, foo2)` says "I return an object of class foo2", meaning that `bar(foo2())` is identical to `bar(super(foo3(), foo2))` in my head

I am interpreting your argument as: If the user wanted the `foo1` behavior, then they should have done `super(foo3, foo1)`, which also feels reasonable (and more explicit). But it still might be nice to discuss this point.

hadley:
We should also consider if `cast()` is the right name. `convert()` might be more likely to evoke the right sense.

ltierney:
Thanks -- that makes sense. So from a user perspective in a call

   f(super(x, B))

super() is just (an optional) part of the syntax of calling a generic,
not an actual function that could be used outside a call to a generic.
Whether it creates an ephemeral wrapper or communicates in some other
way to the dispatch process is part of the implementation, not
something user visible.  If I'm understanding this correctly ...

On Wed, 23 Feb 2022, Hadley Wickham wrote:

> 
> @ltierney no problems; the development has been fairly hectic and I've been
> down a few dead ends. I'd no longer characterise super() as a casting based
> — it creates a transient wrapper that only affects dispatch for the next
> generic. I think this resolves your concerns:
> 
> library(R7)
> A <- new_class("A")
> B <- new_class("B", A)
> C <- new_class("C", B)
> 
> f <- new_generic("f", "x")
> # You (now) have to supply the class you want to dispatch
> # to, so there is no longer any ambiguity:
> method(f, C) <- function(x) c("f-C", f(super(x, B)))
> method(f, B) <- function(x) c("f-B", f(super(x, A)))
> method(f, A) <- function(x) "f-A"
> 
> f(C())
> #> [1] "f-C" "f-B" "f-A"
> 
> method(f, A) <- function(x) c("f-A", g(x))
> 
> # super() only affects dispatch for the next generic, so
> # future generics used by the method dispatch on the original class
> g <- new_generic("g", "x")
> method(g, C) <- function(x) c("g-C", g(super(x, B)))
> method(g, B) <- function(x) c("g-B", g(super(x, A)))
> method(g, A) <- function(x) "g-A"
> 
> f(B())
> #> [1] "f-B" "f-A" "g-B" "g-A"
> 
> Created on 2022-02-23 by the reprex package (v2.0.1)
> 
> —
> Reply to this email directly, view it on GitHub, or unsubscribe.
> Triage notifications on the go with GitHub Mobile for iOS or Android.
> You are receiving this because you werementioned.[AA55UVHV2FQQ6XVK7ARV5X3U4TUPJA5CNFSM5O7M7H7KYY3PNVWWK3TUL52HS4DF
> VREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOH2B6HJY.gif] Message ID:
> ***@***.***>
> 
> 
>

-- 
Luke Tierney
Ralph E. Wareham Professor of Mathematical Sciences
University of Iowa                  Phone:             319-335-3386
Department of Statistics and        Fax:               319-335-3017
    Actuarial Science
241 Schaeffer Hall                  email:   ***@***.***
Iowa City, IA 52242                 WWW:  http://www.stat.uiowa.edu

hadley:
@ltierney Right. Currently `super()` does create a transient object, but that's just an implementation detail. I'll think about what we can do to make that more obvious: it at least needs a print method, and maybe `@` should give an informative error?

R/cast.R
```diff
@@ -0,0 +1,41 @@
...
+    cast(from, to, ...)
+  } else if (class_inherits(from, to)) {
+    if (is_base_class(to)) {
+      R7_data(from)
+    } else if (is_class(to)) {
+      attr(from, "object_class") <- to
+      class(from) <- setdiff(class_dispatch(to), "ANY")
+      from
+    } else if (is_S3_class(to)) {
+      class(from) <- to$class
```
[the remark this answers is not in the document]
hadley: Now that it's convert, it's much more clear to me that you're correct and we should strip the extra properties.

tests/testthat/test-super.R
hadley: BTW this is consistent with `convert()` if you do `convert(x, to = foo2)`, you don't expect to get an instance of `foo1` back.

DavisVaughan: But you do get something back that you can call `bar()` on, and that will dispatch to the `foo1` method?

```r
x <- foo3()

bar(x)
#> [1] 3

bar(convert(x, to = foo2))
#> [1] 1

bar(super(x, to = foo2))
#> Error: Can't find method for generic `bar()` with dispatch classes:
#> - x: foo2
```

hadley: Right. It's not directly analogous, but I think it's in the same spirit.

lawremi:
I like that `super()` is an actual function, rather than special syntax that requires special documentation.

R/super.R
```diff
@@ -0,0 +1,90 @@
...
+#' bar2 <- new_generic("bar2", "x")
+#' method(bar2, foo1) <- function(x) c(1, bar1(x))
+#' method(bar2, foo2) <- function(x) c(2, bar1(x))
+#'
+#' obj <- foo2(1, 2, 3)
+#' bar2(obj)
+#' # convert() affects every generic:
+#' bar2(convert(obj, to = foo1))
+#' # super() only affects the _next_ generic:
+#' bar2(super(obj, foo1))
```
hadley, line 56: Name `to` argument

tests/testthat/test-super.R
hadley: Ok, let's change this so that `super(..., to = foo2)` behaves exactly like a `foo2`

PR #183, opened 2022-02-22
Disambiguate dispatch

### LINKED ISSUE #48: Prefix class name with package
For backward compatibility with S3, we need to generate a character class vector. If we use just the class name for this vector, we are implicitly using the name as an identifier for S3 dispatch. Is this something we need to worry about?

In other words, are we worried that pkgA defines `newClass("A")` and pkgB defines `newClass("A")`, and while dispatch in the new system can disambiguate, S3 dispatch does not?

### LINKED ISSUE #163: Review `object_class()` / `r7_class_name()`

hadley:
I think this is the minimal set of changes to disambiguate between:

* S4 vs R7/S3/base cases: `S4/foo` vs `foo`
* S4 classes with same name in different packages: `S4/pkgA::foo` vs `S4/pkgB::foo`
* R7 classes with same name in difference packages: `pkgA::foo` vs `pkgB::foo`

I went back and forth a bit on the separator, but ultimately I think `::` makes the most sense as running `pkg::class` will typically get you the class.

Fixes #48. Fixes #163

### DIFF
--- a/R/S4.R
+++ b/R/S4.R
@@ -38,11 +38,13 @@ S4_class_dispatch <- function(x) {
   x <- methods::getClass(x)
   self <- S4_class_name(x)
 
-  # Find all extended classes, stripping self
+  # Find all extended classes
   extends <- unname(methods::extends(x, fullInfo = TRUE))
   extends <- Filter(function(x) methods::is(x, "SClassExtension"), extends)
 
   classes <- lapply(extends, function(x) methods::getClass(x@superClass))
+  # Remove virtual classes that aren't S3. This removes unions because R7
+  # handles them in method registration, not dispatch.
   classes <- Filter(function(x) !x@virtual || is_oldClass(x), classes)
 
   c(self, vcapply(classes, S4_class_name))
@@ -69,17 +71,10 @@ S4_class_name <- function(x) {
   }
 }
 
-# R7 handles unions at method registration time, where as S4 handles them at
-# dispatch time.
-S4_strip_union <- function(class_names) {
-  classes <- lapply(class_names, methods::getClass)
-  is_union <- vlapply(classes, methods::is, "ClassUnionRepresentation")
-
-  setdiff(class_names[!is_union], "vector")
-}
-
-S4_remove_classes <- function(classes, where = parent.frame()) {
+S4_remove_classes <- function(classes, where = globalenv()) {
   for (class in classes) {
     methods::removeClass(class, topenv(where))
   }
 }
+
+globalVariables(c("superClass", "virtual"))
--- a/R/S4.R
+++ b/R/S4.R
@@ -33,16 +33,15 @@ S4_to_R7_class <- function(x, error_base = "") {
   }
 }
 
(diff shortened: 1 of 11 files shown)


tests/testthat/test-S4.R
```diff
@@ -33,3 +36,75 @@ test_that("converts S4 representation of S3 classes to R7 representation", {
...
+
+  it("ignores unions", {
+    on.exit(S4_remove_classes(c("Foo1", "Foo2", "Foo3")))
+
+    setClass("Foo1", slots = list("x" = "numeric"), where = globalenv())
+    setClass("Foo2", slots = list("x" = "numeric"), where = globalenv())
+    setClassUnion("Foo3", c("Foo1", "Foo2"), where = globalenv())
+
+    expect_equal(S4_class_dispatch("Foo1"), "S4/Foo1")
+    expect_equal(S4_class_dispatch("Foo2"), "S4/Foo2")
```
DavisVaughan, line 93: What does `S4_class_dispatch("Foo3")` return? I'm assuming that is not allowed in some way, maybe the idea is that we shouldn't check that here because it is handled at registration time instead

tests/testthat/test-class-spec.R
```diff
@@ -16,6 +16,24 @@ test_that("can work with R7 classes", {
   expect_equal(class_inherits(obj, klass), TRUE)
 })
 
+test_that("can work with packages R7 classes", {
```
DavisVaughan, line 19:
```suggestion
test_that("can work with R7 classes in packages", {
```

?

DavisVaughan (approved):
Seems pretty nice. It sounds like you decided we don't ever want this `package` argument in `new_S3_class()`?

R/S4.R
```diff
@@ -33,11 +33,47 @@ S4_to_R7_class <- function(x, error_base = "") {
...
+
+  # Remove virtual classes that aren't S3. This removes unions because R7
+  # handles them in method registration, not dispatch.
+  classes <- Filter(function(x) !x@virtual || is_oldClass(x), classes)
+
+  c(self, vcapply(classes, S4_class_name))
+}
+
+is_oldClass <- function(x) {
+  x@virtual && methods::extends(x, "oldClass") && x@className != "oldClass"
```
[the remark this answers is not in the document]
hadley: Yeah, it's needed to avoid having an "oldClass" class; I don't know why.

tests/testthat/test-S4.R
hadley: Yeah, R7 will never see an instance of that type so we don't need to test it.

hadley:
Yeah, I don't think it makes sense to include it in `new_S3_class()` since S3 classes aren't really associated with packages.

PR #195, opened 2022-02-27
Implement method_explain()

### LINKED ISSUE #194: Function to explain dispatch
Like `method()` but would list all methods that are consulted in order, along with which ones are defined.

Use in dispatch vignette.

hadley:
Fixes #194


R/method-introspect.R
```diff
@@ -31,30 +32,86 @@
...
+#' method(add, list(foo1, foo1)) <- function(x, y) c(1, 1)
+#'
+#' method_explain(add, list(foo2, foo2))
+method_explain <- function(generic, class = NULL, object = NULL) {
+  dispatch <- as_dispatch(generic, class = class, object = object)
+
+  grid <- as.matrix(rev(do.call("expand.grid", rev(dispatch))))
+  colnames(grid) <- generic@dispatch_args
+
+  names <- paste0("[", grid, "]")
```
DavisVaughan, line 70: You like `[]` better than `<>`? It seems like we are printing types here, and `<>` felt more consistent

tests/testthat/test-method-introspect.R
```diff
@@ -32,3 +32,16 @@ describe("method introspection", {
...
+  it("shows all possible methods along with matches", {
+    foo1 <- new_class("foo1")
+    foo2 <- new_class("foo2", foo1)
+
+    add <- new_generic("add", c("x", "y"))
+    method(add, list(foo2, foo1)) <- function(x, y) c(2, 1)
+    method(add, list(foo1, foo1)) <- function(x, y) c(1, 1)
+
+    expect_snapshot_output(method_explain(add, list(foo2, foo2)))
+  })
```
DavisVaughan, line 46: Mayyybe worth a test showing that it "works" when there aren't any matching / existing methods at all. i.e. supplying `method_explain(add, list("integer", "double"))` here or something

DavisVaughan (approved):

DavisVaughan:
I assume https://github.com/RConsortium/OOP-WG/issues/197#issue-1154641071 will fix this, but we probably don't want to show possible methods for R7 unions and base R unions too

PR #204, opened 2022-03-08
Implement $ and $<- methods

### LINKED ISSUE #198: Provide $.R7_method
To give a better error (i.e. hide the fact that it's an S4 object)

hadley:
Fixes #198


R/zzz.R
```diff
@@ -16,6 +16,35 @@ R7_object <- new_class(
 )
 methods::setOldClass("R7_object")
 
+#' @export
+`$.R7_object` <- function(x, name) {
+  if (typeof(x) %in% c("list", "environment")) {
+    NextMethod()
+  } else {
+    msg <- sprintf(
+      "Can't access R7 properties with $. Did you mean %s@%s?",
```
DavisVaughan, line 25:
```suggestion
      "Can't access R7 properties with `$`. Did you mean `%s@%s`?",
```

?

R/zzz.R
```diff
@@ -16,6 +16,35 @@ R7_object <- new_class(
...
+    stop(msg, call. = FALSE)
+  }
+}
+#' @export
+`$<-.R7_object` <- function(x, name, value) {
+  if (typeof(x) %in% c("list", "environment")) {
+    NextMethod()
+  } else {
+    msg <- sprintf(
+      "Can't access R7 properties with $. Did you mean ...@%s <- %s?",
```
DavisVaughan, line 38:
```suggestion
      "Can't set R7 properties with `$<-`. Did you mean `...@%s <- %s`?",
```

^ Not sure how precise you want to be here

Does `deparse1(substitute(x))` give `*tmp*` or something like that here? I was somehow expecting that to be used too

R/zzz.R
```diff
@@ -16,6 +16,35 @@ R7_object <- new_class(
 )
 methods::setOldClass("R7_object")
 
+#' @export
+`$.R7_object` <- function(x, name) {
+  if (typeof(x) %in% c("list", "environment")) {
```
DavisVaughan, line 21: Do we care about:

```r
> x <- expression(foo = 1)
> typeof(x)
[1] "expression"
> x$foo
[1] 1
```

R/zzz.R
```diff
@@ -16,6 +16,35 @@ R7_object <- new_class(
 )
 methods::setOldClass("R7_object")
 
+#' @export
+`$.R7_object` <- function(x, name) {
```
DavisVaughan, line 20: Somewhat interesting to note that if:

```r
# rather than this, where the integer vector is the core data object
new_class("foo", parent = "integer")

# we instead used this as the implementation
# (where new_class("foo", parent = "integer") would basically do this instead)
new_class("foo", properties = list(data = "integer"))
```

Then we could drop `@` altogether and use `$` to access the properties, one of which would be `$data`.

I know you mentioned that we probably want a way to directly build on base atomic objects, so it probably is too late, but it is interesting to think about how `@` could be removed (that would probably make autocomplete work on properties out of the box too, probably through `.DollarNames`)

R/zzz.R
```diff
@@ -16,6 +16,35 @@ R7_object <- new_class(
...
+      "Can't access R7 properties with $. Did you mean %s@%s?",
+      deparse1(substitute(x)),
+      name
+    )
+    stop(msg, call. = FALSE)
+  }
+}
+#' @export
+`$<-.R7_object` <- function(x, name, value) {
+  if (typeof(x) %in% c("list", "environment")) {
```
DavisVaughan, line 34: Same question about expressions

R/zzz.R
DavisVaughan: I guess you still have to think about whether or not you want this with lists though

```
# if this:
new_class("foo", parent = "list")

# was basically this:
foo <- new_class("foo", properties = list(data = "list")

obj <- foo(list(x = 1))

# then you'd have to do this:
obj$data$x
```

R/zzz.R
hadley: Yeah, I tried it and (somewhat obviously) it does give `*tmp*` so it's not useful

R/zzz.R
hadley: I think the problem is then that `is.list()` would be TRUE for R7 objects which seems likely to lead to a bunch of unintended downstream consequences.

DavisVaughan: I was still thinking the core object would be that weird S4 gremlin thing that is returned by `R7_object()`, and we'd still have all the properties as attributes on that

PR #207, opened 2022-03-15
Consistently use `class_` prefix for base wrappers

### LINKED ISSUE #170: Consider eliminating string representation of base types
In favour of `base_integer`, `base_numeric`, etc.

Suggested by @lawremi in February meeting.

hadley:
Fixes #170 

@DavisVaughan I don't think you need to worry too much about the implementation as it should've been hard for me to screw it up, but a pass over the docs would be appreciated.


R/base.R
```diff
@@ -53,17 +53,85 @@ str.R7_base_class <- function(object, ..., nest.lev = 0) {
...
+#' @description
+#' These classes represent base types allowing them to be used within R7.
+#' Most correspond directly to the obvious base type. There are three
+#' exceptions:
+#'
+#' * `class_numeric` is a union of `class_integer` and `class_double`.
+#' * `class_atomic` is a union of `class_logical`, `class_numeric`,
+#'   `class_complex`, and `class_raw`.
+#' * `class_vector` is a union of `class_atomic`, `class_list`, and
+#'   `class_expression`.
```
DavisVaughan, line 67: I had a random thought that `class_numeric` and friends could be `union_numeric` to be more self documenting, but that might not be a great idea

R/base.R
```diff
@@ -53,17 +53,85 @@ str.R7_base_class <- function(object, ..., nest.lev = 0) {
...
+#'
```
DavisVaughan, line 68: Do you give any advice on how to name your own custom class objects anywhere?

R/property.R
```diff
@@ -9,8 +9,8 @@
 #' "dynamic" so that it's computed when accessed or has some non-standard
 #' behaviour when modified.
 #'
-#' @param class If specified, any values must be one of these classes
-#'   (or [class union][new_union]).
+#' @param class Class that property must be an instance of.
```
DavisVaughan, line 12:
```suggestion
#' @param class Class that the property must be an instance of.
```

+ redocument

DavisVaughan (approved):
I think this is a big net positive for consistency!

R/base.R
hadley: Nope currently. I'll add something here saying that we reserve the `class_` prefix and also add a little advice in `new_class()`

R/base.R
hadley: I think it's nice to have a common prefix here; and the prefix really means "some wrapper around a base class/type".

PR #209, opened 2022-03-15
Allow abstract classes

### LINKED ISSUE #199: Should we allow abstract classes?
With (say) `new_class("foo", abstract = TRUE)`.

hadley:
Fixes #199

### DIFF
--- a/tests/testthat/test-class.R
+++ b/tests/testthat/test-class.R
@@ -73,7 +73,7 @@ describe("abstract classes", {
       new_class("foo2", parent = foo1, abstract = TRUE)
     })
   })
-  it("can constructor concrete subclasses", {
+  it("can construct concrete subclasses", {
     foo1 <- new_class("foo1", abstract = TRUE)
     foo2 <- new_class("foo2", parent = foo1)
     expect_s3_class(foo2(), "foo2")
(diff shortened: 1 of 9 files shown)


tests/testthat/test-class.R
```diff
@@ -60,6 +60,26 @@ describe("inheritance", {
...
+      foo()
+    })
+  })
+  it("can't inherit from concrete class", {
+    expect_snapshot(error = TRUE, {
+      foo1 <- new_class("foo1")
+      new_class("foo2", parent = foo1, abstract = TRUE)
+    })
+  })
+  it("can constructor concrete subclasses", {
```
DavisVaughan, line 76:
```suggestion
  it("can construct concrete subclasses", {
```

tests/testthat/test-class.R
```diff
@@ -60,6 +60,26 @@ describe("inheritance", {
...
+    foo1 <- new_class("foo1", abstract = TRUE)
+    foo2 <- new_class("foo2", parent = foo1)
+    expect_s3_class(foo2(), "foo2")
+  })
+})
+
```
DavisVaughan, line 82: I think I was also expecting a few tests showing that you can create methods for abstract classes, and then have concrete subclasses that dispatch off those methods

DavisVaughan (approved):

tests/testthat/test-class.R
hadley: I didn't test that because I think it's highly unlikely that it wouldn't work given the current (and any future) implementation.

PR #213, opened 2022-03-18
Implement R7_inherits() and check_R7_inherits()

### LINKED ISSUE #193: inheritance + check functions
Need equivalent of `is()`/`inherits()`, along with a helper that asserts that an object is of a class (extracting code out of the property checkers). Implementation already exists in `class_inherits()`; just needs a better name.

hadley:
Fixes #193


R/inherits.R
```diff
@@ -0,0 +1,34 @@
+#' Does this object inherit from an R7 class?
+#'
+#' * `R7_inherits()` returns `TRUE` or `FALSE`
```
DavisVaughan, line 3:
```suggestion
#' * `R7_inherits()` returns `TRUE` or `FALSE`.
```

R/inherits.R
```diff
@@ -0,0 +1,34 @@
...
+  inherits(x, "R7_object") && inherits(x, R7_class_name(class))
+}
+
+#' @export
+#' @rdname R7_inherits
+check_R7_inherits <- function(x, class, arg = deparse(substitute(x))) {
+  if (!R7_inherits(x, class)) {
+    stop(sprintf("`%s` is not an %s", arg, class_desc(class)), call. = FALSE)
+  }
+  invisible()
```
DavisVaughan, line 33: Invisibly return nothing or `x`?

tests/testthat/_snaps/inherits.md
```diff
@@ -0,0 +1,16 @@
...
+      `class` is not an <R7_class>
+
+# throws informative error
+
+    Code
+      foo1 <- new_class("foo1")
+      foo2 <- new_class("foo2")
+      check_R7_inherits(foo1(), foo2)
+    Error <simpleError>
+      `foo1()` is not an <foo2>
```
DavisVaughan, line 15:
```suggestion
      `foo1()` is not a <foo2>
```

Should it be `a` rather than `an` in the general error message?

DavisVaughan (approved):

R/inherits.R
hadley: I think nothing is right; you call this purely for its side-effects?

tests/testthat/_snaps/inherits.md
hadley: Yeah, good catch.

hadley: Most of the existing uses seem to be followed with "R7" so an is mostly correct.

R/inherits.R
DavisVaughan: Yea I think so. If it modified `x` along the way while making checks then that'd be different

PR #214, opened 2022-03-22
Use explicit class in `S4_register()`

hadley:


R/S4.R
```diff
@@ -1,17 +1,44 @@
 #' Register an R7 class with S4
 #'
-#' If you want to use [method<-] to register an method for an S4 generic with
-#' an R7 class, you need to call `S4_register()` once.
+#' @description
+#' If you want to use and R7 class with S4 (e.g. to use [method<-] to register an
```
DavisVaughan, line 4:
```suggestion
#' If you want to use an R7 class with S4 (e.g. to use [method<-] to register an
```

R/S4.R
```diff
@@ -1,17 +1,44 @@
...
+#'
+#' * Matches class name and inheritance hierarchy.
+#' * Uses [validate()] as the validity method.
+#' * Defines formal S4 slots to match R7 properties. The slot types are
+#'   matched to the R7 property types, with the exception of R7 unions,
+#'   which are unchecked (due to the challenges of converting R7 unions to
+#'   S4 unions).
+#'
+#' If `class` extends another R7 class or has a property restricted to an
+#' R7 class, you you must register those classes first.
```
DavisVaughan, line 16: If you have a package where you are doing this (i.e. you have 2 R7 classes where one extends the other and you want them both to be S4 registered), this depends on the order that the files are sourced in, right? Since you call `S4_register()` globally in the files? I guess there isn't much we can do about that if that is the case, but it has potential to be confusing.

R/S4.R
```diff
@@ -1,17 +1,44 @@
...
 S4_register <- function(class, env = parent.frame()) {
   if (!is_class(class)) {
     msg <- sprintf("`class` must be an R7 class, not a %s", obj_desc(class))
+    stop(msg)
+  }
+
+  name <- class@name
+  contains <- double_to_numeric(setdiff(class_dispatch(class), "ANY")[-1])
+
+  # S4 classes inherits slots from parent but R7 classes flatten
```
DavisVaughan, line 30:
```suggestion
  # S4 classes inherit slots from parent but R7 classes flatten
```

R/S4.R
```diff
@@ -51,6 +78,25 @@ S4_to_R7_class <- function(x, error_base = "") {
...
+R7_to_S4_class <- function(x) {
+  switch(class_type(x),
+    NULL = "NULL",
+    any = "ANY",
+    S4 = S4_class_name(x),
+    R7 = R7_class_name(x),
+    R7_base = double_to_numeric(x$class),
+    R7_S3 = x$class[[1]],
+    R7_union = "ANY",
+    stop("Unsupported")
```
DavisVaughan, line 90: I guess you could have an R7 property of `class_missing`, and that would trigger this? That would be somewhat weird though. Everything else seems to be handled

tests/testthat/test-S4.R
```diff
@@ -108,3 +108,63 @@ describe("S4_class_dispatch", {
...
   })
 })
+
+describe("S4 registration", {
+  it("can register simple class hierarchy", {
+    foo <- new_class("foo")
+    foo2 <- new_class("foo2", foo)
+
+    S4_register(foo)
+    S4_register(foo2)
```
DavisVaughan, line 118: Do you get an error if you reverse the order of the `S4_register()` calls?

tests/testthat/test-S4.R
```diff
@@ -108,3 +108,63 @@ describe("S4_class_dispatch", {
...
+      properties = list(x = class_double)
+    )
+    S4_register(foo6)
+
+    obj <- new("foo6")
+    expect_type(obj, "double")
+    expect_type(slot(obj, "x"), "double")
+  })
+
+  it("checks it's inputs", {
```
DavisVaughan, line 167:
```suggestion
  it("checks its inputs", {
```

+ rebuild snapshots

DavisVaughan (approved):

R/S4.R
hadley: Right, but for R7 inheritance to work you'd have to have sourced the parent before the child.

lawremi:
What about setting the S4 class prototype using the default values from the R7 properties?

I guess we don't support virtual classes in R7 yet, but if we did, we'd obviously want to translate that, as well.

hadley:
Do you have an example of what `prototype` should look like? This is one of the arguments that the docs suggest you shouldn't use.

lawremi:
The prototype can be constructed with `prototype()`, like `prototype(slot=default_value)`.

hadley:
Closing since this is now very out of date.

PR #217, opened 2022-03-23
Custom dispatch for base Ops

### LINKED ISSUE #128: Considerations for `+` and other special infix operators
I think one of the nice things about R7 would be to be able to write `+` methods which dispatch off both inputs. Like `<R7Date> + <R7Duration> = <R7Date>`.

I think to do this correctly in all cases we will need some base R changes.

The C code for `x + y` currently checks to see if there is a `+` method for either `x` or `y`. If there is a method for exactly one of them, or if they both have the same method, then that method is used. Otherwise if there are methods for both and they aren't the same then a fallback internal implementation is used with a warning about "incompatible methods" (See `?groupGeneric`).

Practically this means that we _can_ currently make `<R7Date> + <R7Duration>` work because both of them will share `R7_object` as a base class. We will need to write a `+` method for `R7_object` which then hands off to something that actually handles the double dispatch right (See `vctrs:::+.vctrs_vctr` and `vctrs::vec_arith()` for an example of this).

But we _can't_ currently make `<Date> + <R7Duration>` work, because we'd end up with conflicting `+.Date` and `+.R7_object` methods and we'd end up getting the fallback behavior.

^ So figuring out some way to make `<Date> + <R7Duration>` work might be challenging, but would be a neat feature to promote about R7.

hadley:
Fixes #128

I'm tracking the changes needed for base R elsewhere (currently an Rmd, but soon to be an issue)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -8,6 +8,9 @@
 
 ## Mar 2022
 
+* Can now register methods for double-dispatch base Ops (currently only
+  works if both classes are R7) (#128).
+
 * All built-in wrappers around base types use `class_`. You can no longer
   refer to a base type with a string or a constructor function (#170).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,8 @@
 ## Mar 2022
 
 * Can now register methods for double-dispatch base Ops (currently only
-  works if both classes are R7) (#128).
+  works if both classes are R7, or the first argument is R7 and the second
+  doesn't have a method for the Ops generic) (#128).
 
 * All built-in wrappers around base types use `class_`. You can no longer
   refer to a base type with a string or a constructor function (#170).
--- a/NAMESPACE
+++ b/NAMESPACE
@@ -3,6 +3,7 @@
 S3method("$",R7_object)
 S3method("$<-",R7_object)
 S3method("@<-",R7_object)
+S3method(Ops,R7_object)
 S3method(print,R7_S3_class)
 S3method(print,R7_any)
 S3method(print,R7_base_class)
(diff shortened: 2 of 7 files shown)


NEWS.md
```diff
@@ -8,6 +8,9 @@
 
 ## Mar 2022
 
+* Can now register methods for double-dispatch base Ops (currently only
+  works if both classes are R7) (#128).
```
DavisVaughan, line 12: I think it also works if one of the sides is R7 and the other side doesn't have a method for the corresponding op? Like `<R7> + <double>`? It just won't work for `<R7> + <Date>` because `+.Date` exists?

i.e.

``` r
foo <- new_class("foo")

foo() + 1
#> Error: Can't find method for generic `+()` with dispatch classes:
#> - x: foo, R7_object, ANY
#> - y: double, ANY

method(`+`, list(foo, class_double)) <- function(x, y) "foobar"

foo() + 1
#> [1] "foobar"
```

vs

``` r
foo <- new_class("foo")

foo() + Sys.Date()
#> Warning: Incompatible methods ("Ops.R7_object", "+.Date") for "+"
#> Error in foo() + Sys.Date(): non-numeric argument to binary operator

method(`+`, list(foo, new_S3_class("Date"))) <- function(x, y) "foobar"

foo() + Sys.Date()
#> Warning: Incompatible methods ("Ops.R7_object", "+.Date") for "+"
#> Error in foo() + Sys.Date(): non-numeric argument to binary operator
```

DavisVaughan (approved):

NEWS.md
hadley: Oh yeah, good point.

PR #219, opened 2022-03-23
Improvements to check_R7_generic()

### LINKED ISSUE #215: Improvements to check_R7_inherits()
* Add "not a %s"

* Set default class to `R7_object  `

* Replace existing use of `check_R7` etc

hadley:
Fixes #215


NAMESPACE
```diff
@@ -36,7 +36,7 @@ export(R7_inherits)
 export(R7_object)
 export(S4_register)
 export(as_class)
-export(check_R7_inherits)
+export(check_is_R7)
```
DavisVaughan, line 39: I guess the idea behind this naming scheme is that we typically do:

```r
is_foo <- function(x) inherits(x, "foo")
```

So `check_is_R7()` maps to `is_foo()` and `R7_inherits()` maps to `inherits()` (for now)

hadley: Yeah, and `check_is_R7()` sounds right in my head. Could possible be `stopifnotR7()` but I don't think there's much prior art in that direction in base.

PR #223, opened 2022-03-24
Flesh out base S3 classes and export

hadley:
@DavisVaughan not sure how much to flesh these out, but at least they check for some of the most important issues.

Note that base R does do some checking when you set the class:

``` r
structure("x", class = "factor")
#> Error in attributes(.Data) <- c(attributes(.Data), attrib): adding class "factor" to an invalid object
```

<sup>Created on 2022-03-24 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Not sure how to get a better error message here without quite a lot of refactoring to set the attributes, call the validator, and then set the class.


R/S3.R
```diff
@@ -124,38 +124,94 @@ is_S3_class <- function(x) {
...
+
+validate_POSIXct <- function(self) {
+  if (!is.numeric(self)) {
+    return("Underlying data must be numeric")
+  }
+
+  tz <- attr(self, "tz")
+  if (!is.character(tz) || length(tz) != 1) {
+    return("attr(, 'tz') must be a single string")
+  }
```
DavisVaughan, line 152: Are you supposed to `c()` the strings together like in `validate_factor()`?

Same question about `validate_data.frame()` I guess

R/S3.R
```diff
@@ -124,38 +124,94 @@ is_S3_class <- function(x) {
...
+
+validate_data.frame <- function(self) {
+  if (!is.list(self)) {
+    return("Underlying data must be a <list>")
+  }
+
+  rn <- attr(self, "row.names")
+  if (!is.character(rn) && !is.integer(rn)) {
+    return("attr(, 'row.names') must be a <character> or <integer>")
+  }
```
DavisVaughan, line 163: Hmm, doesn't this materialize compact row names? We have to use something special in dplyr to avoid this, i.e.

```r
> dplyr:::dplyr_reconstruct.data.frame
function (data, template) 
{
    attrs <- attributes(template)
    attrs$names <- names(data)
    attrs$row.names <- .row_names_info(data, type = 0L)
    attributes(data) <- attrs
    data
}
```

R/S3.R
```diff
@@ -124,38 +124,94 @@ is_S3_class <- function(x) {
...
+
+  if (length(self) >= 1) {
+    ns <- unique(c(lengths(self), length(rn)))
```
DavisVaughan, line 166: The above point likely affects this line because `rn` might be an integer vector of length 2 in the compact form

R/S3.R
```diff
@@ -124,38 +124,94 @@ is_S3_class <- function(x) {
...
+
+    if (is.null(names(self))) {
+      return("Underlying data must be named")
+    }
+  }
+}
+
+#' @export
+#' @rdname base_classes
+#' @format NULL
```
DavisVaughan, line 179: Does this need an `@order` tag?

DavisVaughan (commented):
I don't think I would worry too much about the factor checks that base R does

R/S3.R
hadley: It's basically up to you and how much time you want to put into it. I don't think it's terrible if multiple problems need repeated calls to reveal every last detail.

R/S3.R
hadley: Oh yeah, good catch. It turns out that `structure()` won't let you set non-character/integer row names, so I can just use `type = 2` here to check that the length is correct.

PR #228, opened 2022-05-18
Fix check_is_R7 error message for non-R7 objects.

jonthegeek:
The sprintf had an extra argument (leftover from copy/paste of the other error message).

### DIFF
--- a/tests/testthat/test-inherits.R
+++ b/tests/testthat/test-inherits.R
@@ -17,4 +17,5 @@ test_that("throws informative error", {
     foo2 <- new_class("foo2")
     check_is_R7(foo1(), foo2)
   })
+  expect_snapshot(error = TRUE, check_is_R7("a"))
 })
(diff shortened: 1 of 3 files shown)


hadley:
Thanks! Would you mind adding a snapshot test?

jonthegeek:
> Thanks! Would you mind adding a snapshot test?

Done!

tests/testthat/test-inherits.R
```diff
@@ -17,4 +17,5 @@ test_that("throws informative error", {
     foo2 <- new_class("foo2")
     check_is_R7(foo1(), foo2)
   })
+  expect_snapshot(error = TRUE, check_is_R7("a"))
```
hadley, line 20: For an single expression, I’d put `error` at the end. 

And all the snapshots should be checked in. What difference are you seeing?

hadley:
Thanks!

PR #239, opened 2022-07-12
fix: Combining R7 objects gives error #230

### LINKED ISSUE #230: Make combining classes an error
I accidentally used `c()` rather than `list()` when doing `method(generic, c(class1, class2)) <- ` and got a weird error because the default method for `c()` just combines the classes as lists

``` r
library(R7)

c(class_Date, class_character)
#> $class
#> [1] "Date"
#> 
#> $constructor
#> function (.data = double()) 
#> {
#>     .Date(.data)
#> }
#> <bytecode: 0x7facd1d16320>
#> <environment: namespace:R7>
#> 
#> $validator
#> function (self) 
#> {
#>     if (!is.numeric(self)) {
#>         "Underlying data must be numeric"
#>     }
#> }
#> <bytecode: 0x7facd1d1ce00>
#> <environment: namespace:R7>
#> 
#> $class
#> [1] "character"
#> 
#> $constructor
#> function (.data = class_missing) 
#> {
#>     if (is_class_missing(.data)) {
#>         .data <- base_default(name)
#>     }
#>     .data
#> }
#> <bytecode: 0x7facd1c50ca8>
#> <environment: 0x7face372a3a8>
#> 
#> $validator
#> function (object) 
#> {
#>     if (base_class(object) != name) {
#>         sprintf("Underlying data must be <%s> not <%s>", name, 
#>             base_class(object))
#>     }
#> }
#> <bytecode: 0x7facd1c56050>
#> <environment: 0x7face372a3a8>
```

<sup>Created on 2022-06-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

jamieRowen:
Implement `c.R7_class` to give an error to close #230


tests/testthat/test-class.R
```diff
@@ -183,3 +183,8 @@ describe("default constructor", {
     expect_s3_class(x@x, "foo1")
   })
 })
+
+test_that("c(<R7_class>, ...) gives error", {
+  foo1 <- new_class("foo1")
+  expect_error(c(foo1, foo1))
```
hadley, line 189: This would be a good place to use `expect_snapshot(..., error = TRUE)` because that make it easier to verify the text of the error is useful (not that important here since there's no interpolation, but is a good general pattern, IMO).

hadley (approved):
Similarly, could you please add a short bullet to NEWS?

jamieRowen:
Updated test and NEWS

tests/testthat/test-class.R
```diff
@@ -183,3 +183,10 @@ describe("default constructor", {
     expect_s3_class(x@x, "foo1")
   })
 })
+
+test_that("c(<R7_class>, ...) gives error", {
+  foo1 <- new_class("foo1")
+  expect_snapshot(error = TRUE, {
+    c(foo1, foo1)
+  })
+})
```
hadley, line 192: Missing a trailing nl here

hadley (commented):
Just two last tiny suggestions

tests/testthat/test-class.R
jamieRowen: fixed

hadley:
Thanks!

PR #243, opened 2022-07-27
Respect visibility of method return values

### LINKED ISSUE #196: Ensure that generics preserve visibility
``` r
library(R7)
method(foo, integer) <- function(x) invisible(10)
foo(1L)
#> [1] 10
```

<sup>Created on 2022-02-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

dgkf:
Closes #196

Methods which return invisibly will now retain their invisible status. 

Perhaps there are better ways to manage this, but I took the simple route and just returned the call to evaluate from the R side. It's a bit roundabout, but provides an easy solution with minimal impact to the rest of the code base.

### Changes:

- C-level `method_call_` now returns the method call `SEXP`, not the return value
- `R7_dispatch` now evaluates the call from the R side to take advantage of R's internal `eval.c` visibility handlers, which I don't think are exposed to the R extension headers.

### Adds:

- Unit tests for visibility of method return

### DIFF
--- a/R/method-dispatch.R
+++ b/R/method-dispatch.R
@@ -14,5 +14,6 @@ method_lookup_error <- function(name, args, signatures) {
 #' @order 2
 #' @export
 R7_dispatch <- function() {
-  .Call(method_call_, sys.call(-1), sys.function(-1), sys.frame(-1))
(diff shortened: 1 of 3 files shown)


hadley:
This is such a simpler approach than I had imagined 😁

Issue #256, opened 2022-11-14
Using `...` in constructor of `new_class()` causes improper formals of subsequent child constructors


jtlandis:
It seems that if you were to write your own constructor method and include `...`, this may lead to errors in subsequent child classes if nothing is passed to the dots. Consider the following reprex.

``` r
library(R7)

container <- new_class("container",
                       properties = list(children = class_list),
                       constructor = function(...) new_object(NULL, children = list(...)))

extnd_cntr <- new_class("extnd_cntr", container)

extnd_cntr("foo", "bar")
#> <extnd_cntr>
#>  @ children:List of 2
#>  .. $ : chr "foo"
#>  .. $ : chr "bar"

#passing nothing to dots causes an error

extnd_cntr()
#> Error in .parent %||% class_construct(class@parent): '...' used in an incorrect context
extnd_cntr@constructor
#> function (... = class_missing) 
#> new_object(container(... = ...))
#> <environment: 0x55606e6f6720>

# replacing `... = ...` --> `...`
extnd_cntr2 <- new_class("extnd_cntr2",
                        container,
                        constructor = function(...) new_object(container(... = ...)))

extnd_cntr2()
#> <extnd_cntr2>
#>  @ children: list()
```

In the above example, it seems the true issue is with the formals passing an argument to `...`

<sup>Created on 2022-11-14 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #257, opened 2022-11-14
#256 any args named `...` should not have a formal assignment

jtlandis:
I believe this edit should be sufficient to fix this bug


hadley:
Would you mind separating these two separate issues into two separate PRs and adding a small unit test for each?

jtlandis:
Sure, I'll get on that this weekend

jtlandis:
This Pull Request has now been split into Pull Requests #264 and #265, both containing edits for unit tests as well. I will now close this PR.

Issue #258, opened 2022-11-15
Child objects cannot have abstract parents


jl5000:
Unless I'm misunderstanding what abstract classes are, I thought objects with abstract parent classes should be able to be created?

``` r
library(R7)

parent = new_class("new_class", properties = list(prop1 = class_character))
child = new_class("child", parent = parent)
grandchild = new_class("grandchild", parent = child)

grandchild()
#> <grandchild>
#>  @ prop1: chr(0)

parent = new_class("new_class", abstract = TRUE, properties = list(prop1 = class_character))
child = new_class("child", parent = parent, abstract = TRUE)
grandchild = new_class("grandchild", parent = child)

grandchild()
#> Error: <grandchild> object properties are invalid:
#> - @prop1 must be <character>, not <NULL>
```

<sup>Created on 2022-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

jtlandis:
There seems to be a small typo, I noticed this issue earlier today as well.

hadley:
Looks like something is going wrong with property setting/inheritance:

``` r
library(R7)

parent <- new_class(
  "new_class",
  abstract = TRUE,
  properties = list(prop1 = class_character)
)

child <- new_class("child", parent = parent)
child()
#> Error: <child> object properties are invalid:
#> - @prop1 must be <character>, not <NULL>
child(prop1 = "x")
#> Error: <child> object properties are invalid:
#> - @prop1 must be <character>, not <NULL>
```

<sup>Created on 2022-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

jtlandis:
This issue is a typo that affects how the constructor call is formulated, attaching arguments as the second element of the list instead of binding with `c()`.

``` r
library(R7)

parent <- new_class(
  "new_class",
  abstract = TRUE,
  properties = list(prop1 = class_character)
)

child <- new_class("child", parent = parent)
child@constructor
#> function (prop1 = class_missing) 
#> new_object(R7_object(), list(prop1 = prop1))
#> <environment: namespace:R7>
```

<sup>Created on 2022-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

The following change fixed the issue

![Screenshot from 2022-11-15 16-34-28](https://user-images.githubusercontent.com/30324172/202030026-f896530e-3690-40e0-b2f2-f98329e0675d.png)

hadley:
[...]

PR #265, opened 2022-11-24
Parenthesis typo causes issue with classes inheriting from abstract classes

jtlandis:
This pull request should address issue #258 and be the final part of PR #257. Please review the changes to the testthat.


tests/testthat/test-constructor.R
```diff
@@ -57,5 +57,7 @@ test_that("can generate constructor for inherited abstract classes", {
     foo1 <- new_class("foo1", abstract = TRUE, properties = list(x = class_double))
     new_constructor(foo1, list())
     new_constructor(foo1, as_properties(list(y = class_double)))
+    child <- new_class("child", foo1, properties = list(y = class_double))
+    child(y = 0.5)
```
hadley, line 61: Given that the output of this operation is known (i.e. you can test the values of the properties), I think this would be better outside of the snapshot test.

hadley (commented):
Thanks for working on this 😄

tests/testthat/test-constructor.R
jtlandis: Previously, the tests were not failing because the test was seeing if the internal `new_constructor` function was "working", but it never tested if the class could be made.

I could move this outside the snapshot

hadley: I agree that you should test actually constructing the objects, I just think you should do it outside of the snapshot.

hadley:
Thanks for spotting this and fixing it!

PR #273, opened 2023-01-13
Update dispatch.Rmd

tabauer23:
Just one typo to fix!


hadley:
That was a deliberate typo to illustrate the problem.

PR #280, opened 2023-02-19
Don't create a constructor called "function"

### LINKED ISSUE #249: Class inheriting from function generates invalid constructor
``` r
library(R7)

foo <- new_class("foo", class_function)
unclass(foo)
#> Error in print.default(x, useSource = useSource, ...): badly formed function expression
```

<sup>Created on 2022-08-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### LINKED ISSUE #279: S7_class is unprintable?
This may be related to #272:

``` r
library(S7)

dog <- new_class("dog", properties = list(
  name = class_character,
  age = class_numeric
))

speak <- new_generic("speak", "x")

method(speak, dog) <- function(x) {
  "Woof"
}

fn <- method(speak, dog)
unclass(attr(fn, "S7_class"))
#> Error in print.default(x, useSource = useSource, ...): badly formed function expression
```

I think this error happens because the arglist is `.Data` as a name.  Normally arglists should be pairlists or language objects.

hadley:
Fixes #249. Fixes #279.

### DIFF
--- a/R/base.R
+++ b/R/base.R
@@ -1,4 +1,4 @@
-new_base_class <- function(name) {
+new_base_class <- function(name, constructor_name = name) {
   force(name)
 
   constructor <- function(.data = class_missing) {
@@ -16,6 +16,7 @@ new_base_class <- function(name) {
 
   out <- list(
     class = name,
+    constructor_name = constructor_name,
     constructor = constructor,
     validator = validator
   )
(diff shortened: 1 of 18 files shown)


dmurdoch:
That does fix #279 for me.

PR #288, opened 2023-04-14
Subclasses of abstract class can have readonly properties

### LINKED ISSUE #269: Child objects with abstract parents cannot have read-only properties
``` r
library(R7)

parent <- new_class("new_class", abstract = TRUE)

child <- new_class("child", parent = parent,
                   properties = list(
                     x = new_property(class_character,
                                      getter = function(self) "test") 
                       ))
child()
#> Error: Can't set read-only property <child>@x
```

<sup>Created on 2022-12-12 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
Fixes #269


hadley:
* [ ] Add news bullet

PR #290, opened 2023-04-14
Don't allow inheritance from environments

### LINKED ISSUE #253: R7_data() strips class attributes from class_environment instance
```r
library(R7)
my_class <- new_class(
    "my_class",
    parent = class_environment,
    properties = list(
           test = class_numeric
    )
)
instance <- my_class()
```

```r
print(instance)
#> <my_class> <environment: 0xae77818> 
#>   @ test: int(0) 


R7_data(instance)
print(instance)
#> <environment: 0xae77818> ## class is lost

instance@test
#> Error in do.call(base::`@`, list(object, name)) : 
#>  trying to get slot "test" from an object of a basic class ("environment") with no slots
```

This is due to `R7_data()` calling `zap_attr()` which affects environments differently to other base R classes. There's not a whole lot of point calling R7_data on an environment class, but it's a little confusing to remember when learning the R7 system. One solution could be to just have R7_data be an identity function for environments. i.e.,

```r
R7_data.environment <- function(object) {
    object
}
```

Edit: added a better example

hadley:
At least for now. Fixes #253.


hadley:
* [ ] Add news bullet

DavisVaughan (approved):

PR #291, opened 2023-04-14
Avoid repeated re-validation during construction

### LINKED ISSUE #248: Parent validators are called multiple times during object construction
Naively this feels like it should be true, particularly if the parent class comes next in line when looking up methods.  I.e. if I have a parent class `pet`, and child classes `dog` and `cat`, a valid `dog` object should also be a valid `pet` object.

On the other hand I imagine it could get slow/complicated to be running through a whole chain of ancestor validators.  And possibly a child class might not want to conform to the parent validator if it's too strict, and belongs to somebody else so can't be adjusted.  So is this the sort of thing that should be handled by trying to behave politely instead by enforcing it?

If that's the case, If I'm creating a child class, is there a way I could choose to run the parent class's validator on my objects if I'm trying to play nicely?

I don't have much of an OOP background, so I don't know how this is handled in S4 or in other languages.

hadley:
Fixes #248


hadley:
* [ ] Add news bullet

PR #292, opened 2023-04-15
Better filtering strategy for S4 class hierarchy

### LINKED ISSUE #252: R7 dispatch and S4 inheritance
Is it not possible for an R7 generic to understand S4 inheritance? Is this an intentional limitation or a bug? (Or am I simply confused for expecting this to work?)

```r
methods::setClass("general")
methods::setClass("specific", contains = "general")

methods::is(methods::new("specific"), "general")
#> [1] TRUE

gen <- methods::getClass("general")

fun <- R7::new_generic("fun", "x")

R7::method(fun, gen) <- function(x) "ok"

fun(methods::new("specific"))
#> Error: Can't find method for generic `fun()` with dispatch classes:
#> - x: S4/specific
```

(I'm at `RConsortium/OOP-WG@879da2a`)

hadley:
Fixes #252


hadley:
* [ ] Add news bullet

hadley:
@lawremi do you have enough understanding of S4 classes to review this? The goal is to remove S4 unions (since S7 handles those at registration time, not dispatch time). But I can't find a single criteria that appears to work for both user created unions and system created unions (e.g. vector, oldClass).

mmaechler:
@lawremi does, and I do too.  I think there's a misunderstanding on your side: e.g., `"vector"` is *not* a class union

mmaechler:
thank you.  Maybe we should `class_vector` for the S7, and `"vector"` for the S4 one.

PR #293, opened 2023-04-15
Start on motivation vignette

### LINKED ISSUE #92: Write motivations vignette that describes advantages over S3 and S4
* Printing generic shows the methods
* `@` doesn't do partial matching
* dynamic properties allow for deprecation
* generics create evaluation environments; no funny business with argument values vs local values.
* can evolve arguments to generics without CRAN breakage
* has formal definition

hadley:
Fixes #92

### DIFF
--- a/vignettes/motivation.Rmd
+++ b/vignettes/motivation.Rmd
@@ -1,0 +1,67 @@
+---
+title: "Motivation for S7"
+output: rmarkdown::html_vignette
+vignette: >
+  %\VignetteIndexEntry{Motivation for S7}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+R already has two OO systems built-in (S3 and S4) and many additional OO systems are available in CRAN packages.
+Why did we decide another system (S7) was necessary?
+This vignette will discuss some of the motivations, focussing on the aspects of S3 and S4 that have been found to be particularly challenging in practice.
+
+```{r setup}
+library(S7)
+```
+
+## Challenges with S3
+
+-   S3 is very informal, meaning that there's no formal definition of a class.
+    This makes it impossible to know exactly which properties an object should or could possess, or even what it's parent class should be.
+    S7 resolves this problem with a formal definition encoded in a class object produced by `new_class()`.
+    This includes support for validation (and avoiding validation where needed) as inspired by S4.
+
+-   When a new user encounter an S3 generic, they are often confused because the implementation of the function appears to be missing.
+    S7 has a thoughtfully designed print method that makes it clear what methods are available and how to find their source code.
+
+-   Properties of an S3 class are usually stored in attributes, but, by default, `attr()` does partial matching, which can lead to bugs that are hard to diagnose.
+    Additionally, `attr()` returns `NULL` if an attribute exists, so misspelling an attribute can lead to subtle bugs.
+    `@` fixes both of these problems.
+
+-   S3 method dispatch is complicated for compatibility with S.
+    This complexity affects relatively little code, but when you attempt to dive into the details it makes `UseMethod()` hard to understand.
+    As much as possible, S7 avoids any "funny" business with environments or promises, so that there is no distinction between argument values and local values.
+
+-   `NextMethod()` is unpredictable you can't tell exactly what method will be called by only reading the code; you instead need to know both the complete class hierarchy and what other methods are currently registered (and loading a package might change those methods).
+    S7 takes a difference approach with `super()`, requiring explicit specification of the superclass to be used.
+
+-   Conversion between S3 classes is only implemented via loose convention: if you implement a class `foo`, then you should also provide generic `as.foo()` to convert other objects to that type.
+    S7 avoids this problem by providing the double-dispatch `convert()` generic so that you only need to provide the appropriate methods.
+
+## Challenges with S4
+
+-   In S4, slots are generally thought of as implementation detail that should not be directly accessed by the end-user.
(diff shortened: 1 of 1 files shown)


vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+This vignette will discuss some of the motivations, focussing on the aspects of S3 and S4 that have been found to be particularly challenging in practice.
+
+```{r setup}
+library(S7)
+```
+
+## Challenges with S3
+
+-   S3 is very informal, meaning that there's no formal definition of a class.
+    This makes it impossible to know exactly which properties an object should or could possess, or even what it's parent class should be.
```
DavisVaughan, line 28:
```suggestion
    This makes it impossible to know exactly which properties an object should or could possess, or even what its parent class should be.
```

vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+    S7 resolves this problem with a formal definition encoded in a class object produced by `new_class()`.
+    This includes support for validation (and avoiding validation where needed) as inspired by S4.
+
+-   When a new user encounter an S3 generic, they are often confused because the implementation of the function appears to be missing.
+    S7 has a thoughtfully designed print method that makes it clear what methods are available and how to find their source code.
+
+-   Properties of an S3 class are usually stored in attributes, but, by default, `attr()` does partial matching, which can lead to bugs that are hard to diagnose.
+    Additionally, `attr()` returns `NULL` if an attribute exists, so misspelling an attribute can lead to subtle bugs.
```
DavisVaughan, line 36:
```suggestion
    Additionally, `attr()` returns `NULL` if an attribute doesn't exist, so misspelling an attribute can lead to subtle bugs.
```

vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+    `@` fixes both of these problems.
+
+-   S3 method dispatch is complicated for compatibility with S.
+    This complexity affects relatively little code, but when you attempt to dive into the details it makes `UseMethod()` hard to understand.
+    As much as possible, S7 avoids any "funny" business with environments or promises, so that there is no distinction between argument values and local values.
+
+-   `NextMethod()` is unpredictable you can't tell exactly what method will be called by only reading the code; you instead need to know both the complete class hierarchy and what other methods are currently registered (and loading a package might change those methods).
```
DavisVaughan, line 43:
```suggestion
-   `NextMethod()` is unpredictable since you can't tell exactly which method will be called by only reading the code; you instead need to know both the complete class hierarchy and what other methods are currently registered (and loading a package might change those methods).
```

vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+    S7 takes a difference approach with `super()`, requiring explicit specification of the superclass to be used.
+
+-   Conversion between S3 classes is only implemented via loose convention: if you implement a class `foo`, then you should also provide generic `as.foo()` to convert other objects to that type.
+    S7 avoids this problem by providing the double-dispatch `convert()` generic so that you only need to provide the appropriate methods.
+
```
DavisVaughan, line 48: Possibly an S3 bullet about double dispatch in things like `+` not being very easy to do?

DavisVaughan (approved):

vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+```{r, include = FALSE}
+knitr::opts_chunk$set(
+  collapse = TRUE,
+  comment = "#>"
+)
+```
+
+R already has two OO systems built-in (S3 and S4) and many additional OO systems are available in CRAN packages.
+Why did we decide another system (S7) was necessary?
```
lawremi, line 18: This breaks from the narrative that we are building on top of S3. I think we really want to avoid arguing for a new object system. I would frame it around how can we learn from S4 to improve S3.

vignettes/motivation.Rmd
```diff
@@ -0,0 +1,67 @@
...
+    This leads to two problems.
+    Firstly, implementing an S4 class often also requires a plethora of accessor functions that are a thin wrapper around `@` or `@<-`.
+    Secondly, users know about `@` and often use it even though they're not supposed to.
+    S7 avoids these problems by accepting the fact that R is a data language, and that there's no way to stop users from pulling the data they need out of an object.
+    To make it possible to change the internal implementation details of an object while preserving existing `@` usage, S7 provides dynamic properties.
+
+-   Multiple inheritance seemed like a powerful idea at the time, but in practice it appears to generate more problems than it solves.
+    S7 does not support multiple inheritance.
+
+-   S4's method dispatch uses a thoughtful system of weights to pick the best method when there is ambiguity.
```
lawremi, line 65: Instead of "thoughtful" I would feel comfortable saying "principled but complex". Instead "system of weights" I might say "distance measure". I think it's similar to the Manhattan distance.

hadley:
Thanks @lawremi. 

@ltierney, @mmaechler did you want to provide feedback too? If I don't hear otherwise in a week, I'll assume you're ok with it and merge it.

mmaechler:
I am really quite unhappy with your statement (at the beginning of "S4 .." part)
-   In S4, slots are generally thought of as implementation detail that should not be directly accessed by the end-user.

The above is how the bioconductor project has looked at it and recommended to its pkg developers.
But it is not true for  John Chambers  and what he wrote about S4.   His examples have used explicit slot access (@ syntax) in quite a few places,  and so have we `{Matrix} `package authors and other pkg developers (The whole {`distr`*} package family; my `{Rmpfr}`, our `{copula}`  packages etc.     In all these cases there parly are sometimes natural accessors or "setters", but slot access is common, documented and natural other for parts.

hadley:
@mmaechler ok, I've changed that bullet to make it more clear that it applies to Bioconductor S4 classes.

PR #297, opened 2023-04-17
Syntax sugar for unions

### LINKED ISSUE #224: Lightweight syntax for creating unions
e.g. `class_character | class_double`

### LINKED ISSUE #259: Can't set property default to `NULL`
I'm trying to define some properties which are based on a custom class. Whilst I can do this, I can't figure out a way to make the property optional. In the reprex below, the `potentially_missing` class must contain an x value. However, in `myclass` I can't omit the `y` argument since it attempts to create an empty `potentially_missing` object. I've tried unioning it with `class_missing` but it doesn't seem to make a difference. How do I do something like this?

``` r
library(R7)

potentially_missing <- new_class("potentially_missing",
                                 properties = list(
                                   x = class_numeric
                                 ),
                                 validator = function(self){
                                   if(length(self@x) == 0) return("Error")
                                 }
)

potentially_missing()
#> Error: <potentially_missing> object is invalid:
#> - Error

myclass <- new_class("myclass",
                     properties = list(
                       x = new_property(new_union(potentially_missing, class_missing)),
                       y = new_property(new_union(potentially_missing, class_missing))
                     )
)

myclass(x = potentially_missing(1))
#> Error: <potentially_missing> object is invalid:
#> - Error
```

<sup>Created on 2022-11-15 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley:
Fixes #224

Also documents how the constructor works, and how to create an "optional" value. Fixes #259.


R/union.R
```diff
@@ -1,14 +1,19 @@
...
+#' with `new_union(a, b, c)` or `a | b | c`. Unions can be used in two
 #' places:
 #'
 #' * To allow a property to be one of a set of classes,
-#'   `new_property(new_union("integer", Range))`.
+#'   `new_property(class_integer | Range)`. The default `default` value for the
+#'   property will be the constructor of the first object in the union.
+#'   This means if you want to create an "optional" property (i.e. one that
+#'   can be `NULL` or of a specified type), you'll need to write (e.g.)
+#'   `NULL | class_integer`.
```
DavisVaughan, line 13: FWIW I think this is pretty nice

R/union.R
```diff
@@ -41,6 +48,23 @@ new_union <- function(...) {
...
 }
 
+#' @export
+`|.S7_class` <- function(e1, e2) {
+  new_union(e1, e2)
+}
+#' @export
+`|.S7_union` <- `|.S7_class`
+#' @export
+`|.S7_base_class` <- `|.S7_class`
```
DavisVaughan, line 58: Being able to define methods for `class_character` vs `"character"` is pretty nice. i.e. `"integer" | "character"` wouldn't work the same

DavisVaughan (approved):

R/union.R
hadley: Yeah, I think I put off implementing this because I was worried about introducing ambiguities with `|`, but it lucky there aren't any any more 😄

PR #299, opened 2023-04-17
Improve dispatch failure message

### LINKED ISSUE #231: Use `obj_desc()` in method failure errors
By passing along as an additional argument

And specialise for the common case of a single argument.

hadley:
* Different messages for single vs multiple dispatch
* Use `obj_type()`

Fixes #231


R/class-spec.R
```diff
@@ -187,7 +187,9 @@ class_inherits <- function(x, what) {
 }
 
 obj_type <- function(x) {
-  if (inherits(x, "S7_object")) {
+  if (identical(x, quote(expr = ))) {
```
DavisVaughan, line 190: `rlang::is_missing()` is:

```r
missing(x) || identical(x, quote(expr = ))
```

do we need that `missing(x)` bit?

R/convert.R
```diff
@@ -64,7 +64,12 @@ convert <- function(from, to, ...) {
     }
     from
   } else {
-    method_lookup_error("convert", c("from", "to"), dispatch)
+    msg <- paste0(
+      "Can't find method for generic `convert()` with dispatch classes:\n",
+      "- from: ", obj_desc(from), "\n",
+      "- to  : ", class_desc(to), "\n"
+    )
+    stop(msg, call. = FALSE)
```
DavisVaughan, line 72: Out of curiosity, why not use `method_lookup_error()`? The only thing that looks like it changed in the snapshot is the usage of `<` `>`

DavisVaughan (approved):

R/class-spec.R
hadley: I don't think it's important here.

R/convert.R
hadley: Because `to` is a class, not an object.

PR #301, opened 2023-04-18
Check on more versions

hadley:
And update actions


R/zzz.R
```diff
@@ -123,4 +123,13 @@ chooseOpsMethod.S7_object <- function(x, y, mx, my, cl, reverse) TRUE
   class_numeric <<- new_union(class_integer, class_double)
   class_atomic <<- new_union(class_logical, class_numeric, class_complex, class_character, class_raw)
   class_vector <<- new_union(class_atomic, class_expression, class_list)
+
+  # Dynamic register so that function pointers are the same, avoiding R 4.0
+  # and earlier bug
```
DavisVaughan, line 128:
```suggestion
  # and earlier bug related to incompatible S3 methods during Ops dispatch
```

just to give us a little hint if we wonder about this

DavisVaughan (approved):

PR #303, opened 2023-06-08
Updates for `OBJSXP`

t-kalinowski:
- update `typeof()` usage for new `"object"` return value


hadley:
No longer needed since @mmaechler's fix.

PR #313, opened 2023-07-11
Lazy registration for S3 and S4 methods

hadley:
Build on the existing external registration mechanism to also register S3 and S4 methods on load. Will finish up the docs in #312.


R/generic-spec.R
```diff
@@ -1,40 +1,70 @@
...
+    package = package
+  )
   class(out) <- "S7_S3_generic"
   out
 }
 
+find_S3_package <- function(generic, name, env = parent.frame()) {
+  while (!identical(env, emptyenv())) {
+    candidate <- env[[name]]
+    if (identical(candidate, generic)) {
```
hadley, line 43: There must be a simpler way to do this; `getPackageName()` works but is in methods 😬

hadley: Actually is this even correct if the generic is imported?

tests/testthat/test-method-register.R
```diff
@@ -81,6 +74,8 @@ describe("method registration", {
     expect_snapshot_error(method(bar, S4foo) <- function(x) "foo")
 
     S4_register(S4foo)
+    on.exit(S4_remove_classes("S4foo"))
```
hadley, line 77: So you can run this test repeatedly

DavisVaughan, line 77:
```suggestion
    on.exit(S4_remove_classes("S4foo"), add = TRUE)
```

I have definitely burnt myself in the past by unknowingly having multiple `on.exit()`s in the same block where one nuked the other

R/generic-spec.R
```diff
@@ -1,40 +1,65 @@
...
   use_method <- find_call(body(x), quote(UseMethod))
   if (!is.null(use_method)) {
-    return(S3_generic(x, as.character(use_method[[2]])))
+    name <- as.character(use_method[[2]])
+    pkg <- find_S3_package(x, name, env)
+    return(S3_generic(x, name, pkg))
   } else {
     name <- find_base_name(x)
     if (name %in% names(base_ops)) {
       return(base_ops[[name]])
```
DavisVaughan, line 21: I guess you'd never externally register for a base op, but it probably still works anyways because I think the things stored in this are actually S7 generics?

tests/testthat/_snaps/generic-spec.md
```diff
@@ -12,3 +12,11 @@
     Error <simpleError>
       `generic` must be a function, not a <double>
 
+# clear error if can't find generic
+
+    Code
+      find_S3_package(tibble::as_tibble, "as_tibble")
+    Error <simpleError>
+      Can't find package that generic `as_tibble` belongs to.
```
DavisVaughan, line 20: Nit: `as_tibble()` vs `as_tibble`?

DavisVaughan (approved):

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # S7 0.0.0.9000
 
+## May-July 2023
+
+* `external_methods_register()` now registers S3 and S4 methods too (#306).
```
DavisVaughan, line 5: I feel like the docs of this function could maybe use an update saying that this is supported

DavisVaughan:
Don't you always use it with `new_external_generic()` though? I like that "external" is in both

hadley:
It now use works with external S3 and S4 generics, which don't need `new_external_generic` (assuming they're imported).

mmaechler:
> I wonder if I should rename `external_methods_register()` to just `methods_register()`?

In general, I'm strongly prefering more concise function names.   The `_external` is of course necessary in the case of `new_external_generic()`  because of  `new_generic()` with a different meaning
{ unless, those are merged into   `new_generic(.....,  external=TRUE)`  --  which is more along my taste.

DavisVaughan:
@hadley i saw in the tests that it was being used like

```r
bar <- new_class("bar")
base_sum <- new_external_generic("base", "sum", "x")
method(base_sum, bar) <- function(x, ...) "bar"
```

Which still used `new_external_generic()`, which is why I thought you'd always see at least one `new_external_generic()` call if you were also using `external_methods_register()`, but I'm probably missing something

hadley:
I think I had forgotten the previous limitations of  `new_external_generic()`. Prior to this PR, it was needed whenever you used an S7 generic from another package, regardless of whether you were taking a hard or soft dependency. Now (after the commit I'm about to push) you use it whenever you want to use a generic (S7 or S3 or S4) from a soft dependency.  

@mmaechler I think this more clearly reveals the purpose of `new_external_generic()`, making it more clear that it shouldn't just be an argument to `new_generic()`, and why `external_methods_register()` should definitely be `methods_register()`. (Or possible `register_methods()` since that sounds better to my ear)

PR #314, opened 2023-07-13
Polish convert and super docs

hadley:


R/convert.R
```diff
@@ -1,14 +1,29 @@
...
-#' `convert()` provides built-in implementations if `from` inherits from `to`.
-#' This default strips any properties that `from` possesses that `to` does not,
-#' and resets the class.
+#' * It uses double-dispatch, because conversion depends on both `from` and
+#'   `to`.
+#'
+#' * It uses non-standard dispatch because `to` is a class, not an object.
+#'
+#' * It doesn't respect inheritance because if you convert `x` to `superFoo`
+#'   you shouldn't get an instance of `Foo` back).
```
DavisVaughan, line 13:
```suggestion
#'   you shouldn't get an instance of `Foo` back.
```

DavisVaughan, line 13: I actually don't know what this is trying to say. How could you get an instance of `Foo` back if inheritance was being respected?

I'm sure i understood this at some point in the past, but on reread it isn't obvious to me anymore

R/convert.R
```diff
@@ -1,14 +1,29 @@
...
+#'
+#' `convert()` provides a default implementations when `from` inherits from
```
DavisVaughan, line 15:
```suggestion
#' `convert()` provides a default implementation when `from` inherits from
```

R/convert.R
```diff
@@ -1,14 +1,29 @@
...
+#' `to`. This default strips any properties that `from` possesses that `to`
+#' does not.
+#'
+#' If you are converting an object solely for the purposes of accessing a method
+#' on a superclass, you probably want [super()] instead. See its docs for more
+#' details.
+#'
+#' ## S3 & S4
+#'
+#' `convert()` plays a similar role to the convention of defining `as.foo`
```
DavisVaughan, line 25:
```suggestion
#' `convert()` plays a similar role to the convention of defining `as.foo()`
```

R/convert.R
```diff
@@ -20,20 +35,26 @@
...
-#' convert(foo2(x = 1L, y = 2), to = foo1)
+#' method(convert, list(foo1, class_integer)) <- function(from, to) {
+#'   from@x
+#' }
+#' convert(foo1(x = 1L), to = class_integer)
 #'
-#' # If we define a convert method for integer + foo1:
-#' method(convert, list(class_integer, foo1)) <- function(from, to) foo1(x = from)
+#' # Note that conversion does not respect inheritance so if we define a
+#' # convert method for foo1 to integer:
```
DavisVaughan, line 51:
```suggestion
#' # convert method for integer to foo1:
```

R/convert.R
```diff
@@ -20,20 +35,26 @@
...
+#' method(convert, list(class_integer, foo1)) <- function(from, to) {
+#'   foo1(x = from)
+#' }
 #' convert(1L, to = foo1)
-#' # Converting too foo2 still errors
+#'
+#' # Converting too foo2 will still error
```
DavisVaughan, line 57:
```suggestion
#' # Converting to foo2 will still error
```

DavisVaughan, line 57: This example actually seems somewhat obvious to me, because we don't have `y`, which foo2 needs, so of course it fails!

In my head a more interesting example could be to have

```
bar <- new_class("bar", properties = list(x = class_integer, y = class_double))
```

If you define a `convert()` method for `bar -> foo2`, do you also get `bar -> foo1` "for free"? This would be theoretically possible because `foo1` only needs `x`, and you could get there by "automatically" chaining explicit convert methods like `bar -> foo2 -> foo1`, but I don't think we do this (nor should we)

R/super.R
```diff
@@ -1,11 +1,12 @@
...
+#' `super(from, to)` causes the dispatch for the next generic to use the method
+#' for the superclass `to` instead of the actual class of `from`. It's needed
+#' when you want to implement a method in terms of the implementation of its
+#' superclass.
 #'
-#' # Compared to S3 and S4
+#' ## S3 & S4
 #' `super()` performs a similar role to [NextMethod()] in S3 or
 #' [methods::callNextMethod()] in S4, but is much more explicit:
 #'
```
DavisVaughan, line 12: I'm reading the next bullet here

```
* The class that `super()` will dispatch to is known at the time you
write `super()`, not only when it's called.
```

The wording here after the comma seems slightly confusing, maybe

```
* The class that `super()` will dispatch to is known statically at the time you
write `super()`, as opposed to dynamically determining it at the time the generic
is called.
```

DavisVaughan (approved):

R/convert.R
hadley: I think this is the most common failure more because most subclasses will add additional properties, so this is the common failure mode. Maybe I can just explain that.

Issue #321, opened 2023-08-18
`Ops.S7_object`  containing `%*%`


mmaechler:
Related to #320, I was exploring and found that indeed, dispatch for arithmetic happens with group method for `Ops` :
```
S7:::Ops.S7_object
function (e1, e2) 
{
    base_ops[[.Generic]](e1, e2)
}
<bytecode: 0x80a7f08>
<environment: namespace:S7>
```
and indeed, as mentioned in #320   the `base_ops` involved here are somewhat not initialized (anymore ?) and just empty.
**HOWEVER**  there is more:   The S7 `base_ops`  do contain `%*%`  but really really should NOT.

It seems clear this is currently explicitly "desired" --- probably because we don't want `S7` to depend on R 4.3.0 yet, and hence keep the code simple enough...
     https://github.com/RConsortium/OOP-WG/blame/main/R/zzz.R#L105
but it is really "wrong" and in contrast with base R's documentation of  group generics (`?groupGenerics`) which explains that `"matrixOps"` is a separate group generic.

mmaechler:
Also, the "signature" of `%*%`  is  `(x,y)`,  not `(e1,e2)`   and that's also the case for `crossprod()` and `tcrossprod()` which should become  "matrixOps"  eventually, too.

But yes, this can wait, e.g. for release of R 4.4.x

mmaechler:
I have spent more thinking on this:  I've come the stage of claiming that the whole idea of making `%*%`   a group generic at all was flawed, and only adds unnecessary complications; notably also for R itself. 
My current pretty strong opinion is that R core should abolish the (new in R 4.3.x) `"matrixOps"` group generic, and S7 should also make  `%*%` a  normal 2-argument generic function.
If it can be dealt with as part of a group generic, it can also be dealt with on its own.
I think extra group generic (`"matrixOps"`) just complicates everything and has no benefit or almost no benefit in the case we would add `"crossprod"` and `"tcrossprod"` to it.

t-kalinowski:
If I recall correctly, the original motivation for adding the matrixOps group was partially motivated by economy - the Ops Group generic is the only S3 dispatch mechanism that allows a function to dispatch on the 2nd argument. Wanting to make `%*%` work with S7 (meaning, S3),
we were faced with the choice between making 2-argument dispatch generally available to functions outside of the Ops Group, or extending the Ops group to include a new member. 

[...]

PR #322, opened 2023-08-21
CRAN release

hadley:


mmaechler:
Hmm, regarding the non-working Ops (for S7 objects extending class_double or class_integer).
I've looked a bit more at some of the code  ... and am less and less fan of your original idea to add `%*%` to the "base_ops" -- #321 (!).    It just does not belong there, also because it has  `(x,y)`  signature argument names and the `Ops` should have (but have *not* currently, in S7!)   `signature(e1, e2)`.
I guess the only reason to add it there might be that you want to have all the 2-argument dispatch in one place.  I don't know....  but currently I think that even the   "workaround" to introduce `"matrixOps"`  (as group generic in `base`) was bad because it does not really help anything ... and even we'd augment that group by `crossprod()` and `tcrossprod()` I cannot currently see that the overal code {methods for "matrixOps" instead of for `%*%`  etc} would become easier/less complex,
to the contrary.

Is there any need to have  a `%*%`  S7 generic at this point?   If yes, why can't it just be a 2-argument generic for `%*%` ?
It is just *wrong* in comparison with all the base R (and methods package)  group methods and group member documentation.

So my plea:  Can we think a bit longer, if we couldn't 
1) remove %*% from the Ops / `base_ops` 
2) add a direct generic for `%*%`
3) add *default* methods for the real base_ops such that eg. the simple extensions of `class_double` (and *_complex, *_integer) would automatically work   the same as they work for the dozens of functions of group "Math" --- which are notably working correctly because we did *NOT* define an S7 generic for them .. and they automagically work via base R's C level dispatch  ?!

hadley:
@mmaechler IMO it's not worth waiting (and not just because I accidentally submitted to CRAN before I saw your message) just because no one (effectively) will learn about this package until we advertise it and 1-2 weeks will give us enough time to do another patch release. Getting through the initial submission is more time consuming than any other submission and I don't want to lose momentum now.

PR #324, opened 2023-08-24
Fix the codecov badges on the readme

jimhester:
It seems the badge url should _not_ have an app subdomain, while the link does.


hadley:
Thanks!

PR #328, opened 2023-08-28
fix Ops signature; move matrixOps out of Ops group

### LINKED ISSUE #321: `Ops.S7_object`  containing `%*%`

t-kalinowski:
This PR: 

1. Changes the S7-internal Ops group generics signature from `x, y` to `e1, e2`, matching base R.
2. Moves matrixOps generics out of the Ops dispatch code path. `matrixOps.S7_object` is no longer an alias of `Ops.S7_object`, and `Ops.S7_object` no longer dispatches `%*%`.

closes #321

### DIFF
--- a/R/generic-spec.R
+++ b/R/generic-spec.R
@@ -17,6 +17,8 @@ as_S3_generic <- function(x) {
     name <- find_base_name(x)
     if (name %in% names(base_ops)) {
       return(base_ops[[name]])
+    } else if (name %in% names(base_matrix_ops)) {
+      return(base_matrix_ops[[name]])
     } else if (!is.na(name) && is_internal_generic(name)) {
       return(S3_generic(x, name))
     }
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -118,6 +118,11 @@ matrixOps.S7_object <- function(x, y) {
   base_matrix_ops[[.Generic]](x, y)
 }
 
+if(getRversion() < "4.4.0")
+matrixOps.S7_object <- function(e1, e2) {
+  base_matrix_ops[[.Generic]](e1, e2)
+}
+
 #' @rawNamespace if (getRversion() >= "4.3.0") S3method(chooseOpsMethod, S7_object)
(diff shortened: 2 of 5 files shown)


t-kalinowski:
Looks like one small patch in the tools package will be necessary before S7 can pass R CMD check with the new, updated (and correct) matrixOps signature. https://github.com/r-devel/r-svn/pull/134/files The change is confined to the static-analysis checks only - both `Ops` and `matrixOps` groups match arguments positionally only when evaluated normally, with argument names ignored.

hadley (approved):

hadley:
Why are we seeing build failures on the latest release?

t-kalinowski:
The {tools} package was enforcing the wrong signature for matrixOps methods. This is fixed and will be part of R 4.4.0 (afaik)( https://github.com/r-devel/r-svn/pull/134, https://github.com/r-devel/r-svn/commit/b0abc0edbe4ae1a61ee95bb1f049014b8738a2dd). 

To avoid the R CMD check warning with the current release and older versions of R, I've added a special case that will define `matrixOps.S7_object` with the "wrong" signature for R versions prior to 4.4.0.

hadley:
Perfect, thanks!

PR #332, opened 2023-09-04
Update URLs in DESCRIPTION

jonthegeek:
Note: The pkgdown site link is broken on CRAN. Is it possible to set up a redirect on github pages to the new URL? I think you'd have to create a new repository with that name, which would break the OTHER redirect, so probably not worth it.


hadley:
We'll probably do another CRAN release reasonably soon, so unfortunately I think it's easiest to just let the existing site break.

Issue #275, opened 2023-02-12
Property validation


kkmann:
Dear all,

I was wondering whether it would be feasible to add optional constraints to property definitions like
```
myclass <- new_class("myclass", properties = list(
  a = class_double(min = 0, na_ok = FALSE)
))
```
Ideally, this would then also be picked up by roxygen2 when documenting class properites. Might be out of scope of a OOP system, but would be kind of handy to have...

Alternatively, do you see a way of standardizing input checking like this in the validator to autogenerate documentation of the constraints?

lawremi:
There could be convenience functions for constructing properties with constraints like this. There's precedent for this sort of thing in property implementations from other languages.

hadley:
We'd also need to add the idea of a property `validator`, but I think that makes sense.

Would be worth considering whether we want to provide canned property classes that represent scalars (but that's a fairly big design problem, so might be worth leaving that to a package).

jl5000:
In my own use of `S7` I would certainly find this very useful. I have implemented a set of validator functions dependent on the type of property:

* Min/max length of atomic vectors/lists
* Min/max character length of strings
* Min/max values of numerics
* Ensuring character properties fit regex patterns
* Ensuring character properties take one of a number of set values
* Ensuring lists contain elements of acceptable classes

Incorporating any of these into `S7` would be useful.

PR #336, opened 2023-09-08
Refactor property type validation

hadley:
In preparation for #275

And improve validation for properties with custom getters and setters.

### DIFF
--- a/R/valid.R
+++ b/R/valid.R
@@ -104,16 +104,12 @@ validate_properties <- function(object, class) {
   errors <- character()
 
   for (prop in class@properties) {
-    # Only validate static properties
-    if (!is.null(prop$getter) || !is.null(prop$setter)) {
+    # Don't validate dynamic properties
+    if (!is.null(prop$getter)) {
       next
     }
 
-    value <- prop(object, prop$name)
-    if (!class_inherits(value, prop$class)) {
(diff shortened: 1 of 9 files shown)


R/valid.R
```diff
@@ -104,16 +104,13 @@ validate_properties <- function(object, class) {
   errors <- character()
 
   for (prop in class@properties) {
-    # Only validate static properties
-    if (!is.null(prop$getter) || !is.null(prop$setter)) {
```
DavisVaughan, line 108: Was this basically just wrong before? i.e. if there is _only_ a setter (for whatever reason, maybe deprecation idk) then you should still be able to validate it here

hadley: Yeah, I think so.

tests/testthat/test-valid.R
```diff
@@ -38,6 +38,26 @@ test_that("validate checks base type", {
...
+      setter = function(self, value) {
+        self@x <- as.character(value)
+        self
+      }
+    )
+  ))
+  expect_snapshot(foo(x = 123), error = TRUE)
+})
+
+test_that("validate does not check type of getters", {
```
DavisVaughan, line 54: Maybe a note about why? i.e. something like "because getters can be arbitrarily expensive, so we require the class author to uphold this"

DavisVaughan (approved):

PR #339, opened 2023-09-09
Implement per-property validators

### LINKED ISSUE #275: Property validation

hadley:
Fixes #275

@lawremi this idea and implemented seemed simple enough that I just implemented it. Do you think this needs to be discussed with the committee before merging?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # S7 (development version)
 
+* `new_property()` gains a `validator` argument that allows you to specify
+  a per-property validator (#275).
+
 * `new_object()` works better when custom property setters modify other 
   properties.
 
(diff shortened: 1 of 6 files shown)


t-kalinowski:
Was reviewed and approved at OOP-WG meeting on 2023-09-11.

DavisVaughan (commented):
Delayed approval, but neat idea! ✅

Issue #250, opened 2022-08-14
How to create a nestable R7 class?


KatherineCox:
Sometimes it's useful to be able to nest objects of the same type inside each other.  Is there a recommended way for doing this with R7?

##  Naive approach
It's good that this errors; if you ever tried to instantiate this you'd end up in an infinitely recursing loop.

``` r
nested <- new_class("nested",
  properties = list(child = nested)
)
#> Error in as_properties(properties): object 'nested' not found
```

<sup>Created on 2022-08-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## `new_union()`
It would be nice if this worked, allowing you to specify at least one other type that can be a "base case". Preferably while still protecting you from infinite recursion by not permitting the nested class to be the default.

``` r
nested <- new_class("nested",
  properties = list( child = new_union(NULL, nested) )
)
#> Error in lapply(x, as_class): object 'nested' not found
```

<sup>Created on 2022-08-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## Define it twice

This doesn't error and lets me nest objects, but `properties` holds on to the original class definition (`inner`'s child has no `name` or `child` properties, since the original definition of `nested` had no properties).  I don't know what problems that might cause (and it prevents the infinite recursion problem), but I don't like the feel of it.

``` r
nested <- new_class("nested")

nested <- new_class("nested",
  properties = list( 
    name = class_character,
    child = nested )
)

inner <- nested()
inner
#> <nested>
#>  @ name : chr(0) 
#>  @ child: <nested>

outer <- nested(name = "outer", child = inner)
outer
#> <nested>
#>  @ name : chr "outer"
#>  @ child: <nested>
#>  .. @ name : chr(0) 
#>  .. @ child: <nested>
```

<sup>Created on 2022-08-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

# Use a parent class

This seems to work and I haven't thougth of anything that would go wrong.  It's mildly annoying to have the extra parent class cluttering things up if I don't have any other reason for it to exist, but I can live with it :)

``` r
nested_parent <- new_class("nested_parent")

nested <- new_class("nested", parent = nested_parent,
  properties = list( 
    name = class_character,
    child = nested_parent )
)

inner <- nested()
inner
#> <nested>
#>  @ name : chr(0) 
#>  @ child: <nested_parent>

outer <- nested(name = "outer", child = inner)
[...]

PR #341, opened 2023-09-09
External S7 classes

### LINKED ISSUE #317: Extending a class from another package
Currently if you extend a class from another package, the parent class will be captured at build-time. So if that package is later updated, and your package is not re-build, it's possible for two inconsistent versions of a class to exist. (The methods will be fine since they're already registered dynamically if needed, but the properties and validator might not).

We might be able to fix this and #250 at the same time by creating something like `new_external_class()` which contains just enough information to perform minimal static checks with the actual invocation always happening at run time. This will require some careful thought and analysis.

hadley:
Prior to this PR, if class A in package A extends class B in package B, class A will include a copy of the definition of class B at build-time. Methods are not also copied, so if class B is later updated and package B reinstalled, it's possible to end up with in an inconsistent state where you have the class definition from version 1 and the methods from version 2.

This PR adds a new "external class" object designed to ameliorate this problem by doing more work at run-time rather at build-time. For example, the constructor of class B will now have arguments `...` and it will figure out which arguments go to which constructor at run-time (we prefer to do this at build time in order to generate a constructor with named arguments that can be auto-completed etc).

Fixes #317


hadley:
Since I've now realised that #250 admits a much simpler solution, I think this should take on more of a shape of `new_external_generic()`, with an explicit package argument. I think this will still need to switch to a dynamic constructor (i.e. take `...` and do argument dispatch at run time) but hopefully that won't require too much refactoring of the existing code to determine which arguments go where.

hadley:
To do:

* [ ] Figure out how to test this
* [ ] Avoid inlining the parent properties at build-time, possibly by setting to some sentinel value that indicates they're dynamic and then replacing `class@properties` with a call to a helper.

hadley:
@DavisVaughan @t-kalinowski this will need discussion with the full committee before we merge it, but I'd love to get your initial thoughts on both the overall idea and the implementation.

R/class.R
```diff
@@ -113,6 +113,14 @@ new_class <- function(
   # Don't check arguments for S7_object
   if (!is.null(parent)) {
     check_can_inherit(parent)
+
+    # Automatically use an external class if appropriate
+    if (!is.null(package) && is_class(parent)) {
+      if (!is.null(parent@package) && !identical(parent@package, package)) {
```
DavisVaughan, line 119: Does it need to be a nested if?

R/class.R
```diff
@@ -128,30 +136,44 @@ new_class <- function(
   }
 
   # Combine properties from parent, overriding as needed
```
DavisVaughan, line 138: This comment may be better placed inside `inherit_properties()` now

R/external-class.R
```diff
@@ -0,0 +1,72 @@
...
+
+  if (is.null(constructor_fun)) {
+    constructor_fun <- function() getExportedValue(package, name)
+  } else {
+    check_function(constructor_fun, alist())
+  }
+
+  constructor <- constructor_fun()
+  if (!is_class(constructor)) {
+    stop("`constructor_fun()` must yield an S7 class")
```
DavisVaughan, line 47: This might be a fairly esoteric error if you don't actually provide `constructor_fun` and instead provided `package` and `name` incorrectly

DavisVaughan, line 47:
```suggestion
    stop("`constructor_fun()` must return an S7 class")
```

R/external-class.R
```diff
@@ -0,0 +1,72 @@
+#' Classes from other packages
+#'
+#' @description
+#' You need an explicit external class when you want extend a class defined in
```
DavisVaughan, line 4:
```suggestion
#' You need an explicit external class when you want to extend a class defined in
```

R/external-class.R
```diff
@@ -0,0 +1,72 @@
...
+#' another package. An external class ensures that the class definition from
+#' the other package is not literally inlined in your package, ensuring that
+#' when the other package changes your package doesn't need to be rebuilt to
+#' get those changes.
+#'
+#' [new_class()] will automatically convert an S7 class to an external class
```
DavisVaughan, line 10: It automatically converts the _parent_ right?

R/property.R
```diff
@@ -316,7 +317,7 @@ prop_names <- function(object) {
     c("name", "parent", "package", "properties", "abstract", "constructor", "validator")
   } else {
     class <- S7_class(object)
-    props <- attr(class, "properties", exact = TRUE)
+    props <- attr(class, "properties", exact = TRUE)()
```
DavisVaughan, line 320: Is it worth adding `class_properties()` or something similar that does this, and maybe adds a comment about the fact that properties are dynamic to support external classes? I think this is called in at least 3 places

tests/testthat/_snaps/class.md
```diff
@@ -17,21 +17,7 @@
...
-       ..  ..$ validator: NULL
-       ..  ..$ default  : NULL
-       .. $ y: <S7_property> 
-       ..  ..$ name     : chr "y"
-       ..  ..$ class    : <S7_base_class>: <integer>
-       ..  ..$ getter   : NULL
-       ..  ..$ setter   : NULL
-       ..  ..$ validator: NULL
-       ..  ..$ default  : NULL
+       @ properties : function ()
```
DavisVaughan, line 20: I guess this reduction in info is expected because it is dynamic now?

tests/testthat/test-external-class.R
```diff
@@ -0,0 +1,55 @@
...
+  foo1 <- new_class(
+    "foo",
+    properties = list(x = class_double),
+    validator = function(self) {
+      if (self@x < 0) {
+        "@x must be positive"
+      }
+    }
+  )
+  expect_error(foo2(x = -1), "invalid")
```
DavisVaughan, line 37: Worth a snapshot or no?

R/constructor.R
```diff
@@ -43,6 +43,30 @@ new_constructor <- function(parent, properties) {
...
+    args <- list(...)
+    parent_args <- dynamic_args(args, args_info$parent)
+    self_args <- dynamic_args(args, args_info$self)
+
+    parent_obj <- do.call("parent_class", parent_args)
+    do.call("new_object", c(list(parent_obj), self_args))
+  }
+}
+dynamic_args <- function(args, selected) {
+  missing <- setdiff(selected, names(args))
```
DavisVaughan, line 63: It took me a few minutes to figure this out, maybe add a comment like

```r
# Pad `args` with currently missing arguments
```

DavisVaughan (approved):

t-kalinowski (commented):
The motivation for this PR very clear - the current situation is problematic and can easily lead to problems as packages get updated in a user library. The proposed implementation comes w/ some tradeoffs, (most of which can be mitigated w/ more code). It's worth spelling some of these out: 

- Needing to call `class@properties()` as a function is inconsistent w/ how the other `class@` attributes are accessed. A new user might even call it surprising, (and also, syntactically noisy).  Because this is a user-facing interface, I think we should explore alternative implementations that still lets us inject dynamic properties as needed, w/o converting this property to a closure. 

- The loss of args names in constructor functions in favor of `...` is, I think, a *big* loss. It means users lose interactive autocomplete suggestions (w/o substantial work to compensate in the IDE) E.g., i imagine across tidyverse/tidymodels, many constructors will live in external packages.
- Looking at the example in `new_external_class()` raises a few questions (mostly related to `print()`):
```
> foo <- new_class("foo", properties = list(x = class_integer))
> foo_ex <- new_external_class("S7", "foo", function() foo)
> foo2 <- new_class("foo", parent = foo_ex)
> foo2
<foo> class
@ parent : <S7::foo>
@ constructor: function(...) {...}
@ validator : <NULL>
@ properties :
 $ x: <integer>
> foo2()
<foo>
 @ x: int(0)
```

- Should a properties origin be presented here. When I'm working w/ a subclassed class, I typically want to quickly know what is "new" or "different" from the base class, and having that information presented by default `print()` would go a long way.
- The `...` signature of the constructor is unfortunate. Would it make sense to include at least the "new" constructor args in the signature?  For example, if foo2 had a property  `y`,  defined as `new_class("foo", properties(list(y = class_character)), parent = foo_ex)`, then perhaps the constructor signature could be `(..., y)`, with `...` being forwarded to the parent constructor. 
- (side questions, not directly related to this PR: why are we presenting an empty class validator property?)

Some stray thoughts:
- Would it make sense to convert `@properties` to a locked and sealed environment, rather than a list? Then the environment could be populated once via a call in `.onLoad()`, similar to the way we recommend registering S7 methods. (also, opens the door for active bindings)
- Could other work be done in `.onLoad()`? Perhaps, building the default constructor with the proper signature? Or catch and warn about edge cases when `pkgA` *must* be reinstalled due to an update in `pkgB`?

PR #343, opened 2023-09-11
Store the S7 methods table as an attribute of the S3 methods table

### LINKED ISSUE #342: length method vs R CMD Check
I'm not sure how to reprex this, because so far I only see the issue inside R CMD Check. This is the warning I get:

> ❯ checking for code/documentation mismatches ... WARNING
> Warning in formals(fun) : argument is not a function

I got it by trying to define a `length` method.

`S7::method(length, rapid) <- function(x) length(x@info)`

It otherwise works with the dev version of S7 (but fails with the CRAN version).

I'm assuming that it's somehow related to length being a primitive rather than an actual S3 generic, but beyond that I'm not sure where to look.

hadley:
This prevents it from being listed in `functions_in_S3_table` (https://github.com/wch/r-source/blob/16c4fbf48efb4d43ef5dccadc86576100edd85b4/src/library/tools/R/QC.R#L370-L383) and later generating a warning when calling `formals` on each element of that list.

Fixes #342

### DIFF
--- a/vignettes/packages.Rmd
+++ b/vignettes/packages.Rmd
@@ -45,7 +45,7 @@ If you export a class, you must also set the `package` argument, ensuring that c
 
 You should document generics like regular functions (since they are!).
 If you expect others to create their own methods for your generic, you may want to include an section describing the the properties that you expect all methods to have.
-We plan to provide an easy way to list all methods for a generic, but have not yet implemented it.
+We plan to provide a an easy way to all methods for a generic, but have not yet implemented it.
 You can track progress at <https://github.com/RConsortium/OOP-WG/issues/167>.
 
 We don't currently have any recommendations on documenting methods.
--- a/vignettes/packages.Rmd
+++ b/vignettes/packages.Rmd
@@ -45,7 +45,7 @@ If you export a class, you must also set the `package` argument, ensuring that c
 
 You should document generics like regular functions (since they are!).
 If you expect others to create their own methods for your generic, you may want to include an section describing the the properties that you expect all methods to have.
-We plan to provide a an easy way to all methods for a generic, but have not yet implemented it.
+We plan to provide an easy way to list all methods for a generic, but have not yet implemented it.
 You can track progress at <https://github.com/RConsortium/OOP-WG/issues/167>.
 
 We don't currently have any recommendations on documenting methods.
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # S7 (development version)
 
+* External methods are now registered using an attribute of the S3 methods 
+  table rather than an element of that environment. This prevents a warning
+  being generated during the "code/documentation mismatches" check in
+  `R CMD check` (#342).
+
 * `class_missing` and `class_any` can now be unioned with `|` (#337).
 
 * `new_object()` works better when custom property setters modify other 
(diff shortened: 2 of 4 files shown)


hadley:
I'm pretty sure that this is a backward incompatible change since previously built S7 packages will have stored their registered methods in a place that we no longer look. But I think this breakage is ok because there are CRAN packages that use S7 yet, and we haven't publicly announced it.

jonthegeek:
I'm working on a reprex to locate it exactly, but having a mix of methods-as-traditional-S3s and methods-as-S7::method()<- caused issues with this fix. I can't tell yet if that would be a problem.

(I replaced one method to see if this fixed the issue I was seeing, and then had issues with all the other methods)

jonthegeek:
This is... odd.

See the latest version of https://github.com/jonthegeek/S7primitivereprex

Specifically, this passes: https://github.com/jonthegeek/S7primitivereprex/blob/main/tests/testthat/test-my_class.R

But it fails if I run it outside of the test context:

``` r
pkgload::load_all()
my_other_letters <- my_other_class(letters)
expect_equal(length(my_other_letters), 2)
#> Error: length(my_other_letters) (`actual`) not equal to 2 (`expected`).
#> 
#>   `actual`: 26
#> `expected`:  2
```

It's only for the method that's defined as a "normal" S3 method, but the fact that it's invisible to the test is confusing.

DavisVaughan (approved):
LGTM if we have to do it this way, but here's another idea:

According to `?.onLoad`:

> loadNamespace looks for a hook function named .onLoad and calls it (with two unnamed arguments) before sealing the namespace and processing exports.

So `methods_register()` is called before the namespace is sealed. So is it possible we could just insert `.__S7_methods__.` into the package's namespace itself as an unexported object rather than trying to rely on something that already exists?

hadley:
@DavisVaughan I'm pretty sure I tried that first but I hit namespace is locked errors.

t-kalinowski (approved):
LGTM! I like the rename of `external_methods_get()` to `S7_methods_table()`.

t-kalinowski:
@DavisVaughan The namespace is already locked when `.onLoad()` is called (i.e., removal or addition of new bindings is forbidden, but changing of binding values is still permitted). After .onLoad() is called, then the bindings themselves are locked.

We could document users to define a `.S7_methods` object, or maybe even do so automatically w/ roxygen/Description, but this solution seems clean and pragmatic to me. (🤞 no other problems emerge from attaching attributes to the `.__S3MethodsTable__.` env.)

vignettes/packages.Rmd
```diff
@@ -45,7 +45,7 @@ If you export a class, you must also set the `package` argument, ensuring that c
 
 You should document generics like regular functions (since they are!).
 If you expect others to create their own methods for your generic, you may want to include an section describing the the properties that you expect all methods to have.
-We plan to provide a an easy way to all methods for a generic, but have not yet implemented it.
+We plan to provide an easy way to list all methods for a generic, but have not yet implemented it.
 You can track progress at <https://github.com/RConsortium/OOP-WG/issues/167>.
```
jonthegeek, line 49:
```suggestion
You can track progress at <https://github.com/RConsortium/S7/issues/167>.
```

hadley:
@jonthegeek yeah, I can't immediately see why that wouldn't work, but due to the general weirdness of S3 method lookup/registration, it doesn't surprise me that it doesn't fully for you. I don't think it's related to this issue.

PR #345, opened 2023-09-11
Disallow `NULL` as a `.parent` in `new_object()`

DavisVaughan:
I think this tightens things up a bit. It is technically breaking as `NULL` was allowed before. I think we should either do this or set `.parent = S7_object()` as the default (which would parallel `parent = S7_object` as the default in `new_class()`).

If we did `.parent = S7_object()`, then I'd be ok with this optional argument coming before the named `...` in this instance because if you have a `.parent` that takes arguments, then those arguments would be supplied _before_ the class arguments when the constructor is generated, so you end up with something like:

```r
# good
my_class <- function(x, y) {
  new_object(parent(x = x), y = y)
}

# harder to read
my_class <- function(x, y) {
  new_object(y = y, .parent = parent(x = x))
}
```

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -245,7 +245,8 @@ new_object <- function(.parent, ...) {
   args <- list(...)
   nms <- names(args)
 
-  object <- .parent %||% class_construct(class@parent)
+  # TODO: Some type checking on `.parent`?
+  object <- .parent
   attr(object, "S7_class") <- class
   class(object) <- class_dispatch(class)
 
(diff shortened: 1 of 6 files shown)


R/class.R
```diff
@@ -245,7 +245,8 @@ new_object <- function(.parent, ...) {
   args <- list(...)
   nms <- names(args)
 
-  object <- .parent %||% class_construct(class@parent)
+  # TODO: Some type checking on `.parent`?
+  object <- .parent
```
hadley, line 249: Could maybe use something based on `check_can_inherit()`? But I think it's also fine to leave for this PR, since generally we done little to support custom constructors.

hadley (approved):

PR #349, opened 2023-09-12
Alternative backward compatibility approach for user code

### LINKED ISSUE #326: Explain that require(S7) will cause R CMD Check WARNING & NOTE
https://github.com/RConsortium/OOP-WG/blob/965ab8ff1da9a599d494ed10dee033fa27a99e4c/vignettes/packages.Rmd#L77

The wording here seems a bit too gentle. It generates a WARNING and a NOTE in R CMD Check (rightly), which might be more than users expect.

```
* checking dependencies in R code ... WARNING
Warning: 'library' or 'require' call not declared from: ‘S7’
'library' or 'require' call to ‘S7’ in package code.
  Please use :: or requireNamespace() instead.
  See section 'Suggested packages' in the 'Writing R Extensions' manual.
```

```
* checking R code for possible problems ... NOTE
File ‘rapid/R/zzz.R’:
  .onAttach calls:
    require(S7)
Package startup functions should not change the search path.
See section ‘Good practice’ in '?.onAttach'.
```

hadley:
Fixes #326


jonthegeek (commented):
This looks much friendlier!

t-kalinowski (approved):
This is a very nice approach. Some pros: 
- It doesn't expand the exported API footprint
- It doesn't increase the new-user learning burden nor require any extra work from users
- Improving over the previous hack, it works even if S7 is not attached.
- No deprecation notices will be required once this backwards compatibility hack will no longer be necessary; it will just fade away naturally with time.

Very nice!

R/property.R
```diff
@@ -260,28 +260,16 @@ prop_label <- function(object, name) {
   sprintf("%s@%s", if (!is.null(object)) obj_desc(object) else "", name)
 }
 
-#' @rdname prop
-#' @usage object@name
-#' @aliases @
-#' @rawNamespace if (getRversion() >= "4.3.0") S3method(base::`@`, S7_object) else export("@")
-`@.S7_object` <- prop
-
 # Note: we need to explicitly refer to base with "base::`@`" in the
```
DavisVaughan, line 263: I think you'll want to move this comment now

DavisVaughan (approved):

R/property.R
hadley: This PR moves it to be above the `@rawNamespace`. Do you think it should go elsewhere?

DavisVaughan: Oh ignore me, read it wrong

hadley:
No concerns in WG call 2023-10-16

PR #354, opened 2023-09-14
Pull out operator methods into own file

hadley:
And define more dynamically

### DIFF
--- a/R/method-ops.R
+++ b/R/method-ops.R
@@ -1,0 +1,38 @@
+base_ops <- NULL
+base_matrix_ops <- NULL
+
+on_load_define_ops <- function() {
+  base_ops <<- lapply(
+    setNames(, group_generics()$Ops),
+    new_generic,
+    dispatch_args = c("e1", "e2")
+  )
+  base_matrix_ops <<- lapply(
+    setNames(, group_generics()$matrixOps),
+    new_generic,
+    dispatch_args = c("x", "y")
+  )
+}
+
+#' @export
+Ops.S7_object <- function(e1, e2) {
+  base_ops[[.Generic]](e1, e2)
+}
+
+#' @rawNamespace if (getRversion() >= "4.3.0") S3method(chooseOpsMethod, S7_object)
+chooseOpsMethod.S7_object <- function(x, y, mx, my, cl, reverse) TRUE
+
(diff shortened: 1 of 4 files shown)


R/zzz.R
```diff
@@ -134,7 +108,10 @@ chooseOpsMethod.S7_object <- function(x, y, mx, my, cl, reverse) TRUE
 }
 
 .onLoad <- function(...) {
-  ## "S4"   or [in R-devel 2023-07-x]   "object"
+  on_load_define_ops()
```
hadley, line 111: If you like this pattern, I'll do the rest of `.onLoad()` so the bulk on the code lives close to where it's used.

R/method-ops.R
```diff
@@ -0,0 +1,38 @@
...
+  base_ops[[.Generic]](e1, e2)
+}
+
+#' @rawNamespace if (getRversion() >= "4.3.0") S3method(chooseOpsMethod, S7_object)
+chooseOpsMethod.S7_object <- function(x, y, mx, my, cl, reverse) TRUE
+
+#' @rawNamespace if (getRversion() >= "4.3.0") S3method(matrixOps, S7_object)
+matrixOps.S7_object <- NULL
+
+on_load_define_matrixOps <- function() {
```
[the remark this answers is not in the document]
t-kalinowski: With both approaches, building under one version of R and installing in another, the S7 package would still work, it just wouldn't pass R CMD check.

R/zzz.R
t-kalinowski: I have no preference, both work for me :) 

One minor thought, I originally wrote the code so that, eventually, when we drop support for R 4.3.0, the diff could just be deleting the block of lines (i.e., it was a self-contained block)

t-kalinowski (approved):

R/method-ops.R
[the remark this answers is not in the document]
hadley: Oh yeah, of course, because you never name the arguments to operators.

R/zzz.R
hadley: I think that's a reasonable goal, but the string "4.3.0" now appears in 7 files, so I don't think it's feasible any more.

PR #360, opened 2023-09-15
Refactor Ops tests

hadley:


tests/testthat/test-method-ops.R
```diff
@@ -1,60 +1,38 @@
 test_that("Ops generics dispatch to S7 methods", {
+  on.exit(S4_remove_classes(c("Foo")))
+  FooS4 <- methods::setClass("Foo", contains = "character")
+
+  ClassX <- new_class("ClassX")
+  method(`+`, list(class_any, ClassX)) <- function(e1, e2) "class_any + ClassX"
+  method(`+`, list(ClassX, class_any)) <- function(e1, e2) "ClassX + class_any"
+
+  expect_equal(ClassX() + ClassX(), "ClassX + class_any")
```
t-kalinowski, line 9: I think it's still worthwhile to include some of the edge-cases in the tests, in particular, testing dispatch with:
- `NULL`
- `Sys.time()` (there are some code paths specific to dispatching POSIXct in R sources)
- A custom S3 object, `structure("", class = "foo")`

Also, might be worthwhile to add test cases for:
- A custom S3 object, with it's own competing S3 Ops method like `+.foo <- \() "should not run"`
- `Sys.Date()`

t-kalinowski (approved):

tests/testthat/test-method-ops.R
hadley: Can you give me a bit more info for each of these cases so I can record that knowledge in a comment?

t-kalinowski: Maybe something like:
> In R's C sources `DispatchGroup()` has some checks/code-paths for `POSIXt`, `Date` and `Difftime` that precede the `chooseOpsMethod()` fallback. We include tests for these specific classes to make sure S7 methods still take precedence. 
https://github.com/wch/r-source/blob/5cc4e46fc665bc67be94dddd615b1709473b66a0/src/main/eval.c#L4242C1-L4247C64

(Not sure if `NULL` or competing S3 Ops method need a comment... I see that more comprehensively testing NULL already flushed out a bug 😂)

tests/testthat/test-method-ops.R
```diff
@@ -1,60 +1,87 @@
...
+  method(`+`, list(class_factor, foo)) <- function(e1, e2) "factor-foo"
+  method(`+`, list(foo, class_factor)) <- function(e1, e2) "foo-factor"
+
+  expect_equal(foo() + factor(), "foo-factor")
+  expect_equal(factor() + foo(), "factor-foo")
+
+  # Even if custom method exists
+  foo_S3 <- structure(list(), class = "foo_S3")
+  assign("+.foo_S3", function(e1, e2) stop("Failure!"), envir = globalenv())
+  on.exit(rm("+.foo_S3", envir = globalenv()))
```
t-kalinowski, line 31: Does the `on.exit(add=FALSE)` default clear out the `defer()` call in the `local_methods()` call above?

hadley: ooh yes

PR #365, opened 2023-09-16
Explore group generics implementation

### LINKED ISSUE #353: Group generics
Related to #320, we need some way to register methods for group generics. I think the syntax should probably look something like this?

``` r
library(S7)
foo <- new_class("foo", class_integer)
method(Math, foo) <- function(x) {
   foo(Math(super(x, class_integer)))
}
method(Ops, list(foo, class_any)) <- function(x, y) {
   foo(Ops(super(x, class_integer), y))
}
method(Ops, list(class_any, foo)) <- function(x, y) {
   foo(Ops(x, super(y, class_integer)))
}
```

hadley:
Fixes #353


R/method-group.R
```diff
@@ -0,0 +1,40 @@
...
+  })
+}
+
+#' @export
+Math.S7_object <- function(x, ...) {
+  group_generic_Math(x, ..., .Generic = .Generic)
+}
+
+#' @export
+Ops.S7_object <- function(e1, e2) {
```
hadley, line 21: Need to move this back to `method-ops` since `Ops` is genuinely different to the other group generics because it's the only one where we want to override every specific generic with an S7 generic in order to get S7 multi-dispatch.

hadley:
Finish after #382 since we need that fallback technique.

hadley:
@t-kalinowski @DavisVaughan before I finish up the implementation for the other group generics, could you please take a look at how it works for Math?

R/method-group.R
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#' @export
+Math.S7_object <- function(x, ...) {
+  tryCatch(
+    return(S7_Math(x, ..., .Generic = .Generic)),
+    S7_error_method_not_found = function(cnd) NULL
+  )
+
+  NextMethod()
```
DavisVaughan, line 73: Do we need to call `NextMethod()`? Isn't `S7_object` always the right-most class? It would be nice to throw the S7 method not found error if we could

DavisVaughan (approved):
Looks nice to me!

R/method-group.R
```diff
@@ -0,0 +1,85 @@
...
+    S7_dispatch()
+  })
+
+  S7_Summary <<- new_generic("Summary", "x", function(x, ..., na.rm = FALSE, .Generic) {
+    S7_dispatch()
+  })
+}
+
+#' @export
+Math.S7_object <- function(x, ...) {
```
DavisVaughan, line 67: Do you also need to check for a "specific" generic call here?

t-kalinowski (approved):
This looks like a very nice solution. I see no issues.

mmaechler:
Sorry if I'm a "game spoiler":  Using  `tryCatch()` in such fundamental group's methods will I think be quite a performance killer e.g. for simple arithmetic / math operations.
Is that really really necessary?

hadley:
If you can suggest an alternative implementation, I'm happy to try it. But I couldn't think of another way to do it.

t-kalinowski:
One approach might update `method_` and `method_call_` (c functions), so that, rather than than signaling an error directly, they return a sentinel value like `NULL`, or perhaps, the error call itself (i.e., calling `eval()` on the return object signals the error). Then, instead of calling `S7_dispatch()` in the `S7_Math()` generic, we could inspect for the sentinel instead. For example, `S7_Math()` could be defined like this:
```r
S7_Math <<- new_generic("Math", "x", function(x, ..., .Generic) {
  cl <- .Call(method_call_, sys.call(-1), sys.function(-1), sys.frame(-1))
  if(is.null(cl)) 
    NextMethod()
  eval(cl)
})
```

mmaechler:
> One approach might be to update `method_` and `method_call_` (c functions), so that, rather than than signaling an error directly, they return a sentinel value like `NULL`, or perhaps, the error call itself (i.e., calling `eval()` on the return object signals the error). Then, instead of calling `S7_dispatch()` in the `S7_Math()` generic, we could inspect for the sentinel instead.  
> For example ...

"Nice"!  That would almost surely be an order of magnitude faster!

Possibly we'd even want to provide (and document and use ourselves) a thin wrapper around `.Call(..)` such that S7-users
could make use of this as well in their own method definitions.
{{I do acknowledge I haven't studied what you are doing exactly here, so "I have no clue" ;-)}}

hadley:
Ok, I'll take a stab at that approach.

PR #369, opened 2023-09-27
Mention property validation in vignette

hadley:


vignettes/classes-objects.Rmd
```diff
@@ -220,7 +259,10 @@ You can see the source code for a class's constructor by accessing the `construc
 range@constructor
 ```
 
-In most cases, S7's default constructor will all you need. However, in some cases you might want something custom. For example, for our range class, maybe we'd like to construct it from a vector of numeric values, automatically computing the min and the max. To implement this we could do:
+In most cases, S7's default constructor will all you need.
```
jonthegeek, line 262:
```suggestion
In most cases, S7's default constructor will be all you need.
```

vignettes/classes-objects.Rmd
```diff
@@ -236,6 +278,9 @@ range <- new_class("range",
 range(c(10, 5, 0, 2, 5, 7))
 ```
 
-A constructor must always end with a call to `new_object()`. The first argument to `new_object()` should be an object of the `parent` class (if you haven't specified a `parent` argument to `new_class()`, then you should use `S7_object()` as the parent here). That argument should be followed by one named argument for each property.
+A constructor must always end with a call to `new_object()`.
+The first argument to `new_object()` should be an object of the `parent` class (if you haven't specified a `parent` argument to `new_class()`, then you should use `S7_object()` as the parent here).
+That argument should be followed by one named argument for each property.
 
 There's one drawback of custom constructors that you should be aware: any subclass will also require a custom constructor.
```
jonthegeek, line 285:
```suggestion
There's one drawback of custom constructors that you should be aware of: any subclass will also require a custom constructor.
```

vignettes/classes-objects.Rmd
```diff
@@ -125,7 +128,43 @@ range <- new_class("range",
...
+### Validation
+
+You can optionally provide a validator for each property.
+For example, instead of validating the length of `start` and `end` in the validator of our `Range` class, we could implement those at the property level:
+
+```{r, error = TRUE}
+prop_number <- new_property(
+  class = class_double, 
+  validator = function(value) {
+    if (length(value)) "must be length 1"
```
DavisVaughan, line 142:
```suggestion
    if (length(value) != 1L) "must be length 1"
```

Doesn't this also fail for length 1 right now?

DavisVaughan (approved):

PR #379, opened 2023-11-22
Take list name from property name

### LINKED ISSUE #371: Don't require two property names
``` r
library(S7)
foo <- new_class("foo", properties = list(new_property(name = "bar")))
#> Error: `property[[1]]` is missing a name
```

<sup>Created on 2023-09-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Just need to check for this case in `as_property()`.

hadley:
Fixes #371


R/property.R
```diff
@@ -403,15 +403,22 @@ as_properties <- function(x) {
...
-  if (name == "") {
-    msg <- sprintf("`property[[%i]]` is missing a name", i)
-    stop(msg, call. = FALSE)
-  }
 
   if (is_property(x)) {
-    x$name <- name
+    if (is.null(x$name)) {
+      if (name == "") {
+        msg <- sprintf("`property[[%i]]` must have a name or be named.", i)
```
DavisVaughan, line 410:
```suggestion
        msg <- sprintf("`properties[[%i]]` must have a name or be named.", i)
```

I think?

R/property.R
```diff
@@ -403,15 +403,22 @@ as_properties <- function(x) {
...
+        stop(msg, call. = FALSE)
+      }
+      x$name <- name
+    }
     x
   } else {
+    if (name == "") {
+      msg <- sprintf("`property[[%i]]` must be named.", i)
```
DavisVaughan, line 418:
```suggestion
      msg <- sprintf("`properties[[%i]]` must be named.", i)
```

R/property.R
```diff
@@ -403,15 +403,22 @@ as_properties <- function(x) {
...
 
 as_property <- function(x, name, i) {
-  if (name == "") {
-    msg <- sprintf("`property[[%i]]` is missing a name", i)
-    stop(msg, call. = FALSE)
-  }
 
   if (is_property(x)) {
-    x$name <- name
+    if (is.null(x$name)) {
```
DavisVaughan, line 408: So an existing property `$name` has priority over the name supplied in the `list()` call if you supply both, right? I can't tell if that seems right or not, but don't have strong feelings either way

DavisVaughan (approved):

R/property.R
hadley: Hmmm, I think the list name should win.

DavisVaughan: That was also my gut feeling

PR #382, opened 2023-11-23
Ensure Ops falls back to base behaviour

### LINKED ISSUE #320: `base_ops` are empty  - arithmetic fails e.g.  when extending class_double
A simple class & object `x` extending double  does work  in `sin(x)`   but not with `x + 1` or  ` x < 3`,
because the (internal)  `base_ops` all have no methods defined.
It is easiest to see in an example {latest version of `S7`}:

```
> nvec <- new_class("nvec", class_double)
> (x <- nvec(c(1:3, pi, 4:7)))
<nvec> num [1:8] 1 2 3 3.14 4 ...
> sin(x)
<nvec> num [1:8] 8.41e-01 9.09e-01 1.41e-01 1.22e-16 -7.57e-01 ...
> sin(x) == sin(as.double(x))
Error: Can't find method for generic `==(x, y)`:
- x: <nvec>
- y: <double>
```
Via `traceback()` we are quickly lead that it is cause by the S7-internal  `base_ops`  being "empty", i.e., not having any methods
defined.  
```
> traceback()
5: stop(msg, call. = FALSE)
4: function (name, args, signatures) 
   ................
   ................
       list(c("nvec", "double", "S7_object"), "double"))
3: S7::S7_dispatch()
2: base_ops[[.Generic]](e1, e2)
1: Ops.S7_object(sin(x), sin(as.double(x)))

> S7:::base_ops$`==`
<S7_generic> ==(x, y, ...) with 0 methods:
```
whereas evidently sin() or sinpi() etc are all defined correctly.

hadley:
Fixes #320

### DIFF
--- a/R/method-dispatch.R
+++ b/R/method-dispatch.R
@@ -11,7 +11,8 @@ method_lookup_error <- function(name, args, signatures) {
     msg <- sprintf("Can't find method for generic `%s(%s)`:\n%s", name, arg_names, types)
   }
 
-  stop(msg, call. = FALSE)
+  cnd <- errorCondition(message = msg, class = c("methodNotFound", "error"))
+  stop(cnd)
 }
 
 #' @rdname new_generic
--- a/R/method-ops.R
+++ b/R/method-ops.R
@@ -16,7 +16,18 @@ on_load_define_ops <- function() {
 
 #' @export
 Ops.S7_object <- function(e1, e2) {
-  base_ops[[.Generic]](e1, e2)
+  cnd <- tryCatch(
+    return(base_ops[[.Generic]](e1, e2)),
+    methodNotFound = function(cnd) cnd
+  )
+
+  if (S7_inherits(e1) && S7_inherits(e2)) {
+    stop(cnd)
+  } else {
+    # Must call NextMethod() directly in the method, not wrapped in an
+    # anonymous function.
+    NextMethod()
+  }
 }
 
 #' @rawNamespace if (getRversion() >= "4.3.0") S3method(chooseOpsMethod, S7_object)
--- a/tests/testthat/test-method-ops.R
+++ b/tests/testthat/test-method-ops.R
@@ -12,6 +12,8 @@ test_that("Ops generics dispatch to S7 methods for S7 classes", {
   expect_equal(foo1() + foo2(), "foo1-foo2")
   expect_equal(foo2() + foo1(), "foo2-foo1")
   expect_equal(foo2() + foo2(), "foo2-foo2")
+
+  expect_error(foo1() + new_class("foo3")(), class = "methodNotFound")
 })
 
 test_that("Ops generics dispatch to S3 methods", {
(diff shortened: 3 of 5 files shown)


R/method-dispatch.R
```diff
@@ -11,7 +11,8 @@ method_lookup_error <- function(name, args, signatures) {
     msg <- sprintf("Can't find method for generic `%s(%s)`:\n%s", name, arg_names, types)
   }
 
-  stop(msg, call. = FALSE)
+  cnd <- errorCondition(message = msg, class = c("methodNotFound", "error"))
```
DavisVaughan, line 14:
```suggestion
  cnd <- errorCondition(message = msg, class = c("errorMethodNotFound", "error"))
```

I don't have super strong feelings about it, but in vctrs/rlang we typically include the word `error` in the class name, and we also typically include some package/org related prefix like `vctrs_error_*` or `rlib_error_package_not_found`

R/method-ops.R
```diff
@@ -16,7 +16,18 @@ on_load_define_ops <- function() {
 
 #' @export
 Ops.S7_object <- function(e1, e2) {
-  base_ops[[.Generic]](e1, e2)
+  cnd <- tryCatch(
+    return(base_ops[[.Generic]](e1, e2)),
+    methodNotFound = function(cnd) cnd
+  )
```
DavisVaughan, line 22: I tried to think of any other way to do this to avoid the non-negligible performance hit of `tryCatch()`, but I can't really think of one:
- We don't have an easy to do `findMethod(.Generic, e1 = e1, e2 = e2)` so we can't try looking it up without calling it
- I don't think we can easily use the more performant `withCallingHandlers()` without more effort, since we don't always rethrow the error

tests/testthat/test-method-ops.R
```diff
@@ -76,6 +78,19 @@ test_that("Ops generics dispatch to S7 methods for NULL", {
   expect_equal(NULL + foo(), "NULL-foo")
 })
 
+test_that("Ops generics falls back to base behaviour", {
+  local_methods(base_ops[["+"]])
+
+  foo <- new_class("foo", parent = class_double)
+  expect_equal(foo(1) + 1, foo(2))
+  expect_equal(foo(1) + 1:2, 2:3)
```
DavisVaughan, line 86: Same behavior if you flip the ordering around?

R/method-ops.R
hadley: Given the general cost of dispatch I'm not sure this is worth worrying too much about?

``` r
library(S7)

foo <- new_class("foo")
method(`+`, list(foo, foo)) <- function(e1, e2) 100
f <- foo()

bench::mark(
  f + f,
  tryCatch(f + f)
)
#> # A tibble: 2 × 6
#>   expression           min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>      <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 f + f             5.21µs   5.78µs   167089.    39.9KB     50.1
#> 2 tryCatch(f + f)   6.76µs   7.42µs   132355.    12.4KB     53.0
```

<sup>Created on 2023-11-29 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

R/method-dispatch.R
hadley: Good catch, thanks!

Issue #392, opened 2023-12-05
Levels ignored in empty default factor


jonthegeek:
My real use-case of this is more complicated, but this is enough to dig into it.

``` r
library(S7)
expected_levels <- levels(factor(levels = c("a", "b")))
class_testing <- new_class(
  name = "testing",
  properties = list(
    enum = new_property(
      class = class_factor,
      default = factor(levels = c("a", "b"))
    )
  )
)
default_levels <- levels(class_testing()@enum)
expected_levels
#> [1] "a" "b"
default_levels
#> character(0)
identical(default_levels, expected_levels)
#> [1] FALSE
```

<sup>Created on 2023-12-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I traced it to the definition of `%||%` in `utils.R` (uses `length(x) == 0`, not `is.null(x)`; it's more like the unexported `%|0|%` from rlang, rather than `%||%`).

``` r
`%||%` <- function(x, y) if (length(x) == 0) y else x
```

The given default (`factor(levels = c("a", "b"))`) *is* length 0, but it isn't `NULL`.

I'm not sure yet that this actually matters for what I'm doing, but it was surprising (and made me think I had some deeper issue).

I'd try to fix it myself, but I'm getting other errors when I run the tests, and this one seems like it could break things in surprising ways.

PR #394, opened 2023-12-07
proper `%||%`

mmaechler:
See #392  for one convincing use case  why  `%||%`  in S7  should be the same as "everywhere" else


R/generic-spec.R
```diff
@@ -111,10 +111,16 @@ group_generics <- function() {
     Summary = "Summary",
     Complex = "Complex"
   )
-
+  ## Not yet:
```
hadley, line 114: Is this related to the `%||%` change?

hadley: It's easiest to keep each PR in a separate branch to avoid that sort of problem. But I can do some surgery on this branch to just include that commit, or probably just do it from first principles since it's such a simple change.

Issue #363, opened 2023-09-16
Rewrite `prop` and `prop<-` in C


hadley:
They are called extremely frequently so I bet they're going to be a performance bottleneck.

PR #395, opened 2023-12-11
rewrite `prop()` in C

t-kalinowski:
Addresses #363


t-kalinowski:
Slightly over a 10x speed up by moving `prop()` to C.

``` r
library(S7)

foo <- new_class("foo", properties = list(xyz = class_double))
obj <- foo(1)
propr <- S7:::propr
print(plot(print(df <- bench::mark(
  c = prop(obj, "xyz"),
  r = propr(obj, "xyz")
))))
#> # A tibble: 2 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>
#> 1 c          205.01ns 287.02ns  3215873.        0B      0   10000     0
#> 2 r            3.77µs   4.14µs   227757.    39.8KB     45.6  9998     2
#> # ℹ 5 more variables: total_time <bch:tm>, result <list>, memory <list>,
#> #   time <list>, gc <list>
#> Loading required namespace: tidyr
```



<sup>Created on 2023-12-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

src/prop.c
```diff
@@ -0,0 +1,128 @@
...
+  return i == -1 ? R_NilValue : VECTOR_ELT(list, i);
+}
+
+static inline
+Rboolean has_name(SEXP list, const char* name) {
+  return (Rboolean) name_idx(list, name) != -1;
+}
+
+static inline
+Rboolean inherits2(SEXP object, const char* name) {
```
hadley, line 40: Nice idea!

src/prop.c
```diff
@@ -0,0 +1,128 @@
...
+      is_s7_class(object) && (
+          name_sym == sym_name  ||
+          name_sym == sym_parent  ||
+          name_sym == sym_package  ||
+          name_sym == sym_properties  ||
+          name_sym == sym_abstract  ||
+          name_sym == sym_constructor  ||
+          name_sym == sym_validator
+    ))
+      return value;
```
hadley, line 122: I wonder if it's worth implementing `prop_names()` and `prop_exists()` at the same time because that might influence the design of name validation?

hadley (commented):
Looks good! Do you think it's worth iterating in this PR for bit longer (including a bit of thought about how to track performance more broadly), or it's better to merge this an iterate in future PRs?

t-kalinowski:
> better to merge this an iterate in future PRs

I'd vote to merge now, since it's immediately beneficial, and then we can continue to iterate in future PRs (also, it's looking like `prop<-` is going to involve some discussion, and would probably be better in it's own PR).

hadley:
Sounds good. Can you please add a bullet to news? (We wouldn't normally because it's not user facing, but it'll remind us to discuss in the next WG meeting)

Issue #393, opened 2023-12-05
Validation vs custom properties


jonthegeek:
My mental model is that properties are responsible for validating themselves (for the most part), but the class validator can check to make sure that the properties work with one another. However, a property with a setter breaks that model.

``` r
library(S7)
class_testing <- new_class(
  name = "testing",
  properties = list(
    first = class_character,
    second = new_property(
      setter = function(self, value) {
        # Do some things to prep value, then...
        self@second <- value
        self
      }
    )
  ),
  validator = function(self) {
    message("Validating the properties against one another")
  }
)
testing <- class_testing()
#> Validating the properties against one another
testing@second <- "this does not trigger validation"
testing@first <- "this triggers validation"
#> Validating the properties against one another
```

<sup>Created on 2023-12-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Is this working as intended? Is there a way to trigger overall validation after a property that has a setter is updated?

jonthegeek:
I think MAYBE I'm supposed to manually validate in the setter in cases where I want to validate. That causes an issue during initialization, but I was able to work around it. Long-term, that feels likely to cause confusion, and I'm still not sure whether it's intentional.

hadley:
Hmmmm, not if this is a bug or not, but if it isn't we should document that you need to call `validate()` yourself if you're providing a custom setter.

jonthegeek:
It's a pain during initialization, because things "after" this property aren't set yet (so validation of the overall thing definitely can fail). I worked around it by setting an "initialized" attribute if it didn't exist (and not validating), or validating if the initialized attribute is there. Definitely hacky right now but it gets the job done.

t-kalinowski:
I think this will be fixed once https://github.com/RConsortium/S7/pull/396 is merged. The current `prop<-` method does not run `validate(object)` after running a custom setter, but it should.

t-kalinowski:
This is fixed on main now. Thanks for reporting!

Issue #251, opened 2022-08-19
Named list of property-specific validators in class constructor


mattwarkentin:
Hi,

I just watched the great talk at rstudio::conf(2022) and started playing around with `R7` a bit.

I was wondering if it perhaps makes any sense for the `validator` argument to `new_class()` to be either a single function that globally validates or alternatively a named-list of functions where each name corresponds to an item in `properties`?

Something like:

```r
range <- new_class("range",
  properties = list(
    start = class_double, 
    end = class_double
  ),
  validator = list(
    start = function(self) {
      if (length(self@start) != 1) {
        "@start must be length 1"
      } else if (self@end < self@start) {
        "@start must be less than or equal to @end"
      }
    },
    end = function(self) {
      if (length(self@end) != 1) {
        "@end must be length 1"
      } else if (self@end < self@start) {
        "@end must be greater than or equal to @start"
      }
    }
  )
)
```

This specific example may not be particularly motivating. But it has the advantage of modularity, where you visually see which part of the class constructor code validates which property, and each validation step becomes simpler to maintain (I *think*). 

Also, when only a single property is modified, such as:
```r
x <- range(start = 1, end = 10)
x@start <- 20
```
Only the `validator` for `start` would need to be called, so perhaps a tiny performance advantage over a global validator. 

I am not an expert in OOP by any means, so maybe this is not a reasonable question.

jl5000:
What about if a validation check needs to occur on a combination of values simultaneously? This partitioning then breaks down.

jl5000:
Actually, I've just realised after watching Hadley's presentation again that you could probably do this in the existing implementation. You could ignore the validator function and instead write a setter which includes the validation code for that property. That might work.

lawremi:
Since validation logic can be complex, it's generally a good idea to modularize validation functions; for example a common pattern in S4 is something like:
```
validator <- function(object) {
  c(validator_x(object), validator_y(object))
}
```
which works because `c()` will drop `NULL` values and even return `NULL` when all of the arguments are `NULL`. In principle, this even allows reuse of validation logic across classes, although that might not be common in practice.

I'm not sure there are many advantages to directly supporting per-property validation in the framework itself. It is more self-documenting, but since there would be no easy way to enforce the single-property behavior, it could become incorrect and misleading. Also, there's no way to know which property needs to be checked upon object modification, since the validity of a property can depend on the value of other properties. A compiler approach is a potential path to optimization here.

[...]

PR #398, opened 2024-01-14
Automatic Validation of Properties using Checkmate

jmiahjones:
I just learned about S7, and I was really excited! I've sketched out some code for an idea about property validation, but I wanted to start a discussion here. Some issues may be somewhat related, like #251 or #393.

## Perceived Problem

I've been using R6 classes, but my annoyance is that you have a lot of boilerplate to define the property, specify it as a function argument, and then use the `self$x <- x` pattern. That means there are many places to define properties. Necessarily, if your object has many properties then validation occurs far away from the property.

S7 has solved the first problem, but I still feel a similar problem from number 2. Consider this bit of code:

```r
has_bool <- new_class(
  "has_bool",
  properties = list(
    bool = class_logical
  ),
  validator = function(self){
    if(length(self@bool) > 1) {
      "@bool must be a single logical value"
    }
  })
```

When I define `bool`, I would ideally like it to be known that `bool` is a length one, non-missing logical. This is a toy example, but you can imagine that if the properties list gets rather large, there would be a lot of space between the validator for the property and the property itself. It seems like describing the basic expectations for the property where it is defined is a desirable defensive programming practice.

Of course, there's a couple of other solutions. You could define a new property type that handles this, or just use the validation function in the property. These implementations use the excellent `checkmate` package:

```r
library(checkmate)
has_bool2 <- new_class(
  "has_bool2",
  properties = list(
    bool = class_bool(len=1)
  ))

has_bool3 <- new_class(
  "has_bool3",
  properties = list(
    bool = new_property(
      class = class_logical,
      validator = \(value) {
        res <- check_logical(
          value,
          len = 1
        )
        if(isTRUE(res)){
          return(NULL)
        } else {
          return(res)
        }
      }
    )
  ))
```
These two are essentially the same as `class_bool` constructs the property used in implementation 3. The downside to this, is that a potentially very large number of these property factories must be created. 

## The Proposal

My personal preference is to define "expectations" for the property directly in the property constructor itself. Thinking through this, I realized that many of these "expectations" are implemented as arguments to functions in the checkmate package. That led me to sketch out this `new_checkmate_property` function that essentially constructs the type of validator above by passing `...` to a check function:

```r
has_bool4 <- new_class(
  "has_bool4",
  properties = list(
    bool = new_checkmate_property(
      class = class_logical,
      validator = check_logical,
      len = 1
    )
  ))
```

In `new_checkmate_property`, the `validator` argument corresponds to a `checkmate::check` function that is used to validate the property, and `len = 1` is parsed as a `...` to be passed to it. This lets the user describe the exact expectations for the property inputs, and the failure gets reported very quickly with an automatically-generated and descriptive error message:

```r
> h4 <- has_bool4(bool=T) # no error

> h4_fail <- has_bool4(c(T,T))
Error: <has_bool4> object properties are invalid:
- @bool Must have length 1, but has length 2
> h4_fail <- has_bool4(22.0)
Error: <has_bool4> object properties are invalid:
- @bool must be <logical>, not <double>
```

More complex and inter-related validation can still take place in the `validate` argument for the class itself. I think has_bool4 is slightly more verbose than has_bool2, but each line of the bool definition serves an explicit purpose. Perhaps the only thing that could be changed is to tie the `check` function to the class type, but this method allows more freedom to do validation like the following:

```r
has_switch <- new_class(
  "has_switch",
  properties = list(
    switch = new_checkmate_property(
      class = class_character,
      validator = check_choice,
      choices = letters[1:4]
    )
  ))
> has_switch("e")
Error: <has_switch> object properties are invalid:
- @switch Must be element of set {'a','b','c','d'}, but is 'e'
```

## Downsides

I wanted to get feedback on this idea. One of the downsides is my reliance on the `checkmate` package. I noticed that dependencies are kept out, presumably to make S7 a language feature. Perhaps this is some syntactic sugar that belongs in its own package outside of S7. I'd appreciate any thoughts!


lawremi:
Nice work. I agree that this would be great to have as a separate helper package.

jmiahjones:
Thanks, @lawremi. If that is the route this goes, that would be fine. I do want to make sure there's nothing else on the roadmap to re-work property creation before I start creating a package.

hadley:
Yeah, this looks like it would be a good fit for an external package to me too. Maybe it could go it checkmate itself? I think checkmate would only need to take a suggested dependency on S7.

jmiahjones:
Alternatively, @hadley and @lawremi what about having a suggested dependency on checkmate here, with this simple wrapper unlocked by the package?

hadley:
@jmiahjones I don't think we'd want to "bless" checkmate in that way without an extensive comparison of the options.

tdeenes:
> @jmiahjones I don't think we'd want to "bless" checkmate in that way without an extensive comparison of the options.

I would rather argue that this is not about "blessing" any package. S7 should be as lean as possible without introducing any dependencies (even suggested ones) over the ones which are strictly necessary to the core functionality of the pkg or to unit tests. To my taste even the current list of suggested dependencies is a bit too long. As for validations, S7 already provides the basics, and it can be extended as this PR demonstrates (as far as I see it does not rely on any internals that an extension package could not access).

Disclaimer: I am a contributor to checkmate.

hadley:
@tdeenes obviously we share different opinions about suggested dependencies which primarily affect developers. You are welcome to try suggest PRs that remove if you think they are not useful, but they are mostly there to generate better looking vignettes which are valuable, in my opinion.

PR #311, opened 2023-07-11
Draft blog post

hadley:
Not intending to merge this, but this seemed like the easiest way to collaboratively edit.

PR #400, opened 2024-02-14
Draft R dev blog post

t-kalinowski:
Draft blog post for the R development blog, detailing the recent changes in base R which  enabled S7 to be published to CRAN.

(refocused iteration of #311)

### DIFF
--- a/vignettes/R-dev-blog-post.Rmd
+++ b/vignettes/R-dev-blog-post.Rmd
@@ -7,7 +7,7 @@ output: html_document
 
 R has built-in support for two functional Object Oriented Programming (OOP) systems: S3 and S4, corresponding to the third and fourth version of the S language, respectively. The two systems are, unfortunately, largely incompatible; this fragmentation in the R community and R sources is undesirable.
 
-[S7](https://rconsortium.github.io/S7/index.html) is a new OOP system being developed as a collaboration between representatives from R-Core, Bioconductor, tidyverse/Posit, ROpenSci, and the wider R community, with the goal of unifying the OOP systems and promoting interoperability. The long-term hope is to incorporate S7 into base R, following a period during which the concepts of S7 have been validated in a standalone R package.
+[S7](https://rconsortium.github.io/S7/index.html) is a new OOP system being developed as a collaboration between representatives from R-Core, Bioconductor, tidyverse/Posit, ROpenSci, and the wider R community, with the goal of unifying S3 and S4 and promoting interoperability. The long-term hope is to incorporate S7 into base R, following a period during which the concepts of S7 have been validated in a standalone R package.
 
 This blog post outlines four patches that have been integrated into base R to facilitate the existence of S7 as a standalone R package. Since S7 is built on top of S3, the new capabilities are available not only to S7 but also to all packages built on top of S3.
 
@@ -24,7 +24,7 @@ This blog post describes these new generics, the features in S7 that motivated t
 
 S3 only supports dispatching generics on their first argument, while S4 supports dispatching generics on an arbitrary number of arguments. S7 uses S3 dispatch, but also implements its own double dispatch (i.e. multiple dispatch on exactly two arguments). To enable both S3 support and multiple dispatch, dispatch in S7 is performed in two stages: S3 generics first dispatch on the first method to an S7 aware method, which then invokes S7-internal dispatch routines that support multiple dispatch. These are implementation details which, when everything works, are hidden from S7 users. The complication which motivated the addition of a new S3 generic is the Ops Group.
 
-Unlike other S3 generics, the Ops Group Generics, consisting of infix binary operators like `+`, `-`, `*`, and so on, support dispatching to S3 methods on either the first or second argument. Before R 4.3.0, if both objects had a method for the generic, R would signal an error. Starting with R 4.3.0, when both the first and second argument to an Ops generic implement a suitable method, R will invoke `chooseOpsMethod()` instead of signaling an error, giving an opportunity for the object types to declare that their Ops method implementation can handle the combination. If the `chooseOpsMethod()` method for an object returns `TRUE`, then the method found for that object is used by the Ops generic.
+Unlike other S3 generics, the Ops Group Generics, consisting of infix binary operators like `+`, `-`, `*`, and so on, support dispatching to S3 methods on either the first or second argument. Before R 4.3.0, if the two objects had different methods for the generic, R would signal a warning and use neither. Starting with R 4.3.0, if different methods are found for the first and second argument to the Ops generic, R will invoke `chooseOpsMethod()`, giving an opportunity for the object types to declare that their Ops method implementation can handle the combination. If the `chooseOpsMethod()` method for an object returns `TRUE`, then the method found for that object is used by the Ops generic.
 
 Because S7 implements its own generic dispatch mechanism, the `chooseOpsMethod()` method for S7 object always returns `TRUE`. This permits, for example, an S7 object type to define a `+` method that works with other S3 objects, like `Sys.Date()`.
 
@@ -58,11 +58,11 @@ np_array(1:3) + array(1:3)
 array(1:3) + tensorflow::as_tensor(1:3)
 
 # Adding a NumPy array with a TensorFlow Tensor.
-# Prior to R 4.3.0, this would signal an error, Warning about
+# Prior to R 4.3.0, this would signal an error and warn:
 #   Incompatible methods ("+.tensorflow.tensor", "+.python.builtin.object")
 # Beginning with R 4.3.0, chooseOpsMethod() is called to choose the appropriate
 # Ops method, and this now works
-np_array(1:3) + as_tensor(1:3)
+np_array(1:3) + tensorflow::as_tensor(1:3)
 ```
 
 ### `nameOfClass()`
@@ -91,7 +91,7 @@ inherits(x, tensorflow::tf$Variable)
 
 Prior to R 4.3.0, `@` could be used only with S4 objects. Beginning with R 4.3.0, this operator can now perform dispatch S3 dispatch as well. This bring `@` to parity with `@<-` (which has long been able to do S3 dispatch), and makes `@` a peer (in dispatch-ability) of `$`. This is a strictly backwards compatible change, as S3 dispatch is only performed *after* S4 routines have had an opportunity to dispatch.
 
-The ability to have formally defined properties is one of the benefits of S4 and S7; it enables collaborative work at a greater scale than the ad-hoc conventions of S3 classes, and having a dedicated operator for property access in S7 is desirable. Other than performing S3 dispatch, there are no other user-facing changes with `@` in base R.
+The ability to have formally defined properties is one of the benefits of S4 and S7; it enables collaborative work at a greater scale than the ad-hoc conventions of S3 classes, and having a dedicated operator for property access in S7 is desirable. Other than performing S3 dispatch, there are no other user-facing changes with `@` in base R---the default behavior of `@` in the absence of an S3 method is unchanged.
 
 ### `%*%`
 
--- a/vignettes/R-dev-blog-post.Rmd
+++ b/vignettes/R-dev-blog-post.Rmd
@@ -28,9 +28,9 @@ This blog post outlines four patches that have been integrated into base R to ma
 Each of these four patches introduces a new S3 generic:
 
 -   `chooseOpsMethod()`
+-   `%*%`
 -   `nameOfClass()`
 -   `@`
--   `%*%`
 
 This blog post describes these new generics, the features in S7 that motivated their implementation, and their utility to the broader R community outside of S7.
 These enhancements are generally beneficial for packages integrating or implementing alternative OOP systems.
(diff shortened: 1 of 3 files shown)


vignettes/R-dev-blog-post.qmd
```diff
@@ -0,0 +1,100 @@
...
+-   `chooseOpsMethod()`
+-   `nameOfClass()`
+-   `@`
+-   `%*%`
+
+This blog post describes these new generics, the features in S7 that motivated their implementation, and their utility to the broader R community outside of S7. These enhancements are generally beneficial for packages integrating or implementing alternative OOP systems. To illustrate these additional benefits, we use the reticulate package as an example. Reticulate bridges Python's OOP semantics with R's S3 system.
+
+### `chooseOpsMethod()`
+
+S3 only supports dispatching generics on their first argument, while S4 supports dispatching generics on an arbitrary number of arguments. S7 uses S3 dispatch, but also, implements multiple dispatch. To enable both S3 support and multiple dispatch, dispatch in S7 is performed in two stages: S3 generics first dispatch on the first method to an S7 aware method, which then invokes S7-internal dispatch routines that support multiple dispatch. These are implementation details which, when everything works, are hidden from S7 users. The complication which motivated the addition of a new S3 generic is the Ops Group.
```
hadley, line 26:
```suggestion
S3 only supports dispatching generics on their first argument, while S4 supports dispatching generics on an arbitrary number of arguments. S7 uses S3 dispatch, but also implements its own double dispatch (i.e. multiple dispatch on exactly two arguments). To enable both S3 support and multiple dispatch, dispatch in S7 is performed in two stages: S3 generics first dispatch on the first method to an S7 aware method, which then invokes S7-internal dispatch routines that support multiple dispatch. These are implementation details which, when everything works, are hidden from S7 users. The complication which motivated the addition of a new S3 generic is the Ops Group.
```

vignettes/R-dev-blog-post.qmd
```diff
@@ -0,0 +1,100 @@
...
+  function(e1, e2) "method: X + class_any"
+
+x <- ClassX()
+
+x + Sys.Date()
+```
+
+This also has utility for other packages that implement alternative OO systems. For example, reticulate is an R package that embeds Python in an R session, and allows R users to operate with Python objects directly. Python has its own implementation of method dispatch for infix operators like `+`.
+
+The addition of `chooseOpsMethod()` now allows both arguments in a call like `a + b` to implement a suitable `+` method. Specifically, it enables reticulate to implement a default fallback `+` method that dispatches to the corresponding Python routines, while also allowing R users to define a custom `+` method for specific Python object types. For example, the `+` operator can now select the appropriate method if methods for TensorFlow tensors and Python objects generally are both be implemented. Previously, the restriction of only one of the arguments implementing an S3 method required that the second argument in an Ops call not be an S3 class.
```
hadley, line 47:
```suggestion
This also has utility for other packages that implement alternative OO systems. For example, reticulate is an R package that embeds Python in an R session, and allows R users to operate with Python objects directly. Python has its own implementation of method dispatch for infix operators like `+`. The addition of `chooseOpsMethod()` enables reticulate to implement a default fallback `+` method that dispatches to the corresponding Python routines, while still allowing R users to define a custom `+` method for specific Python object types. 

For example, the `+` operator can now select the appropriate method if methods for TensorFlow tensors and Python objects generally are both be implemented. Previously, the restriction of only one of the arguments implementing an S3 method required that the second argument in an Ops call not be an S3 class.
```

vignettes/R-dev-blog-post.qmd
```diff
@@ -0,0 +1,100 @@
...
+# Beginning with R 4.3.0, chooseOpsMethod() is called to choose the appropriate
+# Ops method, and this now works
+np_array(1:3) + as_tensor(1:3)
+```
+
+### `nameOfClass()`
+
+Though S7 uses S3 dispatch, the class name used for S3 dispatch is considered an implementation detail (it is non-syntatic and mangled to minimize the potential for conflicts, and not intended to be user-facing). Additionally, it can be subject to change if the generic definition migrates between packages.
+
+For these reasons, it is desirable to be able to check for inheritance without directly using the S3 class name. To enable this, beginning with R 4.3.0 `base::inherits()` now accepts arbitrary objects for the second argument. If the second argument is an S3 object, the new `nameOfClass()` S3 generic is invoked to resolve the appropriate class name.
```
hadley, line 72:
```suggestion
For these reasons, it is desirable to be able to check for inheritance without directly using the S3 class name. To enable this, beginning with R 4.3.0 `base::inherits()` now accepts arbitrary objects for the second argument: if the second argument is an S3 object, the new `nameOfClass()` S3 generic is invoked to resolve the appropriate class name.
```

vignettes/R-dev-blog-post.qmd
```diff
@@ -0,0 +1,100 @@
...
+
+```{r, eval = FALSE}
+inherits(x, tensorflow::tf$Variable)
+```
+
+### `@`
+
+Prior to R 4.3.0, `@` could be used only with S4 objects. Beginning with R 4.3.0, this operator can now perform dispatch S3 dispatch as well. This bring `@` to parity with `@<-` (which has long been able to do S3 dispatch), and makes `@` a peer (in dispatch-ability) of `$`. This is a strictly backwards compatible change, as S3 dispatch is only performed *after* S4 routines have had an opportunity to dispatch.
```
hadley, line 92: Maybe note that there's still no default method for `@` but it's available for use in packages?

vignettes/R-dev-blog-post.qmd
```diff
@@ -0,0 +1,100 @@
...
+
+The ability to have formally defined properties is one of the benefits of S4; it enables collaborative work at a greater scale than the ad-hoc conventions of S3 classes, and having a dedicated operator for property access in S7 is desirable. Other than performing S3 dispatch, there are no other user-facing changes with `@` in base R.
```
hadley, line 94:
```suggestion
The ability to have formally defined properties is one of the benefits of S4 and S7; it enables collaborative work at a greater scale than the ad-hoc conventions of S3 classes, and having a dedicated operator for property access in S7 is desirable. Other than performing S3 dispatch, there are no other user-facing changes with `@` in base R.
```

ltierney:
A couple of quick notes:

- It would be better to say it is the _hope_ that S7 can be merged into base; _intention_ is too strong.
- I may be misremembering, but wasn't it only an error in `Ops` for both arguments to provide a method if those methods were not the same?
- The R blog uses `.Rmd`; I'm not sure it can handle `.qmd` at this point.

t-kalinowski:
Thanks @ltierney!

> I may be misremembering, but wasn't it only an error in Ops for both arguments to provide a method if those methods were not the same?

If two methods were found, R would use neither, issue a warning, and then proceed to try to use the built-in routines on the underlying SEXP. This is what I see in R 4.2.3:
```r
foo <- structure(new.env(), class = "foo")
bar <- structure(new.env(), class = "bar")

`+.foo` <- function(e1, e2) "foo"
`+.bar` <- function(e1, e2) "bar"

foo + bar
```
```
Error in foo + bar : non-numeric argument to binary operator
In addition: Warning message:
Incompatible methods ("+.foo", "+.bar") for "+" 
```

ltierney:
> Thanks @ltierney!
> 
> > I may be misremembering, but wasn't it only an error in Ops for both arguments to provide a method if those methods were not the same?
> 
> If two methods were found, R would use neither, issue a warning, and then proceed to try to use the built-in routines on the underlying SEXP. This is what I see in R 4.2.3:
> 
> ```r
> foo <- structure(new.env(), class = "foo")
> bar <- structure(new.env(), class = "bar")
> 
> `+.foo` <- function(e1, e2) "foo"
> `+.bar` <- function(e1, e2) "bar"
> 
> foo + bar
> ```
> 
> ```
> Error in foo + bar : non-numeric argument to binary operator
> In addition: Warning message:
> Incompatible methods ("+.foo", "+.bar") for "+" 
> ```

Right, but this would not signal an error:

```r
foo <- structure(new.env(), class = "foo")
bar <- structure(new.env(), class = "bar")
 
`+.foo` <- function(e1, e2) "foo"
`+.bar` <- function(e1, e2) "foo"

foo + bar
[1] "foo"
```

hadley:
@ltierney Interesting — my memory was that would only work if you did `` `+.bar` <- `+.foo` `` (i.e. they point to the same underlying function) but I guess it actually compares the function objects.

t-kalinowski:
Thanks @ltierney, I tweaked the `chooseOpsMethod()` section to be more precise.

mmaechler:
The `%*%` part seems unfinished.  We did deal quite a bit with it, but are a bit in a partly finished state ... but IIRC, that is only wrt `S4` which is not the topic here.
I did introduce  the new group `"matrixOps"`  to which `%*%` belongs in 4.3.*  and additionally  "crossprod" and "tcrossprod"  in R >= 4.4.0.   The fact that `%*%` is _"special operator"_  ([R language definition](https://cran.r-project.org/doc/manuals/r-release/R-lang.html#Special-operators ), also mentioned in other places in the manual) is not really crucial here.    I still argue that there are functions which should get dual argument / binary dispatch  (and note that there are other binary operators not in `Ops` nor in `matrixOps`,  e.g. `:`, or `::`  ...).

mmaechler:
Meta comments:  
- Using the very long lines in a `*.md` / `*.Rmd`  file gives "horribly" large diffs, when I see a "," missing or another silly typo I'd like to amend;  for _diff_ reason I'd break lines even though that's not so common in markdown
- I have a fork of `S7` but that does not *see* the `R-dev-blog-post` branch ... because I see now it is only a fork of the `RConsortium/S7:main` branch .  How to do reasonable pull requests here?

t-kalinowski:
Thank you @mmaechler! Regarding the meta comments:

- I pushed a change rewrapping the Rmd file so each sentence is on its own line. This should result in smaller diffs. I also added an entry in the front-matter to [`wrap: sentence`](https://rstudio.github.io/visual-markdown-editing/markdown.html#line-wrapping), which should be picked up by the [RStudio Visual Markdown Editor](https://posit.co/blog/exploring-rstudio-visual-markdown-editor/).
----

- I'm in favor of keeping the workflow simple; you can push directly to this branch:

```shell
git checkout -b R-dev-blog-post origin/R-dev-blog-post
# ... make edits to R-dev-blog-post.Rmd
git add vignettes/R-dev-blog-post.Rmd
git commit -m 'made it better'
git push
```

Alternatively, you can propose suggested changes in the github webui as part of a "Review"
1. In the PR page, navigate to the "Files changed" tab. https://github.com/RConsortium/S7/pull/400/files
2. Navigate to the line you want to edit, in the left sidebar (next to the line numbers) click the `+` icon that appears image
4. Click the "Add a suggestion" button, which populates the text input area with a markdown "suggestion" cell, where you can make edits. 
image
5. Click the "Start a Review" button to combine multiple such suggestions in a single batch. 
image

I (or others) can then click "Accept" in the webui to commit the suggestions to the branch.

Yet another alternative would be to make a new branch and open a PR from that branch to *this* branch, which would create a new pull request with an independent conversation thread.

lawremi:
Hi @t-kalinowski, I worked on this a bit yesterday, massaging the story and some of the wording in the technical sections. I'm about halfway through and will commit soon.

vignettes/R-dev-blog-post.Rmd
```diff
@@ -115,12 +119,12 @@ inherits(x, tensorflow::tf$Variable)
 
 ### `@`
 
-Prior to R 4.3.0, `@` could be used only with S4 objects.
+Prior to R 4.3.0, `@` could be used to access slots of S4 objects.
 Beginning with R 4.3.0, this operator can now perform dispatch S3 dispatch as well.
```
mmaechler, line 123: the first `dispatch `  should be deleted here (was there already)

t-kalinowski: Thanks, fixed now.

lawremi:
I started looking at actually posting this on the blog. One of the principles of the blog is that it should be rebuildable by anyone with relative ease. This post depends on several external packages, including the somewhat heavy tensorflow package. The examples are really powerful, so I would not want to drop them, but they add a lot of complexity to maintenance of the blog. I wonder if it would be better to create a simpler version for the dev blog and have it link to the full post hosted somewhere else? Or we could fake it by including the code output in the source file...

t-kalinowski:
I think pre-rendering is the most straight-forward solution that avoids needing to add `tensorflow` as a dependency to the blog engine. 

```r
rmd <- "R-dev-blog-post.Rmd"
knitr::knit(rmd) -> md
# publish(md)

# or, if the blog engine *really* prefers a Rmd extension
file.rename(md, rmd)
```

t-kalinowski:
@lawremi Would it be helpful for me to push a knitted version of the blog post?

lawremi:
@t-kalinowski yea I think it would be best if you generated it, so that you can check it and make sure everything looks as expected (and you already have the dependencies installed). Sorry it took so long for me to get back to this. Thanks for your help.

t-kalinowski:
Thanks @lawremi; done :heavy_check_mark:.

lawremi:
@t-kalinowski I don't see the .md or any rendered chunks in the Rmd. Perhaps you forgot to add the output?

t-kalinowski:
I renamed the md to an Rmd after rendering.(working on the assumption the blog engine wants an Rmd)

As the post is written, none of the chunks are meant to produce included output.

lawremi:
It looks to me that many of them should produce output. For example, the first chunk should demonstrate the intended dispatch by including the output of `x + Sys.date()`, right? While the reader can blindly trust that things behave as expected, I think including the output would improve the overall clarity of the examples. Is there a downside?

t-kalinowski:
Thanks @lawremi, good catch! It indeed makes more sense to have the chunks produce output. (I can't remember why I had set 'eval=FALSE' initially). I just pushed a knitted md file.

t-kalinowski:
The published post: https://blog.r-project.org/2024/05/17/generalizing-support-for-functional-oop-in-r/index.html
This was also on the front page of Hacker News: https://news.ycombinator.com/item?id=40501561

Thanks all!

Issue #404, opened 2024-06-10
Allow read-only properties to be set in the constructor


guslipkin:
Currently the documentation suggests that read-only properties (those with only a getter and no setter) are for computed values such as `now` or calculating using other properties of the object. However, sometimes you want to create an object with a value that doesn't change, isn't allowed to change, and isn't calculated from other values.

Take this person class, for example. We want to allow the birthdate to be set once and only once on construction, but the person's age will be calculated using the current date.

```r
person <- S7::new_class(
  name = 'person',
  properties = list(
    'name' = S7::class_character,
    'birthdate' = S7::new_property(
      class = S7::class_Date,
      getter = \(self) { self@birthdate }
    ),
    'age' = S7::new_property(
      class = S7::class_any,
      getter = \(self) { Sys.Date() - self@birthdate }
    )
  ),
  constructor = \(name, birthdate) {
    S7::new_object(
      S7::S7_object(),
      'name' = name,
      'birthdate' = birthdate
    )
  }
)

person('Unixtime', as.Date('1970-01-01'))
```

lawremi:
I think you could do this by having a setter that throws an error if the Date is already set (not `NULL`). There's also been the suggestion of a property with a dynamic default, see #376.

guslipkin:
That's a good idea, but I can't quite get it to work because there are some barriers in the way, including the issue I mentioned in #405 where you can't use default values and a constructor in tandem.

- If I create a setter that simply assigns `@birthdate` without any checks, it cannot be `NULL` because the `birthdate` property has a class `S7::class_Date`. I would need to create a union with `S7::class_any` because `NULL` does not have a class.
- If I do the above, but set `default = NULL` then call `person('Unixtime')`, I get an error that there is no default value because the `default` argument defaults to `NULL`, so that requires a switch to `NA`
- However, I've so far been unable to get any combination of default value and setter to work so you need to set the default in the constructor which is less safe in my opinion than using the `default` argument.

Our setup now looks like this:

```r
person <- S7::new_class(
  name = 'person',
  properties = list(
    'name' = S7::class_character,
    'birthdate' = S7::new_property(
      class = S7::class_Date,
      getter = \(self) { self@birthdate },
      setter = \(self, value) {
        if (is.na(self@birthdate)) self@birthdate <- value else warning('Cannot set read-only value')
        return(self)
      },
    ),
    'age' = S7::new_property(
      class = S7::class_any,
[...]

PR #406, opened 2024-06-11
Fix getter infinite recursion

### LINKED ISSUE #403: infinite recursion when trying to access an uninitialized property with getter function
When trying to access an uninitialized property that has a getter, a stack overflow occurs

```r

library(S7)

someclass <- new_class("someclass", properties = list(
  action = new_property(
    class_character,
    getter = function(self) self@action,
      self@action <- toupper(value)
      self 
    },
    default = "some action"
  )
))

x <- someclass()
x@action

#> Error: evaluation nested too deeply: infinite recursion / options(expressions=)?

```

as soon as I set a value the code works. Is this behaviour ok? 
default property value also does not help.

t-kalinowski:
Closes #403

### DIFF
--- a/R/constructor.R
+++ b/R/constructor.R
@@ -48,7 +48,7 @@ constructor_args <- function(parent, properties = list()) {
 
   self_args <- names2(properties)
   # Remove dynamic arguments
-  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter))]
+  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter) || !is.null(x$default))]
   if (is_class(parent) && !parent@abstract) {
     # Remove any parent properties; can't use parent_args() since the constructor
     # might automatically set some properties.
--- a/R/constructor.R
+++ b/R/constructor.R
@@ -48,7 +48,7 @@ constructor_args <- function(parent, properties = list()) {
 
   self_args <- names2(properties)
   # Remove dynamic arguments
-  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter) || !is.null(x$default))]
+  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter))]
   if (is_class(parent) && !parent@abstract) {
     # Remove any parent properties; can't use parent_args() since the constructor
     # might automatically set some properties.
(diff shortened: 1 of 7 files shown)


R/constructor.R
```diff
@@ -48,7 +48,7 @@ constructor_args <- function(parent, properties = list()) {
 
   self_args <- names2(properties)
   # Remove dynamic arguments
-  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter))]
+  self_args <- self_args[vlapply(properties, function(x) is.null(x$getter) || !is.null(x$default))]
```
lawremi, line 51: Note that a property with only a getter could be computed entirely dynamically, so we wouldn't want it in the constructor. I still think we should use the prototype approach, where the object just starts with its defaults, and there's no need to set a read-only property during construction. 

Btw, there is actually a bug here, where a property with a setter (and a getter) is dropped, in error. This is addressed by the prop_is_read_only branch.

lawremi:
Are you sure we can use the same list for both getting and setting? Assume a setter is called and it wants to get the value of the property, I think it would be surprising that the getter is not called.

As an example of where this would arise, assume a property is fully dynamic, like the value is being stored in a database, or on a C data structure. Maybe the setter wants to make sure that the value is different before performing the change.

I think we can still share a lot of the code between setting and getting, even if we use two different lists (the list symbol can just become an argument to the functions that manipulate the list).

t-kalinowski:
> Are you sure we can use the same list for both getting and setting?

My thought process was that in property `getter`/`setter` methods, we're in "internal" territory, where accessing the property fetches the underlying attribute. However, you make a compelling case that it would be useful to be able to call a `getter()` from a `setter()` method.

I'll change the approach to allow for that.

R/constructor.R
t-kalinowski: What should the interface be for someone looking to implement a class with a property that can be set when the instance is initially constructed but is read-only after that?

t-kalinowski: By the way, this looks good to me: https://github.com/RConsortium/S7/compare/main...prop_is_read_only

lawremi: See issue #404 for my proposed approach to construct-only properties.

PR #422, opened 2024-08-21
fix S3 method registration across packages

lawremi:


lawremi:
Would be nice to get some eyes on this. Registering S3 methods on a generic in another package (except base) is broken otherwise.

R/method-register.R
```diff
@@ -93,7 +93,7 @@ register_method <- function(generic,
   invisible(generic)
 }
 
-register_S3_method <- function(generic, signature, method) {
+register_S3_method <- function(generic, signature, method, envir = parent.frame()) {
```
hadley, line 96: Call this `env` for consistency with S7 argument names?

hadley (commented):
This seems plausible, but obviously I thought the code I wrote was plausible too 😀. Can you think of someway that we could test this?

R/method-register.R
lawremi: Happy to rename that

lawremi:
There are two ways things broke:
1. Upon building of a package that defined a method on a S3 generic
2. Loading the package and trying to invoke the method only to find that it was not registered at runtime

So yes, it should be possible to test those situations, although testing for those exact errors would be complicated. Easier would be to unit test that S3 method registration puts things in the right (provided) environment. I could add some tests for that.

hadley:
Those are likely to be fiddly to test, but it's worth the effort IMO.

t-kalinowski:
I took a look at adding tests here, but it appears that existing tests already cover this.

https://github.com/RConsortium/S7/blob/e205bdaa8b5e1da760da6f752d4efa63eb36405f/tests/testthat/test-external-generic.R#L75-L85

It's not completely clear at a glance, but these are the packages setup by the test:

### `t0`
```r
#' @export
bar <- S7::new_generic("bar", "x")
```

### `t1`
```r
#' @export
foo <- S7::new_generic("foo", "x")
```

### `t2`
```r
#' @importFrom t0 bar
S7::method(bar, S7::class_character) <- function(x) "bar"

foo <- S7::new_external_generic("t1", "foo", "x")
S7::method(foo, S7::class_character) <- function(x) "foo"

.onLoad <- function(libname, pkgname) {
  S7::methods_register()
}
```


How should the tests be expanded?

hadley:
@t-kalinowski I think the bug here is using `method()` with an S3 generic and an S7 class.

t-kalinowski:
@lawremi, I added tests and more fixes. It would be great if you can take a look. (Github won't let me "request a review" from you because you opened the original PR).

lawremi:
Thanks for all of your hard work, especially on the tests. This looks good to me.

PR #446, opened 2024-09-23
Silently ignore missing arguments in constructor

t-kalinowski:
As discussed in https://github.com/RConsortium/S7/pull/445, it might make sense to allow
missing arguments to a constructor to "skip" setting the property and avoid invoking a custom setter.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -144,6 +144,9 @@ modify_list <- function (x, new_vals) {
   x
 }
 
+list2 <- function(...)
+  .Call(collect_dots_skip_missing_, environment(), substitute(list(...)))
+
 
 # For older versions of R ----------------------------------------------------
 deparse1 <- function(expr, collapse = " ", width.cutoff = 500L, ...) {
--- a/tests/testthat/test-constructor.R
+++ b/tests/testthat/test-constructor.R
@@ -140,8 +140,8 @@ test_that("can create constructors with missing or lazy defaults", {
     birthdate = Sys.Date()
   ))) # no age
 
-  expect_error(Person(), 'argument "first_name" is missing, with no default')
-  expect_error(Person("Alice"), 'argument "last_name" is missing, with no default')
+  expect_error(Person(), "@first_name")
+  expect_error(Person("Alice"), "@last_name")
 
   p <- Person("Alice", ,"Smith")
 
--- a/R/class.R
+++ b/R/class.R
@@ -252,7 +252,7 @@ new_object <- function(.parent, ...) {
     stop(msg)
   }
 
-  args <- list(...)
+  args <- list2(...)
   nms <- names(args)
 
   # TODO: Some type checking on `.parent`?
--- a/src/init.c
+++ b/src/init.c
@@ -10,6 +10,7 @@ extern SEXP S7_class_(SEXP, SEXP);
 extern SEXP S7_object_(void);
 extern SEXP prop_(SEXP, SEXP);
 extern SEXP prop_set_(SEXP, SEXP, SEXP, SEXP);
+extern SEXP collect_dots_skip_missing_(SEXP, SEXP);
 
 static const R_CallMethodDef CallEntries[] = {
     {"method_", (DL_FUNC) &method_, 4},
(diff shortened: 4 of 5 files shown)


tests/testthat/test-constructor.R
```diff
@@ -140,8 +140,8 @@ test_that("can create constructors with missing or lazy defaults", {
     birthdate = Sys.Date()
   ))) # no age
 
-  expect_error(Person(), 'argument "first_name" is missing, with no default')
-  expect_error(Person("Alice"), 'argument "last_name" is missing, with no default')
+  expect_error(Person(), "@first_name")
+  expect_error(Person("Alice"), "@last_name")
```
hadley, line 144: Maybe switch these to snapshots to make it easier for reviewers to see what the new errors are?

tests/testthat/_snaps/constructor.md
```diff
@@ -69,3 +69,23 @@
...
+      - @nick_name must be <character>, not <NULL>
+
+---
+
+    Code
+      Person("Alice")
+    Condition
+      Error:
+      ! <Person> object properties are invalid:
+      - @last_name must be <MISSING> or <character>, not <NULL>
```
hadley, line 90: This seems a bit confusing because it is missing?

R/utils.R
```diff
@@ -144,6 +144,9 @@ modify_list <- function (x, new_vals) {
   x
 }
 
+list2 <- function(...)
```
hadley, line 147: I'd love to see some tests explicitly for this function.

tests/testthat/_snaps/constructor.md
t-kalinowski: The issue happens because `person@last_name <- <value>` is never set since `<value>` is missing. Then, when the validator fetches `person@last_name`, it returns `NULL` (because `attr(x, "does_not_exist")` return `NULL`), which is the wrong type and causes the error.

This whole test probably needs to be rewritten to make sense with the new behavior.

t-kalinowski: https://github.com/RConsortium/S7/blob/15a01a3ab76848981ca03ab30e30e3609c554a01/src/prop.c#L260-L261

t-kalinowski:
If/when promise accessors become part of the C "API", the code in this PR should be updated to use them (https://github.com/posit-dev/ark/pull/538#discussion_r1772858928)

tests/testthat/_snaps/constructor.md
hadley: If it's not set, should we be using `quote(expr = )`?

t-kalinowski: To do that, we would have to use `Rf_setAttrib(object, Rf_install("name"), R_missingArg)` in `prop<-` when the object is first constructed.

`quote(expr=)` is such a pain to work with at the R level; I'm beginning to think that perhaps we should not create a pattern that requires all object instances to be handled with care, checking with `missing()` like this.

To make the "deprecate via getter+setter" pattern possible, maybe we should return to what we discussed in https://github.com/RConsortium/S7/pull/396#discussion_r1486563418 and altogether bypass or not invoke custom setters on initial construction.

We would then need to provide a convenient way to "opt-in" to running the setters without requiring a full custom constructor. 

Perhaps we add an argument to `new_class(..., initialize = function(self) {})`, a function authors can provide. The `initialize` function will be called with the output of the `constructor` after the `constructor()` has returned but before `validator()` is called. (somewhat analogous to `__new__` and `__init__` in Python.)

t-kalinowski: We may also want to add an `initializer` arg to `new_property`, defaulting to `setter`.
```r
new_property(..., initializer = setter)
```

lawremi: Could we just use `class_missing` as the sentinel value? Or perhaps something similar but named like `arg_missing`?

hadley: I think we got a bit side-tracked into a much bigger topic than I was thinking. Looking at this with fresh eyes, I think all that I want is for the error message to be:

```
! <Person> object properties are invalid:
- @last_name must be <character>, not <NULL>
```

PR #448, opened 2024-09-24
Accept `initializer` in `new_class()` and `new_property()`

t-kalinowski:
This PR is a sketch of one possible solution for allowing users some more control over what property `setters` get called at object construction without too much hassle (i.e., without requiring a custom constructor.)

https://github.com/RConsortium/S7/pull/445#issuecomment-2362179831

Now that I look at this, it doesn't seem quite right either. 

Perhaps a `new_property(..., virtual = TRUE)` argument, which would mean the property is not set at construction and is not included as a constructor argument.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -33,6 +33,9 @@
 #'   A custom constructor should call `new_object()` to create the S7 object.
 #'   The first argument, `.data`, should be an instance of the parent class
 #'   (if used). The subsequent arguments are used to set the properties.
+#' @param initializer An optional initializer function. If provided, this
+#'   function is called after `constructor()` but before `validator()`. It
+#'   should take a single argument, `self`.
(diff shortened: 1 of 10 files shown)


lawremi:
I am not sure how removing something from the constructor solves the problem of the deprecated property, because it still needs to be settable for backwards compatibility.

t-kalinowski:
Given the example use case of a deprecated property, the goal is to not warn if the user didn't supply an explicit value in the constructor but warn if they did and also warn in all other cases of the property being set. Decoupling setting at initialization from setting all other times might be one approach towards making that possible. 

However, for this example, I think there is a more minimal approach using `...`, in #445. (We may want to revisit explicit initializers in the future, but that would be a separate issue.)

PR #457, opened 2024-10-10
Enforce validator return value: must be `NULL` or a string.

### LINKED ISSUE #454: bug(?): validators must return NULL
I have observed that validators _must_ return `NULL` otherwise the validators error out. I cannot reason as to why this is. Is this intended behavior? Should the validators always return `self`?  

``` r
params <- S7::new_class(
  "form_params",
  properties = list(params = S7::class_list),
  validator = function(self) {
    lapply(
      self@params,
      rlang:::check_string,
      allow_null = TRUE,
      allow_na = FALSE,
      call = rlang::caller_call()
    )
    # idk why this has to return NULL
    return(NULL)
  }
)

params(list(x = "a"))
#> <form_params>
#>  @ params:List of 1
#>  .. $ x: chr "a"

params <- S7::new_class(
  "form_params",
  properties = list(params = S7::class_list),
  validator = function(self) {
    lapply(
      self@params,
      rlang:::check_string,
      allow_null = TRUE,
      allow_na = FALSE,
      call = rlang::caller_call()
    )
  }
)

params(list(x = "a"))
#> Error: <form_params> object is invalid:
#> - NULL
```

t-kalinowski:
Closes #454


R/aaa.R
```diff
@@ -8,8 +8,20 @@ new_function <- function(args = NULL,
...
 
+
+`append<-` <- function(x, after, value) {
+  if (missing(after))
+    c(x, value)
+  else
+    append(x, value, after = after)
+}
+
 `append1<-` <- function (x, value) {
```
hadley, line 19: Should this be `append_list<-`?

R/property.R
```diff
@@ -293,21 +293,34 @@ prop_error_unknown <- function(object, prop_name) {
...
     } else {
-      NULL
+      return(NULL)
     }
-  } else {
-    NULL
   }
+
+  stop(sprintf(
+    "%s validator is expected to return NULL or a character, not <%s>",
```
hadley, line 321:
```suggestion
    "%s validator must return NULL or a character, not <%s>",
```

R/property.R
```diff
@@ -293,21 +293,34 @@ prop_error_unknown <- function(object, prop_name) {
...
+    prop_label(object, prop$name), typeof(val)
```
hadley, line 322: Needs a trailing full stop.

(Since we're basically following https://style.tidyverse.org/errors.html)

R/valid.R
```diff
@@ -84,11 +84,21 @@ validate <- function(object, recursive = TRUE, properties = TRUE) {
 
   # Next, recursively validate the object
   errors <- character()
-  repeat({
-    errors <- c(errors, class_validate(class, object))
+  repeat {
+    error <- class_validate(class, object)
```
hadley, line 88: Is there enough duplication here it's worth pulling out into a function?

hadley (approved):
Worth a news bullet?

R/aaa.R
t-kalinowski: I don't think so, because it works with atomic vectors too. The constraint is that it only ever appends 1 element, and a guardrail is there to avoid any implicit type casting.

```r
> x <- character()
> append1(x) <- "foo"
> append1(x) <- c("foo", "bar")
Error in `x[[length(x) + 1L]] <- value`:
! more elements supplied than there are to replace
Run `rlang::last_trace()` to see where the error occurred.
```

hadley: Ooooh, that's what the `1` means.

Maybe it would be worth rewriting `append<-` to use `[<-` so it's more obvious what the distinction is?

t-kalinowski: `c()` does a lot more work than `[<-`, and it'd be tricky to match the current feature set with just `[<-`. (FWIW, this pair of `append<-` and `append1<-` is pretty seasoned).

R/valid.R
t-kalinowski: The usage patterns between prop validators and class validators are subtly different, but different enough that I believe a "generalized" function wouldn't be worth the added complexity.

PR #458, opened 2024-10-10
add `nameOfClass.S7_base_class` method

### LINKED ISSUE #432: add nameOfClass.S7_base_class()?
Would it be worth adding a `nameOfClass()` method for base classes? The user might expect `inherits("foo", class_character)` to work.

t-kalinowski:
Closes #432


tests/testthat/test-base.R
```diff
@@ -338,3 +338,16 @@ test_that("Base S3 classes can be properties", {
...
+
+
+test_that("inherits() works with S7_base_class", {
+  # nameOfClass() introduced in R 4.3
+  skip_if(getRversion() < "4.3")
+
+  # test nameOfClass.S7_base_class
+  expect_true(inherits("foo", class_character))
+
+  Foo := new_class(class_character)
```
hadley, line 350: `:=` ??? 🤯

t-kalinowski: Credit goes to @lawremi for the idea!

hadley (approved):

tests/testthat/test-base.R
lawremi: When can we graduate it from a test helper?  :wink: It's super convenient and clarifies code.

t-kalinowski: I'm on board for exporting it. (From the tidyverse, we'd have to figure out how to resolve the clash with  `` rlang`:=` ``)

lawremi: Yea, I think as discussed, the rlang one expects different types of arguments (symbols vs. calls) so disambiguating the intent should be possible.

PR #459, opened 2024-10-10
Automatically detect package name in `new_class(package=)`

### LINKED ISSUE #411: Default value for new_class(package=)
Would it not be reasonable to have the `package=` argument to `new_class()` default to  `if (isNamespace(topenv(parent.frame()))) getNamespaceName(topenv(parent.frame()))`? Obviously we would want  to wrap that up in a function to make it easier to read. Passing the package name explicitly seems unnecessary when code is in a package.

t-kalinowski:
Closes #411


R/aaa.R
```diff
@@ -13,3 +13,11 @@ new_function <- function(args = NULL,
   x[[length(x) + 1L]] <- value
   x
 }
+
+topNamespaceName <- function(env = parent.frame()) {
+  env <- topenv(env)
+  if (isNamespace(env))
+    getNamespaceName(env)
+  else
+    NULL
```
hadley, line 22:
```suggestion
  if (isNamespace(env)) {
    getNamespaceName(env)
  } else {
    NULL
  }
```

tests/testthat/helper.R
```diff
@@ -140,9 +145,23 @@ dbg <- function(..., .display = utils::str) {
...
     )))
   }
 
   invisible(out)
 }
+
+`%error%` <- function(x, y) tryCatch(x, error = function(e) y)
+
+# prevent new_class() from creating `S7::` prefixed S3 class names in tests.
+local({
```
hadley, line 158: I wonder if it's better to be explicit this in `topNamespaceName()` by inlining then using `testthat::testing_package`?

hadley (approved):

tests/testthat/helper.R
t-kalinowski: I don't understand, can you please explain more?

R/aaa.R
mmaechler: well,  `else NULL`  already being a no-op,  I do not really prefer the extra `{ .. }`.
My personal style uses a 1-liner for something so simple:
```
  if (isNamespace(env)) getNamespaceName(env)  # else NULL
```
.. applying  _KISS_

tests/testthat/helper.R
hadley: Should the implementation of `topNamespaceName()` be this?

```R
topNamespaceName <- function(env = parent.frame()) {
  if (testthat::testing_package() == "S7") {
    return()
  }

  env <- topenv(env)
  if (isNamespace(env))
    getNamespaceName(env)
}
```

(but with `testthat::testing_package()` inlined into S7)

tests/testthat/helper.R
```diff
@@ -127,9 +132,29 @@ dbg <- function(..., .display = utils::str) {
...
 
   invisible(out)
 }
+
+`%error%` <- function(x, y) tryCatch(x, error = function(e) y)
+
+# prevent new_class() from creating `S7::` prefixed S3 class names in tests.
+{
+  body(topNamespaceName) <- bquote({
+    if (testthat::testing_package() == "S7") {
```
hadley, line 147: I meant just do it in the typical definition of `topNamespaceName` so we don't need to do any messing around in here. (Which is much harder to discover if something goes wrong)

t-kalinowski: Ideally, I think we want "test drive" the function as it might exist in base, i.e., without accommodations for the S7 unit-tests edge-case.

hadley: Hmmmm, I just generally worry that this sort of magic eventually will cause a lot of head scratching when investigating a test failure (which is the time when you least want any surprises/magic). How much work would it be to explicitly update the tests?

t-kalinowski: `new_class(` appears 234 times in the tests, each call would have to be updated. Happy to do it if that's the best way. 
```r
$ rg -F 'new_class(' tests | wc -l
234
```

Could also do something like this in `tests/helpers.R`:
```r
formals(new_class)['package'] <- list(NULL)
```

hadley: Yeah, but it only causes ~20 non-snapshot test failures, so I think you could just update those?

R/aaa.R
t-kalinowski: I just remembered/rediscovered why I had originally written `else NULL`—it was to make a `NULL` return value print; otherwise, the implicit `else` is returned `invisible()`. 

One could wrap with `(if(...))`, but I had opted for the more explicit `else NULL`.

PR #460, opened 2024-10-10
Include value in error message about invalid value from `register_method()`

### LINKED ISSUE #447: `register_method()` error is missing context
```
Error:
! `generic` is a function, but not an S3 generic function
Hide Traceback
     ▆
  1. ├─devtools::load_all()
  2. │ └─pkgload::load_all(...)
  3. │   └─pkgload:::run_pkg_hook(package, "load")
  4. │     └─nsenv[[f_name]](dirname(ns_path), package)
  5. │       └─rlang::run_on_load() at [elmer/R/zzz.R:2:3](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
  6. │         └─rlang (local) callback()
  7. └─S7::`method<-`(`*tmp*`, openai_model, value = `<cr_gnrtr>`)
  8.   └─S7:::register_method(generic, signature, value, env = parent.frame())
  9.     └─S7:::as_generic(generic)
 10.       └─S7:::as_S3_generic(x)
 ```

Would be nice to know the name of the generic!

### PR DESCRIPTION
**Pull request by t-kalinowski**
Closes #447

### DIFF
--- a/R/generic-spec.R
+++ b/R/generic-spec.R
@@ -28,8 +28,8 @@ as_S3_generic <- function(x) {
     }
   }
 
-  stop("`generic` is a function, but not an S3 generic function: ",
-       deparse1(x, width.cutoff = 100L), call. = FALSE)
+  stop("`generic` is a function, but not an S3 generic function: \n",
+       deparse_trunc(x, 100), call. = FALSE)
 }
 
 S3_generic <- function(generic, name) {
--- a/R/generic-spec.R
+++ b/R/generic-spec.R
@@ -28,7 +28,8 @@ as_S3_generic <- function(x) {
     }
   }
 
-  stop("`generic` is a function, but not an S3 generic function", call. = FALSE)
+  stop("`generic` is a function, but not an S3 generic function: ",
+       deparse1(x, width.cutoff = 100L), call. = FALSE)
 }
 
 S3_generic <- function(generic, name) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -144,6 +144,13 @@ modify_list <- function (x, new_vals) {
   x
 }
 
+deparse_trunc <- function(x, width, collapse = "\n") {
+  x <- deparse1(x, collapse)
+  if (nchar(x)> width)
+    x <- sprintf("%s....", substr(x, 0, width-4))
+  x
+}
+
 
 # For older versions of R ----------------------------------------------------
 deparse1 <- function(expr, collapse = " ", width.cutoff = 500L, ...) {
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by hadley** [comment]
Hmmmm, I wonder if we should add a error wrapper in `method<-` since it has a bit more context. I think ideally I'd want

```
foo <- function() {}
method(foo, class_character) <- function(...) {}
#> Error: Failed to register <character> method for <foo> generic
#> `generic` is a function, but not an S3 generic function
```

But looking at the traceback, maybe we can't get the name of the generic because of the `*tmp*` substitution?

**Comment by hadley** [comment]
Yeah, so the best we could do is mention the method. Maybe something like this?

```R
foo <- function() {}
method(foo, class_character) <- function(...) {}
#> Error: Failed to register <character> method for generic
#> `generic`, function(...) is a function, but not a generic
```

R/generic-spec.R
```diff
@@ -28,7 +28,8 @@ as_S3_generic <- function(x) {
     }
   }
 
-  stop("`generic` is a function, but not an S3 generic function", call. = FALSE)
+  stop("`generic` is a function, but not an S3 generic function: ",
+       deparse1(x, width.cutoff = 100L), call. = FALSE)
```
hadley, line 32: I think displaying the arguments, without the body, would be sufficient. (Like `args(mean)` but without the `NULL`).

t-kalinowski: `function (x, ...)` doesn't seem like a very useful breadcrumb, especially if we can easily include the head of the first expression in the function body.

hadley: I guess I'm worried about this spitting out a very large function body, making it hard to see the actual error message.

t-kalinowski: > `width.cutoff = 100L` 

maybe we truncate this more/better?

hadley: I think that just determines the line length?

t-kalinowski: TIL! (I had thought `width.cutoff` capped `nchar()` of the returned string)

Maybe something like:

```
deparse_trunc <- function(x, width, collapse = "\n") {
  x <- deparse1(x, collapse)
  if (nchar(x)> width) 
     x <- sprintf("%s....", substr(x, 0, width-4))
  x
}
```
which gives:
```r
> cat(deparse_trunc(plot.default, 80))
function (x, y = NULL, type = "p", xlim = NULL, ylim = NULL, log = "", main ....
```

hadley: Yeah, that sounds good to me.

PR #461, opened 2024-10-10
Add class lookup table

t-kalinowski:
This PR enables usage of the base constructors like `base::character` in `new_class()`, `new_property()`, and similar.

For example:

``` r
library(S7)
Foo <- new_class("foo", parent = character)

Foo
#> <foo> class
#> @ parent     : <character>
#> @ constructor: function(.data) {...}
#> @ validator  : <NULL>
#> @ properties :
Foo()
#> <foo> chr(0)
```


lawremi:
What is the rationale? We had considered this early on, since classes are constructors, but we went with reified class objects instead. I'm not sure about having both.

Issue #401, opened 2024-03-15
S7 arrays


sjcowtan:
Is there a good way to create an S7 object which inherits from "array" (or "matrix", but my specific use case is "array")? If not, will this be added?

TimTaylor:
FWIW - Looks like it's currently in draft stage (https://github.com/RConsortium/S7/pull/434) so hopefully going to be added.

sjcowtan:
yay!

t-kalinowski:
This is implemented in the dev version of S7.

```r
new_class("my_array", parent = class_array)
```

hadley:
@t-kalinowski how does that work? what type of array do you get?

t-kalinowski:
length 0 logical (this makes the most sense, for the same reason that `NA` is logical; it can easily be promoted to another atomic)
``` r
library(S7)

new_class("my_array", parent = class_array)
#> <my_array> class
#> @ parent     : S3<array>
#> @ constructor: function(.data, dim, dimnames) {...}
#> @ validator  : <NULL>
#> @ properties :
new_class("my_array", parent = class_array)()
#> <my_array>  logi[0 (1d)]
```

<sup>Created on 2024-10-11 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hadley:
@t-kalinowski hmmm, that feels weird to me because `dim` is orthogonal to class in S3. So I would have expected something more like `class_array_logical`, `class_array_numeric` etc.

t-kalinowski:
I agree, this should be easy. 

We only provide `class_*` for actual base classes, that seems like a compound class. 

I wonder if we should introduce a `&` / `new_intersection()` for classes, complimenting `|` /`new_union()`. Then, `class_array_logical` could be achieved with:

```r
class_array & class_logical
```

hadley:
It feels like we should discuss this design a bit more before committing to it in a release. Maybe roll back for this release?

mmaechler:
> I agree, this should be easy.
> 
> We only provide `class_*` for actual base classes, that seems like a compound class.
> 
> I wonder if we should introduce a `&` / `new_intersection()` for classes, complimenting `|` /`new_union()`. Then, `class_array_logical` could be achieved with:
> 
> ```r
> class_array & class_logical
> ```

This sounds nice to me ((without further deep thinking though)  independent of `class_array` etc.

lawremi:
[...]

PR #463, opened 2024-10-15
revert `class_array` and `class_matrix`

t-kalinowski:
This PR removes `class_array` and `class_matrix`, pending further design and discussion. 

https://github.com/RConsortium/S7/issues/401#issuecomment-2407417482

`array` and `matrix` are unique because these are implicit S3 classes based on the presence of a `dim` attribute. 
 
Quoting from `?UseMethod`:

> If the object does not have a class attribute, it has an _implicit_ class. Matrices and arrays have class `"matrix"` or `"array"` followed by the class of the underlying vector. Most vectors have class the result of `mode(x)`, except that integer vectors have class `c("integer", "numeric")` and real vectors have class `c("double", "numeric")`.

E.g.:
```r
> x <- 1
> attr(x, "dim") <- c(1, 1)
> .class2(x)
[1] "matrix"  "array"   "double"  "numeric"
```


lawremi:
I feel like this might have jumped the gun a little. An `S7_S3_class` for matrix and array is still useful, because there are objects with that class. Sure, the hierarchy varies, but we can still use these for dispatch and properties.

To make something responsive to `is.matrix()` is orthogonal to these classes, in my opinion.

hadley:
@lawremi I think we just need to discuss this a bit more before committing to an API. And we want to do a minor release soon, so pulling these out just gives us a bit more time to think about them.

PR #470, opened 2024-10-21
Fix custom `getter`s with `class_call` objects

### LINKED ISSUE #469: Custom `getter` / `setter` evaluates objects of `class_call`
``` r
library(S7)
Foo <- new_class("Foo", class_call)
foo <- Foo(quote(stop("boom")))
foo
#> <Foo>  language stop("boom")

Foo <- new_class("Foo", class_call, properties = list(
  bar = new_property(getter = function(self) {
    self # evaluates `stop('boom')` call
    "bar"
  })
))

Foo(quote(stop("boom")))@bar
#> Error in (function (self) : boom

traceback()
#> No traceback available
```

Related: The error message + traceback from a custom getter/setter is not very helpful.

t-kalinowski:
Fixes #469

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -24,9 +24,8 @@ new_function <- function(args = NULL,
 
 
 topNamespaceName <- function(env = parent.frame()) {
-  env <- topenv(env)
-  if (isNamespace(env))
-    getNamespaceName(env)
+  (if (isNamespace(env <- topenv(env)))
+    getNamespaceName(env))
 }
 
 is_string <- function(x) {
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -24,8 +24,12 @@ new_function <- function(args = NULL,
 
 
 topNamespaceName <- function(env = parent.frame()) {
-  (if (isNamespace(env <- topenv(env)))
-    getNamespaceName(env))
+  env <- topenv(env)
+  if (!isNamespace(env)) {
+    return()
+  }
+
+  getNamespaceName(env)
 }
 
 is_string <- function(x) {
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -84,7 +84,7 @@ is_S7_type <- function(x) {
 check_subsettable <- function(x, allow_env = FALSE) {
   allowed_types <- c("list", "language", "pairlist", if (allow_env) "environment")
   if (!typeof(x) %in% allowed_types) {
-    stop("S7 objects are not subsettable.", call. = TRUE)
+    stop("S7 objects are not subsettable.")
   }
   invisible(TRUE)
 }
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -82,8 +82,7 @@ is_S7_type <- function(x) {
 }
 
 check_subsettable <- function(x, allow_env = FALSE) {
-  allowed_types <- c("list", if (allow_env) "environment",
-                     "language", "pairlist")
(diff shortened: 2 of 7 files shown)


t-kalinowski:
Is there a better way to protect a call object from evaluation by `Rf_eval()` than wrapping the object in `base::quote()`?

R/zzz.R
```diff
@@ -82,9 +82,10 @@ is_S7_type <- function(x) {
 }
 
 check_subsettable <- function(x, allow_env = FALSE) {
-  allowed_types <- c("list", if (allow_env) "environment")
+  allowed_types <- c("list", if (allow_env) "environment",
+                     "language", "pairlist")
   if (!typeof(x) %in% allowed_types) {
-    stop("S7 objects are not subsettable.", call. = FALSE)
+    stop("S7 objects are not subsettable.", call. = TRUE)
```
hadley, line 88: We don't usually show the call when it's a non-exported function.

R/zzz.R
```diff
@@ -82,9 +82,10 @@ is_S7_type <- function(x) {
 }
 
 check_subsettable <- function(x, allow_env = FALSE) {
-  allowed_types <- c("list", if (allow_env) "environment")
+  allowed_types <- c("list", if (allow_env) "environment",
+                     "language", "pairlist")
```
hadley, line 86:
```suggestion
  allowed_types <- c("list", "language", "pairlist", if (allow_env) "environment")
```
?

R/aaa.R
```diff
@@ -24,9 +24,8 @@ new_function <- function(args = NULL,
 
 
 topNamespaceName <- function(env = parent.frame()) {
-  env <- topenv(env)
-  if (isNamespace(env))
-    getNamespaceName(env)
+  (if (isNamespace(env <- topenv(env)))
+    getNamespaceName(env))
```
hadley, line 28: Personally, I'd write this more like:

```R
env <- topenv()
if (!isNamespace(env) {
  return()
}

getNamespaceName(env)
```

I find that logic easier to follow

hadley (approved):

R/zzz.R
t-kalinowski: See the `QuotedCall` class in tests for how `[[` on a `class_call` (`"language"`) is useful.

R/zzz.R
```diff
@@ -82,9 +82,9 @@ is_S7_type <- function(x) {
 }
 
 check_subsettable <- function(x, allow_env = FALSE) {
-  allowed_types <- c("list", if (allow_env) "environment")
+  allowed_types <- c("list", "language", "pairlist", if (allow_env) "environment")
   if (!typeof(x) %in% allowed_types) {
-    stop("S7 objects are not subsettable.", call. = FALSE)
+    stop("S7 objects are not subsettable.", call. = TRUE)
```
hadley, line 87:
```suggestion
    stop("S7 objects are not subsettable.")
```

😄

t-kalinowski: We really need a better error call and traceback from custom getter/setter calls, because right now, debugging any errors signaled from them is unnecessarily difficult. But that should be a separate issue.

R/zzz.R
t-kalinowski: We really need a better error call and traceback from custom getter/setter calls, because right now, debugging any errors signaled from them is unnecessarily difficult. But that should be a separate issue.

PR #473, opened 2024-10-21
Let child classes change default values of parent properties

### LINKED ISSUE #467: Default value introduced in subclass does not propagate to constructor
This [line](https://github.com/RConsortium/S7/blob/b5cf389d7c6dc676c2d938e2feac6d7ec04d0c50/R/constructor.R#L64) drops the parent properties from the constructor by name; however, I think it's valid for subclasses to add/change the default for a parent property. Since we're dropping the subclass properties, the changed default does not make it to the constructor. Self args should probably be kept (and override the parent) if they have a different default.

t-kalinowski:
closes #467

### DIFF
--- a/R/constructor.R
+++ b/R/constructor.R
@@ -61,7 +61,12 @@ constructor_args <- function(parent, properties = list()) {
   if (is_class(parent) && !parent@abstract) {
     # Remove any parent properties; can't use parent_args() since the constructor
     # might automatically set some properties.
-    self_arg_nms <- setdiff(self_arg_nms, names2(parent@properties))
+    parent_prop_nms <- names2(parent@properties)
+    overridden <- intersect(self_arg_nms, parent_prop_nms)
+    for(name in overridden)
+      parent_args[[name]] <- prop_default(properties[[name]])
+
+    self_arg_nms <- setdiff(self_arg_nms, parent_prop_nms)
   }
 
   self_args <- as.pairlist(lapply(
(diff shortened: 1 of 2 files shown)


t-kalinowski:
This is tricky when a parent class has a custom constructor, where the default value is essentially unknown, and implicitly different from `prop$default`.

I.e., this test: https://github.com/RConsortium/S7/blob/b5cf389d7c6dc676c2d938e2feac6d7ec04d0c50/tests/testthat/test-constructor.R#L80

@lawremi, let me know if you have any thoughts about how to resolve this.

lawremi:
Is my understanding correct that initialization of properties with default values only happens through the constructor? If so, then it seems tricky to implement a custom constructor, as it will need to support overriding of defaults. 

One solution is the prototype idea, where the object starts with its default values. The constructor can always set them a second time, but just making sure they are set initially really simplifies the constructor contract. 

Or, we just tell the user that custom constructors is an expert-level feature. Personally, I would avoid the type of constructor in the cited test and just leave the conveniences to a separate, user-facing constructor. Leaving the actual constructor to low-level property setting is a much more cohesive approach. In the core of Bioc, we rarely if ever overrode `initialize()`, which is the equivalent of a custom constructor, for these reasons.

lawremi:
Would be good to revisit this, as it keeps biting me. The test that fails should not be relevant, since the subclass is not actually overriding any properties. Obviously, overriding a property (like inheritance itself) is strong coupling, so if the parent starts doing weird things in the constructor, things are expected to break.

hadley:
Fixed in the meantime.

PR #481, opened 2024-10-25
Don't inline classes in constructor if possible

### LINKED ISSUE #477: inlining obfucscates constructor defaults
It seems that we now inline S7 constructors/classes into the constructor formals when a class is composed of other S7 objects. This makes the defaults difficult to read, such as in the Rd usage. Class objects are also quite heavy (perhaps a problem to address in itself), so serializing "empty" objects results in multi-megabyte files and consumes a corresponding amount of time.

Perhaps we could rely on a heuristic to resolve class objects to their symbols in their originating namespace.

t-kalinowski:
This PR changes a constructor default formal value from an inlined class call, to a call like `pkgname::classname()`, if `class@package` is not `NULL`. 

closes #477

With this PR:
``` r
library(S7)
Coordinate <- new_class(
  "Coordinate",
  properties = list(
    x = class_double,
    y = class_double
  ),
  package = "draw"
)

Shape <- new_class(
  "Shape",
  properties = list(
    position = Coordinate, # Coordinate would normally be inlined 
    name = class_character
  ),
  package = "draw"
)

## Shape and Coordinate are from the same package, 
## so Coordinate() call is not namespaced with :: 
str(formals(Shape))
#> Dotted pair list of 2
#>  $ position: language Coordinate()
#>  $ name    : chr(0)

stopifnot(identical(
  formals(Shape)$position,
  quote(Coordinate())
))

## different package name, now Coordinate() is inlined as a namespaced :: call
Shape2 <- new_class(
  "Shape",
  properties = list(
    position = Coordinate, # Coordinate would normally be inlined 
    name = class_character
  ),
  package = "draw2"
)

str(formals(Shape2))
#> Dotted pair list of 2
#>  $ position: language draw::Coordinate()
#>  $ name    : chr(0)

stopifnot(identical(
  formals(Shape2)$position,
  quote(draw::Coordinate())
))
```


t-kalinowski:
This approach of using `pkgname::classname()` might now work, because `new_class()` is expected to run at package build time, before package exports are fully known (i.e., before a call like `::` or `getNamespaceExports()` or `getExportedValue` will work correctly).

We might have to change the parent environment of the constructor to the frame where `new_class()` is called. Though that also brings challenges even in S7 itself, like:
```
** testing if installed package can be loaded from temporary location
Error: package or namespace load failed for ‘S7’:
 .onLoad failed in loadNamespace() for 'S7', details:
  call: NULL
  error: 'new_object' is not an exported object from 'namespace:S7'
Error: loading failed
```

lawremi:
Qualifying calls by the namespace should only be necessary when referencing class from another package. Would avoiding unnecessary qualification resolve the issue?

hadley:
This is starting to feel related to #341

t-kalinowski:
OOPWG notes: This should be fixed and merged before release.

R/class-spec.R
```diff
@@ -81,27 +81,36 @@ class_friendly <- function(x) {
...
+  # If the constructor is an S7 class that is exported from a package, avoid
+  # inlining the full class def instead, inline an expression like
+  # `pkgname::classname()` or `classname()`
+  if (is_class(f) && !is.null(f@package)) {
+    # Check if the class can be resolved as a bare symbol without pkgname::
+    if (identical(package, f@package)) {
+      return(call(f@name))
+    } else {
+      # namespace the pkgname::classname() call
+      cl <- as.call(list(quote(`::`), as.name(f@package), as.name(f@name)))
```
hadley, line 100: Is it worth double-checking that the method is actually exported from the package here? (And erroring informatively, if not)

R/constructor.R
```diff
@@ -91,3 +109,14 @@ as_names <- function(x, named = FALSE) {
   }
   lapply(x, as.name)
 }
+
+has_S7_symbols <- function(env, ...) {
```
hadley, line 113: Maybe `has_S7_imports`?

R/constructor.R
```diff
@@ -91,3 +109,14 @@ as_names <- function(x, named = FALSE) {
...
+  env <- topenv(env)
+  if (identical(env, asNamespace("S7")))
+    return (TRUE)
+  if (!isNamespace(env))
+    return (FALSE)
+  imports <- getNamespaceImports(env)[["S7"]]
+  symbols <- c(...) %||% getNamespaceExports("S7")
```
hadley, line 120: Personally, for an internal function like this, I wouldn't use `...` (instead requiring a character vector) and I wouldn't provide a default value.

tests/testthat/test-constructor.R
```diff
@@ -193,3 +191,14 @@ test_that("Dynamic settable properties are included in constructor", {
   expect_equal(foo@dynamic_settable, 1)
 
 })
+
+test_that("package exported classes are not inlined in constructor formals", {
+  # https://github.com/RConsortium/S7/issues/477
+  Foo := new_class(package = "pkgname")
+  Bar := new_class(properties = list(foo = Foo))
+
+  expect_identical(
```
hadley, line 200: Maybe a snapshot test would be useful here too, just so we can see what printing the formals looks like?

R/class-spec.R
```diff
@@ -81,27 +81,36 @@ class_friendly <- function(x) {
...
-class_construct_expr <- function(.x, ...) {
+class_construct_expr <- function(.x, envir = NULL, package = NULL) {
   f <- class_constructor(.x)
+
+  # If the constructor is an S7 class that is exported from a package, avoid
+  # inlining the full class def instead, inline an expression like
+  # `pkgname::classname()` or `classname()`
+  if (is_class(f) && !is.null(f@package)) {
+    # Check if the class can be resolved as a bare symbol without pkgname::
+    if (identical(package, f@package)) {
```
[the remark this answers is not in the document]
t-kalinowski: We probably want to update the docs in `?new_class`, `package` arg. Currently, it says:

> Setting package implies that the class is available for external use, so should be accompanied by exporting the constructor.

R/constructor.R
t-kalinowski: That's fine, but slightly less accurate. It also needs to return `TRUE` for the S7 namespace itself, which does not "import" S7 symbols.

R/class-spec.R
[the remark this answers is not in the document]
t-kalinowski: > e.g. when you're extending a non-exported class from the current package?

Whether it's exported or not, using `pkgname::` within package `pkgname` won’t work, because at build time, `::` will throw an error. Therefore, we need to use a bare symbol here.

R/constructor.R
hadley: Hmmm, true.

R/class-spec.R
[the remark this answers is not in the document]
hadley: Could you add a comment to that effect? I'm not sure this will be obvious to us (or at least me) later on.

tests/testthat/test-constructor.R
```diff
@@ -193,3 +191,16 @@ test_that("Dynamic settable properties are included in constructor", {
...
+    formals(Bar)$foo,
+    quote(pkgname::Foo())
+  )
+
+  expect_snapshot(formals(Bar))
```
hadley, line 205: I was thinking using `args()` to show the formals here.

tests/testthat/_snaps/external-generic.md
```diff
@@ -5,3 +5,41 @@
...
+
+    Code
+      args(Foo)
+    Output
+      function (bar = t0::AnS7Class()) 
+      NULL
+    Code
+      args(t2::AnS7Class2)
+    Output
+      function (bar = t0::AnS7Class())
```
hadley, line 18: Random thought — should probably check with a class with a non-syntactic name too.

hadley (approved):

R/class-spec.R
hadley: I like this 😄

t-kalinowski: Done. To do it properly required updating the relevant unit tests to use an actual package.

tests/testthat/test-constructor.R
t-kalinowski: Replaced this test with an equivalent one using the `t0` and `t2` test packages.

R/constructor.R
t-kalinowski: worth updating?

PR #482, opened 2024-10-25
Show generic signature in error message

### LINKED ISSUE #480: Method arg mismatch error is not actionable
```
Error: chat_request() lacks `...` so method formals must match generic formals exactly
```

Would be helpful to include what it has and what it's supposed to have.

t-kalinowski:
closes #480

### DIFF
--- a/R/method-register.R
+++ b/R/method-register.R
@@ -181,8 +181,8 @@ check_method <- function(method, generic, name = paste0(generic@name, "(???)")) 
 
   if (!"..." %in% generic_args && !identical(generic_formals, method_formals)) {
     msg <- sprintf(
-      "%s() lacks `...` so method formals must match generic formals exactly",
-      generic@name
+      "%s lacks `...` so method formals must match generic formals exactly",
+      show_args(formals(generic), name = generic@name)
     )
     stop(msg, call. = FALSE)
   }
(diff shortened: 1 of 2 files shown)


hadley:
I think it also needs to show what you need, i.e. the args to the generic.

t-kalinowski:
I might be misunderstanding, but it does show the generic args. Do you mean we should also show the "wrong" method args in the error message?

hadley:
Oh sorry yeah, it should show both the generic and the method args.

R/method-register.R
```diff
@@ -181,9 +181,14 @@ check_method <- function(method, generic, name = paste0(generic@name, "(???)"))
...
   if (!"..." %in% generic_args && !identical(generic_formals, method_formals)) {
     msg <- sprintf(
-      "%s lacks `...` so method formals must match generic formals exactly",
-      show_args(formals(generic), name = generic@name)
+      "%s() generic lacks `...` so method formals must match generic formals exactly.",
+      generic@name
     )
+    bullets <- c(
+      sprintf("- generic formals: %s", show_args(generic_formals, name = generic@name)),
+      sprintf("-  method formals: %s", show_args(method_formals, name = generic@name))
```
hadley, line 189:
```suggestion
      sprintf("- method formals: %s", show_args(method_formals, name = generic@name))
```

hadley: That makes the alignment look better to me.

hadley (approved):
Perfect, thanks!

Issue #471, opened 2024-10-21
`PRCODE(.)`  and  `SET_PRVALUE()`  are not API


mmaechler:
The first is  `    if (PRCODE(arg) == R_MissingArg) {  .... } ` here
https://github.com/RConsortium/S7/blob/b5cf389d7c6dc676c2d938e2feac6d7ec04d0c50/src/method-dispatch.c#L53
and basically identically (with a `!=` instead of `==`) also on line 156 .

As we know, we (R core) *have*  "solidified" R's  C API  recently, and visibly in `R-devel` (to be released in Spring 2025),
and with (somewhat extended) tests, one now gets
```
* checking compiled code ... NOTE
File ‘S7/libs/S7.so’:
  Found non-API calls to R: ‘PRCODE’, ‘SET_PRVALUE’

Compiled code should not call non-API entry points in R.

See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
and section ‘Moving into C API compliance’ for issues with the use of
non-API entry points.
```
The WRE section is https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Moving-into-C-API-compliance

t-kalinowski:
Is there an API for working with promises?

I might be mistaken, but I don't think there are API entrypoints yet for working with promises. As far as I can tell, there is currently no approved way to:
  - Determine if a symbol is a promise in a frame.
  - Get a reference to a promise in a frame.
  - Safely test if a symbol is missing without calling R-level `base::missing()`.
  - Access the promise expression or environment.

t-kalinowski:
Following up on this, I've experimented with some approaches to avoid directly accessing promise internals while maintaining efficient method dispatch. In PR #483, I drafted a solution using `Rf_eval()` and an R-level call to `base::missing()` to avoid accessing promise internals. While this works in most cases, there's one significant regression: when a method calls `substitute(some_arg)`, the original expression is lost if we've forced the promise of a dispatch argument. This regression makes the solution unsuitable for release.

However, another potential approach uses the relatively new `Exec()` function to perform dispatch while preserving dispatch argument promises for method introspection. Since `Exec()` is relatively new and wasn't specifically designed for this purpose, we should likely defer this implementation to a subsequent release. This would give us time to identify any unintended consequences. Additionally, we'll need to maintain the current implementation for older R versions where `Exec()` isn't available.

Here's a sketch of how method dispatch might work using `Exec()`:

``` r
some_generic <- function(x, ...) {
[...]

PR #483, opened 2024-10-28
Refactor `method-dispatch.c`

t-kalinowski:
This PR has fixes for current `rchk` warning, as well as non-API usage warnings. 

rchk warnings:
https://raw.githubusercontent.com/kalibera/cran-checks/master/rchk/results/S7.out

```r
Package S7 version 0.1.1
Package built using 86189/R 4.4.0; x86_64-pc-linux-gnu; 2024-03-25 22:22:05 UTC; unix   
Checked with rchk version fdc068715daa3a256062cc20e0d4a5157dacc9a4 LLVM version 14.0.6
More information at https://github.com/kalibera/cran-checks/blob/master/rchk/PROTECT.md
For rchk in docker image see https://github.com/kalibera/rchk/blob/master/doc/DOCKER.md

Function S7_obj_dispatch
  [UP] allocating function Rf_findVarInFrame(?,S:obj_dispatch) may destroy its unprotected argument (ns <arg 1>), which is later used. S7/src/method-dispatch.c:111
  [UP] calling allocating function Rf_findVarInFrame(?,S:obj_dispatch) with a fresh pointer (ns <arg 1>) S7/src/method-dispatch.c:111
  [UP] unprotected variable ns while calling allocating function Rf_install S7/src/method-dispatch.c:111
  [UP] unprotected variable ns while calling allocating function Rf_lang2 S7/src/method-dispatch.c:114
  [UP] calling allocating function Rf_eval with a fresh pointer (ns <arg 2>) S7/src/method-dispatch.c:115

Function generic_args
  [UP] calling allocating function Rf_eval with a fresh pointer (arg <arg 1>) S7/src/method-dispatch.c:57

Function method_
  [UP] unprotected variable m while calling allocating function Rf_asInteger S7/src/method-dispatch.c:98

Function method_call_
  [UP] allocating function Rf_eval may destroy its unprotected argument (arg <arg 1>), which is later used. S7/src/method-dispatch.c:158
  [UP] calling allocating function Rf_eval with a fresh pointer (arg <arg 1>) S7/src/method-dispatch.c:158

Function method_rec
  [UP] calling allocating function method_rec with a fresh pointer (val <arg 1>) S7/src/method-dispatch.c:20
  [UP] calling allocating function method_rec with a fresh pointer (val <arg 1>) S7/src/method-dispatch.c:30
```

non-API: https://github.com/RConsortium/S7/issues/471
```
Result: NOTE
  File ‘S7/libs/S7.so’:
    Found non-API calls to R: ‘PRCODE’, ‘SET_PRVALUE’
  
  Compiled code should not call non-API entry points in R.
  
  See ‘Writing portable packages’ in the ‘Writing R Extensions’ manual,
  and section ‘Moving into C API compliance’ for issues with the use of
  non-API entry points.
```

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -5,23 +5,26 @@
 
 /* .Call calls */
 extern SEXP method_(SEXP, SEXP, SEXP, SEXP);
-extern SEXP method_call_(SEXP, SEXP, SEXP);
+extern SEXP method_call_(SEXP, SEXP, SEXP, SEXP);
+extern SEXP test_call_(SEXP, SEXP, SEXP, SEXP);
 extern SEXP S7_class_(SEXP, SEXP);
 extern SEXP S7_object_(void);
 extern SEXP prop_(SEXP, SEXP);
 extern SEXP prop_set_(SEXP, SEXP, SEXP, SEXP);
 
+#define CALLDEF(name, n)  {#name, (DL_FUNC) &name, n}
+
 static const R_CallMethodDef CallEntries[] = {
-    {"method_", (DL_FUNC) &method_, 4},
-    {"S7_object_", (DL_FUNC) &S7_object_, 0},
-    {"prop_", (DL_FUNC) &prop_, 2},
-    {"prop_set_", (DL_FUNC) &prop_set_, 4},
+    CALLDEF(method_, 4),
+    CALLDEF(S7_object_, 0),
+    CALLDEF(prop_, 2),
+    CALLDEF(prop_set_, 4),
     {NULL, NULL, 0}
 };
 
 static const R_ExternalMethodDef ExternalEntries[] = {
-  {"method_call_", (DL_FUNC) &method_call_, 2},
-  {NULL, NULL, 0}
+    CALLDEF(method_call_, 2),
+    {NULL, NULL, 0}
 };
 
 SEXP sym_ANY;
--- a/src/init.c
+++ b/src/init.c
@@ -13,11 +13,15 @@ extern SEXP prop_set_(SEXP, SEXP, SEXP, SEXP);
 
 static const R_CallMethodDef CallEntries[] = {
     {"method_", (DL_FUNC) &method_, 4},
-    {"method_call_", (DL_FUNC) &method_call_, 3},
     {"S7_object_", (DL_FUNC) &S7_object_, 0},
     {"prop_", (DL_FUNC) &prop_, 2},
     {"prop_set_", (DL_FUNC) &prop_set_, 4},
     {NULL, NULL, 0}
+};
+
+static const R_ExternalMethodDef ExternalEntries[] = {
+  {"method_call_", (DL_FUNC) &method_call_, 2},
+  {NULL, NULL, 0}
 };
 
 SEXP sym_ANY;
@@ -47,11 +51,11 @@ SEXP ns_S7;
 
 void R_init_S7(DllInfo *dll)
 {
-    R_registerRoutines(dll, NULL, CallEntries, NULL, NULL);
+    R_registerRoutines(dll, NULL, CallEntries, NULL, ExternalEntries);
     R_useDynamicSymbols(dll, FALSE);
+
     sym_ANY = Rf_install("ANY");
     sym_S7_class = Rf_install("S7_class");
-
     sym_name = Rf_install("name");
     sym_parent = Rf_install("parent");
     sym_package = Rf_install("package");
@@ -70,5 +74,8 @@ void R_init_S7(DllInfo *dll)
     fn_base_quote = Rf_eval(Rf_install("quote"), R_BaseEnv);
     fn_base_missing = Rf_eval(Rf_install("missing"), R_BaseEnv);
 
-    ns_S7 = Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("S7"));
+    SEXP S7_str = PROTECT(Rf_mkString("S7"));
+    SEXP getS7ns_call = PROTECT(Rf_lang2(Rf_install("getNamespace"), S7_str));
+    ns_S7 = Rf_eval(getS7ns_call, R_BaseEnv);
+    UNPROTECT(2);
 }
--- a/src/init.c
+++ b/src/init.c
@@ -74,8 +74,5 @@ void R_init_S7(DllInfo *dll)
     fn_base_quote = Rf_eval(Rf_install("quote"), R_BaseEnv);
     fn_base_missing = Rf_eval(Rf_install("missing"), R_BaseEnv);
 
-    SEXP S7_str = PROTECT(Rf_mkString("S7"));
-    SEXP getS7ns_call = PROTECT(Rf_lang2(Rf_install("getNamespace"), S7_str));
-    ns_S7 = Rf_eval(getS7ns_call, R_BaseEnv);
-    UNPROTECT(2);
+    ns_S7 = Rf_findVarInFrame(R_NamespaceRegistry, Rf_install("S7"));
 }
--- a/src/init.c
+++ b/src/init.c
@@ -35,8 +35,12 @@ SEXP sym_getter;
 SEXP sym_dot_should_validate;
 SEXP sym_dot_getting_prop;
 SEXP sym_dot_setting_prop;
+SEXP sym_obj_dispatch;
+SEXP sym_dispatch_args;
+SEXP sym_methods;
 
 SEXP fn_base_quote;
+SEXP fn_base_missing;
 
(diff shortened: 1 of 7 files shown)


mmaechler (commented):
I cannot really approve (as I did not study the check failures),
but __congratulations__ on this refactoring, including no longer working with "promise-property"s which really are meant to be off-API

t-kalinowski:
After exploring the removal of `PRCODE()`, I unfortunately had to keep it, as the tradeoffs are too costly until additional API entry points become available. I shared some findings and discussion in issue #471.

That said, I did update non-API calls where possible while I was at it. I also addressed all rchk flagged warnings (many of which are spurious or false positives, but adding a few extra `PROTECT()` calls seems fine).

`rchk` now flags no issues when run locally; rchk can now be run CI too: https://github.com/RConsortium/S7/actions/runs/11582187471/job/32244624698

t-kalinowski:
I thought I had moved on, but woke up this morning with the solution for how to remove `PRCODE` and `SET_PRVALUE`, with minimal impact. Merging this should close #471 now.

mmaechler (approved):

src/init.c
```diff
@@ -5,18 +5,25 @@
...
-    {"prop_set_", (DL_FUNC) &prop_set_, 4},
+    CALLDEF(method_, 4),
+    CALLDEF(S7_object_, 0),
+    CALLDEF(prop_, 2),
+    CALLDEF(prop_set_, 4),
+    {NULL, NULL, 0}
+};
+
+static const R_ExternalMethodDef ExternalEntries[] = {
+    CALLDEF(method_call_, 2),
```
hadley, line 26: Can you explain why you made this external?

src/method-dispatch.c
```diff
@@ -4,6 +4,38 @@
...
+extern SEXP sym_dispatch_args;
+extern SEXP sym_methods;
+extern SEXP fn_base_quote;
+extern SEXP fn_base_missing;
+
+extern SEXP R_TRUE;
+
+
+static inline
+void APPEND_NODE(SEXP node, SEXP val, SEXP tag) {
```
hadley, line 18: I would've put `tag` before `val`? Not sure why, but that ordering feels more logical to me.

src/method-dispatch.c
```diff
@@ -145,47 +183,68 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
...
+
   // For each of the arguments to the generic
   for (R_xlen_t i = 0; i < n_args; ++i) {
 
     // Find its name and look up its value (a promise)
     SEXP name = TAG(formals);
-    SEXP arg = Rf_findVar(name, envir);
+    SEXP arg = Rf_findVarInFrame(envir, name);
 
+    // n_dispatch is always either 1 or 2.
```
hadley, line 197: Can't it be any number?

src/method-dispatch.c
```diff
@@ -145,47 +183,68 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
...
   SEXP mcall_tail = mcall;
 
+  PROTECT_INDEX arg_pi;
+  PROTECT_WITH_INDEX(R_NilValue, &arg_pi);
+  ++n_protect;
+
   // For each of the arguments to the generic
   for (R_xlen_t i = 0; i < n_args; ++i) {
 
     // Find its name and look up its value (a promise)
```
hadley, line 193: Is this comment no longer correct?

src/method-dispatch.c
```diff
@@ -145,47 +183,68 @@ SEXP method_call_(SEXP call, SEXP generic, SEXP envir) {
...
-            SET_PRVALUE(arg, val);
-          }
+      } else { // arg not missing, is a PROMSXP
 
-          // Then add to arguments of method call
-          SETCDR(mcall_tail, Rf_cons(arg, R_NilValue));
+        // Force the promise so we can look up its class.
+        // However, we preserve and pass along the promise itself so that
+        // methods can still call substitute()
+        // Instead of Rf_eval(arg, R_EmptyEnv), we do Rf_eval(name, envir), so that
```
hadley, line 210: Nice!!

tests/testthat/test-method-dispatch.R
```diff
@@ -186,3 +186,17 @@ test_that("can dispatch on evaluated arguments", {
...
+test_that("method dispatch works for class_missing", {
+
+  foo <- new_generic("foo", "x")
+  method(foo, class_missing) <- function(x) missing(x)
+
+  expect_true(foo())
+
+  # dispatch on class_missing only works directly in the generic call
+  foo_wrapper <- function(xx) foo(xx)
+  expect_error(foo_wrapper(), 'argument "xx" is missing, with no default')
```
hadley, line 200: Slightly safer to use `expect_snapshot()` here, or otherwise guard so this code isn't run on CRAN. Otherwise if base R ever changes the error message, this test would spuriously fail.

hadley (approved):
Nice work!

Does this happen to fix this commented out test?

```R
  # method(foo, class_character) <- function(x, ..., z = 1) substitute(z)
  # expect_equal(foo("x", z = letters), quote(letters))
```

I also double checked that we already had a test that ensures that the dispatch argument is only evaluated once, so there's no chance of a regression there.

hadley:
Also would you mind doing a little benchmark of before and after just to verify that there's no performance regression?

src/init.c
t-kalinowski: `.External2` is the only interface that preserves the visibility of the returned value, `.Call` and the other interfaces all force visibility on (the "2" in "201"). 

image
image

src/method-dispatch.c
t-kalinowski: Hmm, when writing the R-dev blog I remember an edit made (I forget by who) adding/correcting this line: 

> while S4 generics can dispatch on an arbitrary number of arguments. S7 implements double dispatch (i.e. multiple dispatch on exactly two arguments) 

but testing locally just now it seems we *do* support dispatch on more than one arg. 

Thanks for the correction!

t-kalinowski:
I knitted the "performance" vignette locally and compared it before and after, and generally, the performance is similar or slightly faster. I'll verify one last time and post here before merging.

t-kalinowski:
> Does this happen to fix this commented out test?

I tweaked how we pass along the non-dispatch arguments and now `substitute()` works on those too! I also expanded the unit tests to cover `...` args better.

t-kalinowski:
I ran some quick benchmarks on dispatch by re-rendering the "performance" vignette on this branch and the main branch. At a glance, performance seems to have improved by about 20%.

For now, I’m including some screenshots of the vignette diffs. At some point, we may want to set up more consistent and rigorous benchmarks to track performance over time.


```sh
# Helper function to install and render vignette
render_vignette() {
  local branch=$(git branch --show-current)
  local output_file="performance_${branch}.md"
  local output_dir="~/github/RConsortium/S7/benchmarks/${branch}"

  # Install package and render vignette (warmup and actual run)
  Rscript -e 'remotes::install_local(force=TRUE)'
  Rscript -e "rmarkdown::render('~/github/RConsortium/S7/vignettes/performance.Rmd', encoding = 'UTF-8', output_format = 'github_document', output_file = '${output_file}', output_dir = '${output_dir}');"
  Rscript -e "rmarkdown::render('~/github/RConsortium/S7/vignettes/performance.Rmd', encoding = 'UTF-8', output_format = 'github_document', output_file = '${output_file}', output_dir = '${output_dir}');"
}

# Render vignette on main branch
git switch main
render_vignette "main"

# Render vignette on PR branch
export branch_name=rchk-and-non-API-fixes
git switch $branch_name
render_vignette $branch_name

# Diff the two generated Markdown files
code --diff ~/github/RConsortium/S7/benchmarks/$branch_name/performance_$branch_name.md ~/github/RConsortium/S7/benchmarks/main/performance_main.md
```

(this branch on the left, main branch on the right)

image

image

image

image

src/init.c
hadley: Good to know!

hadley:
Great! Thanks so much for all your work on this.

hadley:
Should we add a test for the visibility issue?

t-kalinowski:
There’s already a unit test here: https://github.com/RConsortium/S7/blob/498cfad96f727b3f90c540e196d3ec16c9d34777/tests/testthat/test-method-dispatch.R#L190-L202. Do you think this covers the visibility issue?

hadley:
Oh yeah, that's perfect. I thought it was a new feature due to the use of `.External`.

Issue #442, opened 2024-09-15
Defining a property named "..." should not be allowed


lawremi:
Yea, I come up with some crazy stuff.

```r
> cls <- new_class("cls", properties = list(... = class_list))
> cls()
Error in cls() : '...' used in an incorrect context
```

hadley:
Looks like this is already implemented

``` r
library(S7)
cls <- new_class("cls", properties = list(... = class_list))
#> Error in `new_class()`:
#> ! property can't be named: ...
```

<sup>Created on 2026-05-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

PR #484, opened 2024-10-30
Reserve some property names

### LINKED ISSUE #425: Using `class` property seems to clash with object's class
``` r
library(S7)
packageVersion("S7")
#> [1] '0.1.1.9000'

Foo <- new_class(
  "Foo",
  properties = list(
    class = new_property(class_character, default = "Class"),
    other = new_property(class_character, default = "Other")
  )
)

x <- Foo()

x@other
#> [1] "Other"
x@class # retrieves information about x
#> [1] "Foo"       "S7_object"
class(x)
#> [1] "Foo"       "S7_object"

# sets fine
x@other <- "New"
x@other
#> [1] "New"

# fails
try({
  x@class <- "New"
  x@class
})
#> Error : `object` must be an <S7_object>, not a S3<New>
```

<sup>Created on 2024-08-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

t-kalinowski:
closes #425, #442

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -333,7 +333,7 @@ S7_class <- function(object) {
 check_prop_names <- function(properties, error_call = sys.call(-1L)) {
   # these attributes have special C handlers in base R
   forbidden <- c("names", "dim", "dimnames", "class",
-                 "tsp", "comment", "row.names")
+                 "tsp", "comment", "row.names", "...")
   forbidden <- intersect(forbidden, names(properties))
   if (length(forbidden)) {
     msg <- paste0("property can't be named: ",
--- a/R/class.R
+++ b/R/class.R
@@ -130,6 +130,7 @@ new_class <- function(
   # Combine properties from parent, overriding as needed
   all_props <- attr(parent, "properties", exact = TRUE) %||% list()
   new_props <- as_properties(properties)
+  check_prop_names(parent, new_props)
   all_props[names(new_props)] <- new_props
 
   if (is.null(constructor)) {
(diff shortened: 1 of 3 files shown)


R/class.R
```diff
@@ -327,3 +328,16 @@ str.S7_object <- function(object, ..., nest.lev = 0) {
 S7_class <- function(object) {
   attr(object, "S7_class", exact = TRUE)
 }
+
+
+check_prop_names <- function(parent, properties, error_call = sys.call(-1L)) {
+  forbidden <- c("names", "dim", "dimnames", "class", "tsp",
+                 "comment", "row.names",
+                 if ("factor" %in% class_dispatch(parent)) "levels")
```
hadley, line 336: What's the justification for `levels`? I know the others get special handling at the C-level, but I didn't know about `factor`.

Also add `...` to handle #442?

mmaechler: Good point.  Very early in R's history,  factors *were* dealt with in C code in parts,  but I think this and `"levels"` in particular should no longer be specials.    {{but then I did not search R's C sources now}}

t-kalinowski: Indeed, levels isn't handled specially at the C level: https://github.com/wch/r-source/blob/c06d2576f10fc82a4e06120df79f32eeaf044f99/src/main/attrib.c#L257-L270

I included it not to avoid special C handlers, but rather to acknowledge that S7 might not "own" the attribute or be responsible for managing it. However, that's probably unnecessarily restrictive. I'll remove it and update the forbidden names list to include only the attribute names with special C handlers.

hadley (approved):

PR #490, opened 2024-11-06
fix unary Ops methods: `-`, `+`, `!`

t-kalinowski:


mjskay:
I found my way to this PR when I discovered that adding a method for `!` did not work in a class I was writing. 

From my tests, this PR does not fix this issue for `!`; e.g. I adapted [your test for `Double`](https://github.com/RConsortium/S7/pull/490/files#diff-321694582d52112c44db84864522dea900fb4018f3dd36c1f90f29722e992f5cR128-R133) to a `Logical` and it fails:

```r
Logical := new_class(class_logical)

method(`!`, list(Logical, class_missing)) <- function(e1, e2) {
    Logical(!as.logical(e1))
}
## Error: `generic` is a function, but not an S3 generic function: 
## .Primitive("!")
```

I think the problem is that `"!" %in% group_generics()$Ops` is `FALSE` when it should be `TRUE`, which stems from the fact that `!` is in the `Ops` S3 group generic but not the `Arith`, `Compare`, `Logic`, or `Ops` S4 group generics.

As a hackish solution, if I add `out$Ops <- union(out$Ops, "!")` somewhere after line 116 here:

https://github.com/RConsortium/S7/blob/f75e089526bd59d217aeb9f42b5f9dcbab51c7ce/R/generic-spec.R#L107-L121

the test on `!` succeeds, though I imagine there should be a more elegant solution.

t-kalinowski:
Thanks for the report and workaround! 

This PR will likely be finished / rolled-into https://github.com/RConsortium/S7/pull/365.

One workaround you can use until this is fixed is to define an S3 method the S3 way.

```r
#' @export
`!.mypackage::Logical` <- function(x) { .... }
```

That workaround should work now, and continue to work once this is fixed in S7.

mjskay:
Great thanks! That helps, because I was defining `!.ClassName` in a previous version of S7 and it stopped working in the latest version; I had missed that the package name is now prepended to the class name.

R/method-ops.R
```diff
@@ -17,11 +17,16 @@ on_load_define_ops <- function() {
 #' @export
 Ops.S7_object <- function(e1, e2) {
   cnd <- tryCatch(
-    return(base_ops[[.Generic]](e1, e2)),
+    return(
+      if (missing(e2))
+        base_ops[[.Generic]](e1)
```
hadley: Put the `return()` inside the `if`?

hadley (approved):

hadley:
I finished off the implementation, adding a few more tests (revealing a problem with the logic) and then also fixing `!` (which is a weird member of `Ops`).

I explored letting `method(`-`, foo) <- function(e1, e2) ... ` work but IMO the juice wasn't worth squeeze.

PR #492, opened 2024-11-09
Vignette typos fixes

mattheaphy:
A few grammatical fixes, updates of old function / argument names, and example code updates in 3 vignettes.

Regarding code updates:

#### `classes-objects.Rmd`

In the example that included a `getter` and a `setter` for the `Range` class, the `setter` was getting called on object instantiation. This resulted in the supplied value of `@end` being overridden to an empty vector because `@length`'s default value is empty. The resulting output of `Range(start = 1, end = 10)` was this:

``` .r
<Range>
 @ start : num 1
 @ end   : num(0) # should be 10
 @ length: num(0) # should be 9
```

I updated the length `setter` to short circuit and return the original object when `value` is empty, and that gave me the correct output. I wanted to call this out because the maintainers might have a better solution in mind for handling instantiation matters like this.

#### `compatibility.Rmd`

I noticed that the more natural representation of the `rle` class using S7 was overwriting the `rle()` function, and this was creating incorrect output for `rle(1:10)`. 

``` .r
Run Length Encoding
  lengths: int [1:10] 1 2 3 4 5 6 7 8 9 10 # should be all 1's
  values : logi(0) # should be 1-10
```

I used the name `new_rle` here instead that seemed to clean things up.

### DIFF
--- a/vignettes/classes-objects.Rmd
+++ b/vignettes/classes-objects.Rmd
@@ -257,6 +257,7 @@ Range <- new_class("Range",
       class = class_double,
       getter = function(self) self@end - self@start,
       setter = function(self, value) {
+        if (length(value) == 0) return(self)
         self@end <- self@start + value
         self
       }
--- a/vignettes/classes-objects.Rmd
+++ b/vignettes/classes-objects.Rmd
@@ -257,7 +257,11 @@ Range <- new_class("Range",
       class = class_double,
       getter = function(self) self@end - self@start,
       setter = function(self, value) {
-        if (length(value) == 0) return(self)
+        if (!length(value)) {
+          # Do nothing if called with the constructor default
+          # value for this property, a zero-length double vector.
+          return(self)
+        }
         self@end <- self@start + value
         self
       }
@@ -393,8 +397,8 @@ Range <- new_class("Range",
     end = class_numeric
   ),
   constructor = function(x) {
-    new_object(S7_object(), 
-               start = min(x, na.rm = TRUE), 
+    new_object(S7_object(),
+               start = min(x, na.rm = TRUE),
                end = max(x, na.rm = TRUE))
   }
 )
--- a/vignettes/classes-objects.Rmd
+++ b/vignettes/classes-objects.Rmd
@@ -24,7 +24,7 @@ library(S7)
 ## Validation
 
 S7 classes can have an optional **validator** that checks that the values of the properties are OK.
-A validator is a function that takes the object (called `self`) and returns `NULL` if its valid or returns a character vector listing the problems.
+A validator is a function that takes the object (called `self`) and returns `NULL` if it's valid or returns a character vector listing the problems.
 
 ### Basics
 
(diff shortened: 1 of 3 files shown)


vignettes/classes-objects.Rmd
```diff
@@ -257,6 +257,7 @@ Range <- new_class("Range",
       class = class_double,
       getter = function(self) self@end - self@start,
       setter = function(self, value) {
+        if (length(value) == 0) return(self)
```
hadley, line 260: @t-kalinowski it probably makes sense for this to be an explicit `is.null()` test, right?

vignettes/compatibility.Rmd
```diff
@@ -109,7 +109,7 @@ rle(1:10)
 Alternatively you could convert it to the most natural representation using S7:
 
 ```{r}
-rle <- new_class("rle", properties = list(
+new_rle <- new_class("rle", properties = list(
```
hadley, line 112: Maybe `rle2`? I think calling it `new_` is a bit confusing because the while the object is constructor, it's also the class.

vignettes/generics-methods.Rmd
```diff
@@ -213,7 +213,7 @@ The only downside to performing error checking is that you constraint the interf
 
 ## `super()`
 
-Sometimes it's useful to define a method for in terms of its superclass.
+Sometimes it's useful to define a method for a class that relies on its superclass.
```
hadley, line 216:
```suggestion
Sometimes it's useful to define a method for a class that relies on the implementation for its superclass.
```
?

vignettes/classes-objects.Rmd
t-kalinowski: I don't think so. The default value passed in by the constructor will be `double(0)`. If we want the default to be `NULL`, we'd have to change the property class to `NULL | class_double`, or implement a custom constructor.

vignettes/compatibility.Rmd
t-kalinowski: I agree, `rle2` would be clearer than `new_rle`.

mattheaphy: There's a few other instances north of this line where the name `new_rle` is used that would have to be updated as well for the examples to work as the original author intended. I'll add that change. I agree that the examples are in this section of the vignette are a little difficult to reason about because the new S7 class is being asked to wear the hat of a formal class definition and a constructor for the "rle" class.

t-kalinowski: Thanks for catching this! After looking more closely, I see that `new_rle()` is indeed the best choice here. The intent is for `rle()` to serve as a convenient constructor, and we show how to redefine the underlying `new_rle()` from the “old-style” S3 approach to the S7 approach (in two different ways) while keeping the user-facing `rle()` constructor unchanged.

Strictly speaking, it might make sense to replace the user-facing `rle()` with an S7 `rle` class that has a custom constructor, but that would increase the example’s size and complexity without adding much clarity, in my opinion. I’ve restored the previous `new_rle()` approach.

t-kalinowski (approved):
Thank you for the edits!

hadley:
Thank you!

Issue #502, opened 2024-11-22
`object is invalid` error when `new_object` argument `.parent` is assigned `S7_object` instead of `S7_object()`


tschwarzl:
I would like to report a user error which leads to creation of seemingly valid S7 object, however, throws an error later.

When within the S7 `constructor` the function `new_object` is called and `.parent` is assigned `S7_object` instead of `S7_object()` a class is created without error. 
However, when assigning a property to an object of said class it **can** lead to `object is invalid: - Underlying data is corrupt` error.

Unfortunately, (and to my surprise) I was not able to create a minimal example reproducing the error so far. 

To prevent other users running into this rather hard to debug problem, I would like to discuss a type check for `.parent`

```{r}
> class(S7_object)
[1] "S7_class"  "S7_object"
> class(S7_object())
[1] "S7_object"
```

hadley:
Duplicate of #409

PR #503, opened 2024-11-22
added type checks for new_object() argument .parent 

tschwarzl:
addressing issue #502 

added type checks for `new_object()` argument `.parent` asserting that it is of type `S7 object` not `S7_class`

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -238,8 +238,8 @@ is_class <- function(x) inherits(x, "S7_class")
 
 # Object ------------------------------------------------------------------
 
-#' @param .parent,... Parent object and named properties used to construct the
-#'   object.
+#' @param .parent,... Parent S7 object and named properties used to construct
+#'   the object.
 #' @rdname new_class
 #' @export
 new_object <- function(.parent, ...) {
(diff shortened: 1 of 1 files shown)


R/class.R
```diff
@@ -238,8 +238,8 @@ is_class <- function(x) inherits(x, "S7_class")
 
 # Object ------------------------------------------------------------------
 
-#' @param .parent,... Parent object and named properties used to construct the
-#'   object.
+#' @param .parent,... Parent S7 object and named properties used to construct
```
hadley, line 241: This isn't necessarily an S7 object, because the class might extend a base class.

R/class.R
```diff
@@ -253,7 +253,15 @@ new_object <- function(.parent, ...) {
   }
 
   # force .parent before ...
-  # TODO: Some type checking on `.parent`?
+  # TODO: Some additional type checking on `.parent`?
+
+  if (!inherits(.parent, "S7_object"))
+    stop("`.parent` needs to be an S7_object")
+  if (inherits(.parent, "S7_class")) {
+    stop(paste("`.parent` cannot be of type S7_class. Did you type",
```
hadley, line 261: Can you please put this on two lines and add a snapshot test for it?

hadley (commented):
Thanks for working on this. This will be a nice improvement in user friendliness.

hadley:
Sorry I completely forgot about this and implemented #625 instead.

PR #505, opened 2024-11-23
Typo: remove superflous 'a'

lgatto:


hadley:
S7 uses roxygen2, so you'll need to make the edit in the underlying file, `R/property.R`.

lgatto:
Ah, yes, sorry. Will do and send another PR.

PR #507, opened 2024-11-25
Typo: remove superflous 'a' in R/property.R man page

lgatto:


mmaechler (approved):

hadley:
@mmaechler roxygen2 needs to be re-run so that the change makes it into the rendered docs. I'll do that now.

PR #575, opened 2025-11-04
Allow skipping validation steps in `props<-` and `set_props`

### LINKED ISSUE #574: `set_props` may be too expensive to use inside a package
Since validation steps generally tend to take a long time, I'm wondering if it's worth adding "unsafe" version of the functions for developers.

```r
# Because validation step is very expensive,
# this is a copy of `props<-`, the only difference is `check = FALSE`
`props_uncheck<-` <- function(object, value) {
  for (name in names(value)) {
    prop(object, name, check = FALSE) <- value[[name]]
  }

  object
}

# A copy of `set_props()`, using `props_uncheck<-`
# instead of `props<-` to avoid validation
set_props_uncheck <- function(object, ...) {
  props_uncheck(object) <- list(...)
  object
}
```

It may be possible to simply add the `check` argument, similar to `prop<-`, however, in that case, users would need to be aware of the risks involved, so I thought it might be better to separate the functions.

eitsupi:
Close #574

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # S7 (development version)
 
+* `props<-()` and `set_props()` gain `check`/`.check` arguments, letting you
+  set properties without calling `validate()` (#574, #575).
+
 # S7 0.2.0
 
 ## New features


eitsupi:
Thanks!

t-kalinowski:
Thank _you_!

PR #581, opened 2025-11-26
Allow creating another constructor (without validation) with `new_empty_object()`

### LINKED ISSUE #522: Multiple constructors? / Separation of S7 objects and constructors
Thank you for pushing this wonderful project forward. It looks very promising.

I've read the documentation, but in the current implementation, S7 class instances can only exist via the single constructor?
I have tried to imagine creating a wrapper object for an external pointer (specifically something like Series in the polars package) in S7.

The constructor must take a name (string) and values (R vector) and construct an external pointer stored in `_s`. (Note: Because it mimics Python's polars package)
The name can be changed later so it becomes a property.

```r
s_foo <- Series(name = "foo", values = 1)

s_foo@name
#> foo
s_foo@`_s`
#> <external pointer>

# The name can be changed
s_foo@name <- "bar"
s_foo@name
#> bar

# The pointer should be frozen
s_foo@`_s` <- other_pointer
#> Error
```

But apart from this, it also requires an (internal-only) function to create objects by directly setting an external pointer.
Is it possible to create such a thing?
Ideally, it would be called a class method or an associated function (related to #202), but it does not have to be.

```py
# In Python Polars, there are two class methods to construct Series
Series._from_pyseries() # From Rust / Rust Polars' Series
Series._import_from_c() # From C / Apache Arrow C interface
```

I want to export the S7 object itself (`Series`) for use in the `convert` function, etc., but do not want to export the constructor for manipulating the external pointer.
Therefore, I do not want the default constructor to be one that takes an external pointer.

```r
# Should be exported the S7 object
Series

# But I don't want to expose the constructor that builds the Series from an external pointer
Series(.data = rust_series)
```

eitsupi:
Close #522

Allows us to create a constructor separately from the S7 class definition.
The expected use is to create an empty (invalid) object when the package is loaded, and then the second constructor sets properties on it and creates a valid object.

It is also possible for the first constructor of the S7 class to depend on the second constructor.

The second constructor may completely bypass the validation step, making object creation much faster, which I think is very important in packages like polars where we create objects multiple times.
(It is entirely up to the package developer to ensure that the second constructor creates a valid object without validation.)

``` r
library(S7)

new_sums <- function(x) {
  out <- EMPTY_SUMS
  value <- x |> Reduce(f = `+`) |> as.integer()
  prop(out, "value", check = FALSE) <- value
  out
}

Sums <- new_class(
  "Sums",
  properties = list(value = class_integer),
  constructor = function(...) {
    v <- new_sums(list(...))@value
    new_object(
      S7_object(),
      value = v
    )
  }
)

EMPTY_SUMS <- new_empty_object(Sums)

bench::mark(
  sums1 = Sums(1, 2, 3),
  sums2 = new_sums(list(1, 2, 3))
)
#> # A tibble: 2 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 sums1       68.81µs     80µs    11654.   39.27KB     19.0
#> 2 sums2        4.92µs   6.29µs   143152.    4.14MB     28.6
```

<sup>Created on 2025-11-26 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


hadley:
Would you mind starting with an issue first? That will allow us to build consensus around the problem and the best solution.

hadley:
Closing since I think this needs further discussion and the implementation of `new_object()` is changing so this PR is no longer so relevant.

PR #582, opened 2025-11-26
Feature: permit (most) reserved attributes names as property names

### LINKED ISSUE #579: Allow fallback method of storing properties for "forbidden" names
I'm not sure if I've just had bad luck or if these forbidden properties names are just so useful that they pop up frequently, but this has been a prickly part of `S7` for me in a couple different projects now. I've only tried to use `S7` in earnest in two packages and both have butted heads with this constraint in different ways.

I'm referring to these [forbidden property names](https://github.com/RConsortium/S7/blob/bd70ec391abeb37ed560b16d89b40a3439b1e1dd/R/class.R#L334-L335).

Just to illustrate a couple cases where this has come up in my exploration so far, I've encountered this during:
- implementing the debug adapter protocol, which has classes that have `names` fields. I'm trying to be faithful to the protocol, but I need to have a special case for this one field because this is a forbidden property name.
- implementing a specialized version of a function that does return type checking, I store the return type as a `class` property, but again hit this constraint. Here I have more control over the naming convention, but was surprised to see that there are these limits.

I'd like to propose a couple alternative designs

1. The simplest solution would be to store properties in a `attr(obj, "properties")` `list` to avoid any constraints on the names. However, I assume that this was consciously avoided because of the extra lookup.
2. If a `list` is undesirable, perhaps introducing it only as a backup option for these specific elements so that nearly all properties remain surface-level attributes, but a few select forbidden names get relegated to an internal list. During property getting and setting, there would be branching logic based on whether the property name is forbidden or not.

dgkf:
Closes #579 

- In calls to `prop` and `prop<-`, I've remapped attribute names for use in `Rf_setAttrib` in `prop_` and `prop_set_` respectively.
- In `new_class`, remaps attributes where they're assigned directly
- Still raises an error for `...` as it raises a parse error when trying to use it via `obj@...`, but this is an R syntax error and it would still work via `prop(obj, "...")`. I'd probably lean toward even permitting this, but it seemed explicitly undesirable in #442
- Added a few tests
- Affected files were also formatted by `air`, let me know if you'd like whitespace changes reverted.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -95,14 +95,14 @@
 #' r <- Range(start = 10, end = 20)
 #' try(r@start <- 25)
 new_class <- function(
-    name,
-    parent = S7_object,
-    package = topNamespaceName(parent.frame()),
-    properties = list(),
-    abstract = FALSE,
-    constructor = NULL,
-    validator = NULL) {
-
+  name,
+  parent = S7_object,
+  package = topNamespaceName(parent.frame()),
+  properties = list(),
+  abstract = FALSE,
+  constructor = NULL,
+  validator = NULL
+) {
   check_name(name)
 
   parent <- as_class(parent)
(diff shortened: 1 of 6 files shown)


hadley:
I think we'll need to implement this in C to get the performance we want for prop getting/setting.

PR #583, opened 2025-11-28
Draft: Fix: Ignores validation internal to setters upon class creation

### LINKED ISSUE #416: Class & property names are missing from error message when `setter` is set
When `setter` is missing, the error prints out the message `@abracadabra must be <integer>, not <double>` so users know the cause of error

``` r
some_int <- S7::new_class(
  name = "some_int", 
  properties = list(
    abracadabra = S7::new_property(
      class = S7::class_integer
    )
  )
)
some_int(abracadabra = 1.1)
#> Error: <some_int> object properties are invalid:
#> - @abracadabra must be <integer>, not <double>

```

When `setter` is made, the error message is not informative

```r

another_int <- S7::new_class(
  name = "another_int", 
  properties = list(
    abracadabra = S7::new_property(
      class = S7::class_integer,
        self@abracadabra <- as.integer(value)
      }
    )
  )
)
another_int(abracadabra = 1.1)
#> Error: `object` must be an <S7_object>, not a <integer>
```

<sup>Created on 2024-08-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

dgkf:
Closes #416

- Property validators now get applied after a setter is called
- During `new_object` calls, property validators no longer raise errors immediately, instead reporting errors when the object is validated.

### DIFF
--- a/src/prop.c
+++ b/src/prop.c
@@ -355,14 +355,6 @@ SEXP prop_set_(SEXP object, SEXP name, SEXP check_sexp, SEXP value) {
   Rboolean should_validate_obj = check;
   Rboolean should_validate_prop = check;
 
-  // if already flagged for non-recursion, do not validate as this will be
-  // validated after top-level return
-  SEXP no_recurse_list = Rf_getAttrib(object, sym_dot_setting_prop);
-  if (pairlist_contains(no_recurse_list, name_sym)) {
-    should_validate_obj = FALSE;
-    should_validate_prop = FALSE;
-  }
-
   SEXP S7_class = Rf_getAttrib(object, sym_S7_class);
   SEXP properties = Rf_getAttrib(S7_class, sym_properties);
   SEXP property = extract_name(properties, name_char);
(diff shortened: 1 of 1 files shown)


src/prop.c
```diff
@@ -355,6 +355,14 @@ SEXP prop_set_(SEXP object, SEXP name, SEXP check_sexp, SEXP value) {
   Rboolean should_validate_obj = check;
   Rboolean should_validate_prop = check;
 
+  // if already flagged for non-recursion, do not validate as this will be
+  // validated after top-level return
+  SEXP no_recurse_list = Rf_getAttrib(object, sym_dot_setting_prop);
+  if (pairlist_contains(no_recurse_list, name_sym)) {
+    should_validate_obj = FALSE;
+    should_validate_prop = FALSE;
+  }
```
dgkf: This is necessary for `new_object` calls that initialize properties using setters. Previously, when a setter would contain a call to `prop()<-`, this call's `check = TRUE` would force immediate validation.

Now, even if `check = TRUE`, we also look at this recursive property setting flag to see if we should ignore it.

dgkf:
Closing this PR. After reviewing a few edge cases, I think this would require substantially larger changes to the way that recursion through various properties is handled. It was a fun dive, nevertheless.

Specifically, in cases where a setter affects _another_ property - one that hasn't yet been added to this `.setting_prop` flag in order to delay internal validation.

PR #604, opened 2026-04-11
fix: duplicate "error" in S7_error_method_not_found

jjjermiah:
purely cosmetic, but prevents having to `unique` the `class` for logs.

```r
fail <- S7::new_generic("fail", "x")
tryCatch(fail(TRUE), S7_error_method_not_found = \(e) print(class(e)))
[1] "S7_error_method_not_found" "error"                     "error"                     "condition" 
```

unsure if tightening the tests is useful

### DIFF
--- a/tests/testthat/test-method-dispatch.R
+++ b/tests/testthat/test-method-dispatch.R
@@ -167,8 +167,7 @@ test_that("single dispatch fails with informative messages", {
     fail(Foo(x = 1))
   })
 
-  cnd <- tryCatch(fail(TRUE), S7_error_method_not_found = identity)
-  expect_s3_class(cnd, c("S7_error_method_not_found", "error", "condition"), exact = TRUE)
+  expect_error(fail(TRUE), class = "S7_error_method_not_found")
 })
 
 test_that("multiple dispatch fails with informative messages", {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-method-dispatch.R
```diff
@@ -167,7 +167,8 @@ test_that("single dispatch fails with informative messages", {
     fail(Foo(x = 1))
   })
 
-  expect_error(fail(TRUE), class = "S7_error_method_not_found")
+  cnd <- tryCatch(fail(TRUE), S7_error_method_not_found = identity)
```
hadley: Given the minimal impact of this problem, I'd suggest creating one new test specifically for this issue like "error class not duplicated" and leaving the existing two tests as is.

hadley (commented):
Thanks for spotting this and making a fix!

tests/testthat/test-method-dispatch.R
jjjermiah: fixed, it looks a bit silly as a standalone test haha

hadley:
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

jjjermiah:
> Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

yup done!

PR #605, opened 2026-04-11
refactor: Add S7_error_validation classes (fixes #602)

### LINKED ISSUE #602: feature request:  validation error class
it would be nice if the S7 validation errors had their own subclass, so it can be handled differently

jjjermiah:
fixes #602

todo: 
- I think that this could be even more useful if it was documented for users as well?

### DIFF
--- a/R/valid.R
+++ b/R/valid.R
@@ -19,7 +19,26 @@
 #' @param recursive If `TRUE`, calls validator of parent classes recursively.
 #' @param properties If `TRUE`, the default, checks property types before
 #'   executing the validator.
-#' @returns Either `object` invisibly if valid, otherwise an error.
+#' @returns `object`, invisibly, if valid. Otherwise an error is signalled.
+#'
+#' @details # Validation errors 
+#'
+#' When validation fails, [validate()] signals one of the following condition
+#' classes, which can be caught with [tryCatch()]:
+#'
+#' * `S7_error_validation_property` for property type validation failures.
+#' * `S7_error_validation_object` for custom validator failures.
+#' * `S7_error_validation` as the parent class of both.
+#'
+#' These conditions include the following fields:
+#'
+#' * `message`: the error message.
+#' * `object_class`: the formatted class description, like `"<Range>"`.
+#' * `errors`: a character vector of validation errors.
+#'
+#' `S7_error_validation_property` is used both for a single property assignment
+#' failure from `prop<-` / `@<-` and for batched property validation from
+#' [validate()].
 #' @export
 #' @examples
 #' # A range class might validate that the start is less than the end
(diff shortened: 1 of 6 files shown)


R/valid.R
```diff
@@ -78,7 +78,8 @@ validate <- function(object, recursive = TRUE, properties = TRUE) {
     if (length(errors) > 0) {
       bullets <- paste0("- ", errors, collapse = "\n")
       msg <- sprintf("%s object properties are invalid:\n%s", obj_desc(object), bullets)
-      stop(msg, call. = FALSE)
+      stop(errorCondition(msg, call = NULL,
+        class = c("S7_error_validation_property", "S7_error_validation")))
```
hadley: Given the name of the other error class, maybe `S7_error_validation_failed"? I think you could just use the one class, since it's not very useful to know that an individual property failed to validate, if you don't know which one it was. (And it's probably not worth trying to track that).

tests/testthat/test-valid.R
```diff
@@ -90,3 +90,39 @@ test_that("valid implicitly does _not_ call the validation function", {
   })
   expect_error(validate(obj), "must be positive")
 })
+
+
+describe("validation error classes", {
```
hadley: Can you please use `test_that()` for new tests?

tests/testthat/test-valid.R
```diff
@@ -90,3 +90,39 @@ test_that("valid implicitly does _not_ call the validation function", {
...
+  klass <- new_class("klass", package = NULL,
+    properties = list(x = class_double),
+    validator = function(self) if (self@x < 0) "x must be positive"
+  )
+
+  it("property errors have class S7_error_validation_property", {
+    obj <- klass(1)
+    attr(obj, "x") <- "bad"
+
+    cnd <- tryCatch(validate(obj), S7_error_validation_property = identity)
```
hadley: Fine to just use `expect_error()` here I think.

hadley:
I think it would be good to document these, but it's not clear to me where to put that documentation such that folks are likely to find it.

tests/testthat/test-valid.R
jjjermiah: yup, curious if there is a preference because of it not behaving as expected? or is it the nature of the test?

R/valid.R
jjjermiah: sure thing!
I wanted to make the case for having more informative error types with structured metadata fields but ended up reverting the commit to keep this pr minimal and save for a more thorough discussion

hadley:
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks!

PR #612, opened 2026-05-22
Upkeep 2026-05

hadley:


t-kalinowski (commented):
I suggest we move these files, so other agent harnesses can find them too:

- `.claude/skills` -> `.agents/skills`
- `.claude/CLAUDE.md` -> `AGENTS.md`

Also, my suggestion would be to move the changes from `air format .` to another PR, or merge this PR without squashing, so that, and then we can add a `.git-blame-ignore-revs` and keep `git blame` useful.

hadley:
The `claude.md` is from `usethis::use_claude_code()`, so I've made a note to fix it upstream. I'll split the air formatting into another PR.

PR #617, opened 2026-05-22
Clarify that `method()` only works with S7 generics

### LINKED ISSUE #568: `S7::method` does not accept an S3 generic while the doc suggest it should.
In R, with S7 v0.2.0:

```r
> isS3stdGeneric(print)
print 
 TRUE 
> S7::method(print)
Error: `generic` must be a <S7_generic>, not a <closure>
```

The doc says:
```
Usage:

     method(generic, class = NULL, object = NULL)
     
Arguments:

 generic: A generic function, i.e. an S7 generic, an external generic,
          an S3 generic, or an S4 generic.
```

hadley:
Fixes #568

### DIFF
--- a/R/method-introspect.R
+++ b/R/method-introspect.R
@@ -7,9 +7,11 @@
 #' the implementation of a specific method.
 #'
 #' @seealso [method_explain()] to explain why a specific method was picked.
-#' @inheritParams method<-
 #' @returns Either a function with class `S7_method` or an error if no
 #'   matching method is found.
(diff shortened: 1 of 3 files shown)


hadley:
@t-kalinowski it took me a while to work through this. Can you please confirm my logic is correct by reading through my comments on #568?

t-kalinowski (approved):
Makes sense to me. I posted a comment on #568.

hadley:
Oooh, good point!

PR #620, opened 2026-05-22
Class/generic/methods introspection

### LINKED ISSUE #335: Functions to list classes/generics in a package
And to list all S7 methods for a S3/S4 generic.

Extracted from #333. 

cc @mmaechler do you have any additional thoughts about what sort of API you'd like here?

### LINKED ISSUE #435: Missing (equivalent of) S3 / S4   `methods(class = .)`  to get all generics for which there are <class>-methods
I was just mentioning  `methods(class = class(.))`  in an [e-mail advice on R-help](https://stat.ethz.ch/pipermail/r-help/2024-September/479917.html).
I've advertized this in my intermediate-to-advanced R teaching as a way to help users get familiar with what they can do some
(in statistics) model fit or similar.
Note that `methods(class=*)` is even not at all perfect (as it *only* finds methods for the exact class signature, but not those from parent classes which also apply).

When S7 becomes "main stream" we'd be happy to instrument  `utils::methods()` to  also work for S7 classes, but that's definitely for later,  and I think we should provide the functionality within S7.

hadley:
Fixes #335. Fixes #435

For the first round of review I think it would be most useful to get feedback on the interface, especially the data frame that `S7_methods()` returns.


R/introspect.R
```diff
@@ -0,0 +1,155 @@
...
+#' S7_methods(generic = my_gen)
+#' S7_methods(class = Foo)
+S7_methods <- function(generic = NULL, class = NULL) {
+  if (!is.null(generic)) {
+    if (!is_S7_generic(generic)) {
+      stop("`generic` must be an S7 generic.")
+    }
+    generics <- list(list(generic = generic, package = NA_character_))
+  } else {
+    generics <- attached_generics()
```
t-kalinowski, line 60: Should we also discover generics in loaded but not attached namespaces?

hadley: My gut feeling is no? But that's a not particularly strongly held opinion.

lawremi: Agree with @hadley.

lawremi (approved):
The returned `data.frame` makes sense to me; it's what I had imagined, as well.

PR #621, opened 2026-05-22
Ensure dispatch can occur on wrapped missing arguments

hadley:
Fixes #595


src/method-dispatch.c
```diff
@@ -188,6 +188,8 @@ SEXP method_call_(SEXP call_, SEXP op_, SEXP args_, SEXP env_) {
   PROTECT_WITH_INDEX(R_NilValue, &arg_pi); // unnecessary, for rchk only
   PROTECT_WITH_INDEX(R_NilValue, &val_pi); // unnecessary, for rchk only
 
+  SEXP missing_call = PROTECT(Rf_lang2(fn_base_missing, R_NilValue));
```
hadley: Copied from `generic_args()` above.

t-kalinowski (approved):

src/method-dispatch.c
t-kalinowski: We should probably define the call object as a package `static`, and then initialize it just once in `R_init_S7`.

Issue #494, opened 2024-11-12
Dims in str for data.frame-inheriting object


jonthegeek:
This line is causing me issues:
https://github.com/RConsortium/S7/blob/f5b23ab53534f041c16f9776f51e74f03522c307/R/class.R#L305

I have a subclass of data.frame, ["paths"](https://github.com/jonthegeek/rapid/blob/main/R/paths.R). This is the hackiest part of my package and likely to change, but the tests work before and now throw an error, so I want to report this while I see it.

When my test tries to print an object that contains a "rapid::paths" object (and thus calls `str()`), I get an error. You can reproduce it like this:

``` r
library(S7)
df <- data.frame(a = 1:2, b = 1:2)
# Imagine that df belongs to an S7 subclass of data.frame.
class(df) <- c("myclass", "data.frame", "S7_object")
str(df)
#> Classes 'myclass', 'S7_object' and 'data.frame': 2 obs. of  2 variables:
#> <myclass>
#> Error in attributes(object) <- list(names = names(object), dim = dim(object)): dims [product 4] do not match the length of object [2]
```

<sup>Created on 2024-11-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Or more directly like this:

``` r
x <- data.frame(a = 1:2, b = 1:2)
attributes(x) <- list(dim = dim(x))
```

I suspect I can avoid this by implementing a `str` method like I've been meaning to do anyway (not to mention that this object likely won't inherit from data.frame much longer), but it also seems like this is probably a bug.

hadley:
Hmmmmmmm, I think if #380 was fixed then we could have:

```R
str.S7_object <- function(object, ..., nest.lev = 0) {
  cat(if (nest.lev > 0) " ")
  cat(obj_desc(object))

  if (!is_S7_type(object)) {
    if (!typeof(object) %in% c("numeric", "integer", "character", "double")) {
      cat(" ")
    }

    str(S7_data(object), nest.lev = nest.lev)
  } else {
    cat("\n")
  }

  str_nest(props(object), "@", ..., nest.lev = nest.lev)
}
```

Issue #380, opened 2023-11-22
Should S7 data return S3 objects?


hadley:
e.g. 

``` r
library(S7)

mydf <- new_class("mydf", class_data.frame)
S7_data(mydf(data.frame(x = 1, y = 2)))
#> $x
#> [1] 1
#> 
#> $y
#> [1] 2
#> 
#> attr(,"row.names")
#> [1] 1
```

<sup>Created on 2023-11-22 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

Shouldn't that return a data frame?

PR #622, opened 2026-05-22
Re-allow environments as parents

### LINKED ISSUE #590: Allow `class_environment` as parent.
Currently `class_environment` as parent is disallowed by: #290 to fix the issue: #253 . 

But environments are really useful objects in R. As they are the only object that support in place updating. It would be nice if we can have this type of object participate in the S7 dispatch system and define validation rules on environments.

I opened a similar issue before where I requested the feature of allowing in-place updates for S7 properties: https://github.com/RConsortium/S7/issues/525. But I can understand if that feature gets rejected duo complexity. Though, I still think it is valid to resurrect the possibility to create a S7 environment object directly. And hence I am creating this separate issue to distance it from #525. 

### The problem with environments:

As explained in the issue #253, the root cause with environments is when calling `S7_data()` a function a [zap_attr](https://github.com/RConsortium/S7/blob/bd70ec391abeb37ed560b16d89b40a3439b1e1dd/R/data.R#L40) removes (among others) the attributes: "class" and "S7_class" . Eg: `attr(my_s7_env_obj, 'class') <- NULL`. 

But duo the unique nature of environments, this attribute property update is done in place and thus the S7 nature is lost permanently.

### How could we fix it?

1. Special case `class_environment`. Since environments in R are the ONLY base objects that exhibit this behavior we could simply special case it in our logic. There are then still different options how we want to deal with the environment when S7_data is called.
   a. Disallow it for environments. In a sense we move the "disallowing" of it being a parent to just disallowing it being called on S7_data(). 
   b. Allow it but document it that it will destroy the S7 nature and this it a one way trip. One is still free to re instantiate the "S7" object by calling the constructor again with the backed-up S7 properties (provided the user did back-up the properties this would also need to get documented).
   c. Allow it but document that a COPY is returned. Though I am not sure if this would be useful because the cool thing about environments is that they support in place updating. So if you then do further manipulations on this copy one needs to remember to update the original S7 object also. But maybe there are use cases where it can be handy. 
2. Rework S7_data and zap_attr mechanism that would support environments. A refactor of this code is needed anyway if want to solve: https://github.com/RConsortium/S7/issues/380 . Maybe the rework could also take environments into account. 
   - Though I am not sure how such a refactor could look like, duo the fact that there is only 1 copy of an environment and it either has the S7 classes or not.

The above suggestions do assume S7_data is the only place where environments give trouble with the S7 system. It might be other places in the code base do not assume in place attribute updating. But again here I would argue to maybe special case environments. And just document upfront that if you are going to inherit from environments be aware that it behaves differently then the other base objects. 

Just as users should be aware in normal R programming that environments are different from the copy-on-write principle that other objects exhibit. So S7 would then just extend that nature of R to its system.

----

Personally, I would not call the `S7_data` function as I would just interact with the environment via `$` and `[`, `[[` methods on the S7 environment object directly. So whatever gets chosen for how to deal with `S7_data` it would unblock me.

hadley:
Now forbidding the dangerous `S7_data()` instead.

Fixes #590


hadley:
Also need to double check `is.environment()` path in `str()` 

...

I think the combination resolution of #380  + #494 implies we should provide a `str.S7_environment` that either errors or works differently. Now in #641.

...

But also need to think about `S7_data<-` and need to check it as a sub-subclass since `new_object()` also edits attributes (but in a way I think is ok).

t-kalinowski (approved):

hadley:
Need to consider attribute handling in convert too

PR #624, opened 2026-05-22
Allow a property `setter` to (optionally) take `name`

### LINKED ISSUE #552: Feature request: pass `name` to property setter
I've found it useful to declare properties outside a class definition and recycle the property definition for various classes or for similar properties within one class.
The issue I'm running in to is that property setters don't have access to the name of the property that is being set.
In the example below, I cannot recycle the colour property for `fill`, because the setter function hardcodes the name.

``` r
library(S7)

property_colour <- new_property(
  class = class_character,
   # Must I hardcode the property name?
    prop(self, "colour") <- as.character(value)
    self
  }
)

rectangle <- new_class(
  "rectangle",
  properties = list(
    colour = property_colour, 
    fill = property_colour
  )
)

x <- rectangle("red", "blue")
#> Error: <rectangle> object properties are invalid:
#> - @fill must be <character>, not <NULL>
```

<sup>Created on 2025-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

My current solution to this problem is to make a property factory like so:

``` r
library(S7)

property_colour <- function(name) {
  force(name)
  new_property(
    class = class_character,
      prop(self, name) <- as.character(value)
      self
    }
  )
}

rectangle <- new_class(
  "rectangle",
  properties = list(
    colour = property_colour("colour"), 
    fill = property_colour("fill")
  )
)

x <- rectangle("red", "blue")
```

<sup>Created on 2025-07-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

But this solution is a bit unwieldy because I still have to declare the `"colour"` and `"fill"` names twice: once in the list names and once as argument to `property_colour()`. If we could somehow automatically access the property name that is being set from the context of the setter function, this would be a much smoother experience.

hadley:
Fixes #552


R/property.R
hadley: Is it still worthwhile to maintain this copy of the actual C implementation?

hadley:
@t-kalinowski there's relatively little to review for the actual implementation — it's mostly the infrastructure needed to make it work, i.e. a more flexible `check_function()` (now with tests) and some C level changes.

t-kalinowski (commented):
I think we'll want to revisit this once #638 or #627 lands, then rebase and update this PR as appropriate.

Depending on how the traceback improvement fixes lands, we might also be able to make the name accessible through other ways too, without expanding the signature. E.g., deriving it from `sys.call()[[1]]` or `last(attr(object, "...."))` (probably behind a helper). 

But in the abstract, I'm 👍🏻 on expanding the signature like this.

R/property.R
t-kalinowski: I think we can delete it.

hadley: https://github.com/RConsortium/S7/issues/647

hadley:
I'd definitely prefer to make it explicit in the argument list.

t-kalinowski (approved):
Should we add similar support for custom getters too?

hadley:
Maybe? But it seems somehow less compelling. I think we can wait until someone asks for it and provides a compelling use case.

PR #626, opened 2026-05-22
Optionally filter `prop_names()`

### LINKED ISSUE #551: Allow `prop_names()` to enumerate readable/writeable properties
Maybe with an enum? e.g. `prop_names(object, "read-only")`, `prop_names(object, "write-only")` etc?

hadley:
Not sure if this is the right way to factor the split.

Fixes #551


hadley:
I don't like it; the name options seem too arbitrary.

PR #627, opened 2026-05-23
Wrap property getting and setting errors

### LINKED ISSUE #536: Including function call in error messages inside setter
Dear S7 team,
I would like to include function calls when throwing errors from inside a setter function. Due to the complexity of the call stack, `rlang::caller_env()` does not work (see below). What would be the best way to include the function call? Thanks! 
(Currently, my best idea is to use `rlang::trace_back()` and craft the call based on its first entry.)

```r
library(S7)
library(rlang)
library(cli)

testclass <- new_class("testclass",
  properties = list(
    pro = new_property(
      class = class_character,
      validator = function(value) {
        if (identical(value, "hello")) {
          "must not be hello"
        }
      },
        if (identical(value, "hello")) {
          cli_abort("Can't do that.",
                    call = caller_env())
        }
        
        self@pro <- value
        self
      }
    )
  )
)

t <- testclass(pro = "hello")
#> Error:
#> ! Can't do that.
```

hadley:
This is a significant improvement in messaging, but this implementation might not be worth it because of the performance cost (getting a property currently takes ~2µs on my machine)

```R
x <- new_class("foo")()
f <- function(x) 1

bench::mark(
  f(),
  withCallingHandlers(f()),
  with_prop_call(x, "foo", f())
)[1:4]

#> # A tibble: 3 × 4
#>   expression                             min   median `itr/sec`
#>   <bch:expr>                        <bch:tm> <bch:tm>     <dbl>
#> 1 "f()"                                    0    123ns  8495343.
#> 2 "withCallingHandlers(f())"           328ns    533ns  1803805.
#> 3 "with_prop_call(x, \"foo\", f())"    533ns    697ns  1333597.
```

Nevertheless, it might suggest a faster approach.

Fixes #536

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # S7 (development version)
 
+* Errors raised by `prop()` and `prop<-()` (including from custom getters and setters) now report a synthetic `<Class>@<prop>` call site, making it obvious which property triggered the error (#536).
 * `new_object()` no longer materialises ALTREP parent values (e.g. `seq_len()`), so constructing an S7 object that wraps a large compact integer sequence is now O(1) in memory instead of O(n) (@kschaubroeck, #607).
 * Internal changes to support R-devel (4.6) (#592, #593, #598, #600).
 * `S7_error_method_not_found` now has a correct class vector without a duplicate `"error"` entry (@jjjermiah, #604)
--- a/R/property.R
+++ b/R/property.R
@@ -194,7 +194,7 @@ prop_default <- function(prop, envir, package) {
 #' lexington@height <- 14
 #' prop(lexington, "height") <- 15
 prop <- function(object, name) {
-  .Call(prop_, object, name)
+  with_prop_call(object, name, .Call(prop_, object, name))
 }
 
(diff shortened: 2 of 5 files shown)


hadley:
@t-kalinowski would be good to get your thoughts here. My sense is we don't want to give up on your hard work to make prop getting and setting performant, but this at least demonstrates where we'd see improvements.

t-kalinowski:
I sketched out an alternative approach in https://github.com/RConsortium/S7/pull/638

t-kalinowski (commented):
It'd be nice if we could avoid taking the overhead penalty of `withCallingHandlers()` on every invocation of `prop()` and `` `prop<-`() ``, including ones that never invoke a custom getter or setter and also ones that never error.

I sketched out an alternative approach based on `R_UnwindProtect` in https://github.com/RConsortium/S7/pull/638 that preserves the fast happy path and is near parity with the current benchmarks.

t-kalinowski:
Superceded by https://github.com/RConsortium/S7/pull/638

PR #628, opened 2026-05-25
Implement `prop_info()`

### LINKED ISSUE #551: Allow `prop_names()` to enumerate readable/writeable properties

hadley:
I tried implement a filtering mechanism on top of `prop_names()` but it felt too limiting and too arbitrary. Instead, I provide an easy way to get all metadata about the properties.

Fixes #551


R/property.R
```diff
@@ -356,20 +356,47 @@ prop_label <- function(object, name) {
...
+#' @param object Either an S7 object (an instance) or an S7 class.
+#' @returns
+#' * `prop_names()` returns a character vector
+#' * `prop_exists()` returns a single `TRUE` or `FALSE`.
+#' * `prop_info()` returns a  data frame with one row per property and the
+#'   following columns:
+#'   * `name`: a character vector of property names.
+#'   * `default`: a list column of property defaults.
+#'   * `class`: a character description of the class.
+#'   * `getter`, `setter`, `validator`: logical vectors indicating whether
```
t-kalinowski, line 376: What do you think about returning a list column of the closures instead of a logical vector?

t-kalinowski (commented):
This is a good addition, and we should add it. 

But it doesn't really address the original motivation that sparked this PR: an easy way to tell if a property is read-only. I wonder if we should pre-populate the dataframe with some additional columns like `is_readonly`.

R/property.R
hadley: That feels too clever 😄

hadley:
Yeah, but it gives me an easy way to compute that. I think for the motivating problem, IIRC, this is better than what I proposed in the issue.

hadley:
I've merged as is, but we can easily add more columns if we discover their utility in the future.

PR #619, opened 2026-05-22
Support class specifications in `S7_inherits()`

hadley:
Fixes #556

PR #630, opened 2026-05-25
Improve consistency of errors

hadley:
* Make them sentences (i.e. they end in `.`)
* Use `call. = TRUE` iff called from a non-exported helper. The idea is to only show the call if it will be useful to the user, i.e. they can read the docs for it.
* Spread the `as_class()` error over multiple lines

@t-kalinowski it should be adequate to skim through the snapshot changes.


t-kalinowski (approved):

R/class-spec.R
```diff
@@ -31,11 +31,18 @@ as_class <- function(x, arg = deparse(substitute(x))) {
...
+        obj_desc(x)
+      ),
+      " * An S7 class object",
+      " * The result of `new_S3_class()`",
+      " * An S4 class object",
+      " * A base class"
     )
-    stop(paste0(error_base, msg), call. = FALSE)
+
+    stop(paste0(c(error_base, msg), collapse = "\n"), call. = FALSE)
```
hadley: Should be `call. = TRUE` since `as_class` is user facing

t-kalinowski: 👍🏻. (Related discussion in https://github.com/RConsortium/S7/pull/619#discussion_r3300150672 and #646.)

PR #632, opened 2026-05-25
Improvements to packages vignette

### LINKED ISSUE #325: Document Need to Define Classes in Order
This is somewhat discussed in #250 (in particular [here](https://github.com/RConsortium/OOP-WG/issues/250#issuecomment-1229131166)), but before fully solving that issue it feels like it would be helpful to point out that a class needs to be defined earlier in a script or package if you want to use it as a property in another class. This is particularly confusing in a package context, so perhaps add it to the packages vignette?

``` r
# Causes the package to fail to build.
s7_parent <- S7::new_class(
  "s7_parent",
  package = "newS7fanciness",
  properties = list(name = s7_child)
)
#> Error in eval(expr, envir, enclos): object 's7_child' not found

s7_child <- S7::new_class(
  "s7_child",
  package = "newS7fanciness",
  properties = list(x = S7::class_character)
)
```

<sup>Created on 2023-08-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

``` r
# Works
s7_child <- S7::new_class(
  "s7_child",
  package = "newS7fanciness",
  properties = list(x = S7::class_character)
)

s7_parent <- S7::new_class(
  "s7_parent",
  package = "newS7fanciness",
  properties = list(name = s7_child)
)
```

<sup>Created on 2023-08-25 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #453: doc: is exporting a class the same as the constructor
Related https://github.com/RConsortium/S7/issues/348

The [package website says](https://rconsortium.github.io/S7/articles/packages.html#documentation-and-exports): 

> "If you export a class, you must also set the package argument, ensuring that classes with the same name are disambiguated across packages."

What I am not clear of here is if exporting a class is the same as exporting the constructor. Is this the case? 

I think a super simple code example with `#' @export` in this section would do wonders to clarify that. 

Additionally, in this context, is there a naming convention we should follow for the class constructors? (I could make a new issue for this if desired)

### LINKED ISSUE #501: Can't register S3 generic for S7 classes when using package namespacing in S7 class name
When using S7 classes in a package, there are times when I want to be able to register an S3 generic for an S7 class. When not using package namespacing in the S7 class name (`new_class(..., package = NULL, ...)`), this works as expected.

For example, here I'm registering a print method:

``` r
range <- S7::new_class("range",
  package = NULL,
  properties = list(
    start = S7::class_double,
    end = S7::class_double
  )
)

#' @export
print.range <- function(x, ...) {
  message("range(", x@start, ", ", x@end, ")")
}

x <- range(start = 1, end = 10)

class(x)
#> [1] "range"     "S7_object"

x
#> range(1, 10)
```

However, I don't know how to do this when using the package namespacing:

``` r
range <- S7::new_class("range",
  package = "mypackage", # to simulate this being defined in a package called "mypackage"
  properties = list(
    start = S7::class_double,
    end = S7::class_double
  )
)

x <- range(start = 1, end = 10)

class(x)
#> [1] "mypackage::range" "S7_object"
```

I can't define an s3 method `print.mypackage::range <- function(x, ...) {}` - R is not a fan of the `::`.

Is there another way to define the method? If so, it would be good to document [here](https://rconsortium.github.io/S7/articles/compatibility.html#methods). If not, perhaps the way the namespaced class name gets created should be reconsidered.

### LINKED ISSUE #508: S7 naming convetions
Related to https://github.com/RConsortium/S7/issues/453

I am working on implementing S7 classes for [arcgisutils](https://github.com/R-ArcGIS/arcgisutils/) and one thing I'm not quite sure on is how to handle the naming convention for S7 classes. 

S7 exports classes with the prefix `class_`. Is it recommended that R packages that export S7 classes export it with the struct `class_{class_name}`?

### LINKED ISSUE #530: Error: object 'S7' not found, when registering a new method for S7::convert
Dear developers,

Within my package I want to create a new method for  S7::convert generic to convert the `Foo1` into a `S7::class_data.frame`. I used the following:
```R
Foo1 <- S7::new_class("Foo1", properties = list(x = S7::class_data.frame))

S7::method(S7::convert, list(Foo1, S7::class_data.frame)) <- \(from, to) from@x
```

Unfortunately this results in error:
```
Error: object 'S7' not found
```

## Workaround 
I had to use `@importFrom S7 convert` then use it like this:
```
S7::method(convert, list(Foo1, S7::class_data.frame)) <- \(from, to) from@x
```

## Question
I was wondering whether this is the best way of doing it and thanks for your help.

hadley:
* Recommend importing all S7 functions. Fixes #530.
* Describe how to manually register an S3 method. Fixes #501.
* Discuss class hierarchies and file order. Fixes #325
* Clarify that class = constructor. Fixes #453.
* Recommend CamelCase for class names. Fixes #508 (This is not in the packages vignette, but it feels like it hangs with the other changes)

Also lightly proofread the whole doc.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -7,9 +7,12 @@
 #'
 #' Learn more in `vignette("classes-objects")`
 #'
-#' @param name The name of the class, as a string. The result of calling
-#'   `new_class()` should always be assigned to a variable with this name,
-#'   i.e. `Foo <- new_class("Foo")`.
+#' @param name The name of the class, as a string. (We recommend using
+#'   CamelCase for S7 class names, but it is not required.)
+#'
+#'   The result of calling `new_class()` should always be assigned to a variable
+#'   with this name, i.e. `Foo <- new_class("Foo")`. This object both represents
+#'   the class and is used to construct new instances of the class.
 #' @param parent The parent class to inherit behavior from.
 #'   There are three options:
 #'
--- a/man/new_class.Rd
+++ b/man/new_class.Rd
@@ -20,7 +20,9 @@ new_object(.parent, ...)
 \arguments{
 \item{name}{The name of the class, as a string. The result of calling
 \code{new_class()} should always be assigned to a variable with this name,
-i.e. \code{Foo <- new_class("Foo")}.}
+i.e. \code{Foo <- new_class("Foo")}.
+
+We recommend using CamelCase for S7 class names.}
 
 \item{parent}{The parent class to inherit behavior from.
 There are three options:
--- a/vignettes/packages.Rmd
+++ b/vignettes/packages.Rmd
@@ -15,16 +15,18 @@ knitr::opts_chunk$set(
 ```
 
 This vignette outlines the most important things you need to know about using S7 in a package.
-S7 is new, so few people have used it in a package yet; this means that this vignette is likely incomplete, and we'd love your help to make it better.
+S7 is new, so few people have used it in a package yet; this vignette is likely incomplete, and we'd love your help to make it better.
 Please [let us know](https://github.com/RConsortium/S7/issues/new) if you have questions that this vignette doesn't answer.
 
 ```{r setup}
 library(S7)
 ```
 
-## Method registration
+## Getting started
 
-You should always call `methods_register()` in your `.onLoad()`:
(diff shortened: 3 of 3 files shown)


t-kalinowski (approved):
LGTM!

PR #634, opened 2026-05-25
Show property defaults when printing a class

### LINKED ISSUE #439: Update `S7_class` `print()` method to show default property values
The current printout does not show the default values for properties. The default could be shown in the constructor or next to the property. Displaying it next to the property might make more sense; otherwise, constructor printouts will get long, and we'd need to implement line breaks in the constructor printout.

``` r
library(S7)


Person <- new_class(
  name = "Person",
  properties = list(
    # non-dynamic, default missing (required constructor arg)
    first_name = new_property(class_character, default = quote(expr = )),
    
    # non-dynamic, static default (optional constructor arg)
    middle_name = new_property(class_character, default = ""),
    
    # non-dynamic, default missing (required constructor arg) (same as first_name)
    last_name = new_property(class_missing | class_character),
    
    # non-dynamic, but defaults to the value of another property
    nick_name = new_property(class_character, default = quote(first_name)),
    
    # non-dynamic, optional constructor argument, read-only after construction.
    birthdate = new_property(
      class = class_Date,
      default = quote(Sys.Date()),
        if (!is.null(self@birthdate))
          stop("Can't set read-only property Person@birthdate")
        self@birthdate <- value
        self
      }
    ),
    
    # dynamic property, not a constructor argument
    age = new_property(class = class_any, getter = \(self) {
      Sys.Date() - self@birthdate
    })
  )
)
```
```r
Person
#> <Person> class
#> @ parent     : <S7_object>
#> @ constructor: function(first_name, middle_name, last_name, nick_name, birthdate) {...}
#> @ validator  : <NULL>
#> @ properties :
#>  $ first_name : <character>             
#>  $ middle_name: <character>             
#>  $ last_name  : <MISSING> or <character>
#>  $ nick_name  : <character>             
#>  $ birthdate  : S3<Date>                
#>  $ age        : <ANY>
```

hadley:
Fixes #439


tests/testthat/_snaps/class.md
```diff
@@ -52,6 +52,21 @@
...
+    Output
+      <Person> class
+      @ parent     : <S7_object>
+      @ constructor: function(implicit_default, default_value, default_expr) {...}
+      @ validator  : <NULL>
+      @ properties :
+       $ implicit_default (<character>)
+       $ default_value = "" (<character>)
+       $ default_expr = Sys.Date() (S3<Date>)
+       $ read_only <read-only> (<ANY>)
```
hadley: Maybe `[read-only]` since it's not a class?

tests/testthat/_snaps/class.md
```diff
@@ -52,6 +52,21 @@
...
+
+    Code
+      Person
+    Output
+      <Person> class
+      @ parent     : <S7_object>
+      @ constructor: function(implicit_default, default_value, default_expr) {...}
+      @ validator  : <NULL>
+      @ properties :
+       $ implicit_default (<character>)
```
hadley: Should we figure out what the implicit value is and show it?

tests/testthat/_snaps/class.md
```diff
@@ -52,6 +52,21 @@
...
+       $ default_value = "" (<character>)
+       $ default_expr = Sys.Date() (S3<Date>)
+       $ read_only <read-only> (<ANY>)
+
```
hadley, line 71: Need to add a test case where the property is another S7 class

hadley:
@t-kalinowski I don't think this is done, but it's at a point where it would be useful for you to react to. In particular, the challenge is default values for thinks like `class_factor` and `class_Date`, and generally how far we want to try to go to figure them out.

tests/testthat/_snaps/property.md
```diff
@@ -110,13 +110,13 @@
...
-       $ anything: <ANY>                 
-       $ null    : <NULL>                
-       $ base    : <integer>             
-       $ S3      : S3<factor>            
-       $ S4      : S4<class_S4>          
-       $ S7      : <class_S7>            
-       $ S7_union: <integer> or <logical>
+       $ anything (<ANY>)
+       $ null (<NULL>)
+       $ base = integer(0) (<integer>)
```
t-kalinowski: What do you think about using Rust / Python syntax for printing, like:

```
       $ anything: <ANY> 
       $ null: <NULL> = NULL
       $ base: <integer> = integer(0)
       $ S3:  S3<factor> = factor()
```

t-kalinowski (commented):
This will be a nice addition!

tests/testthat/_snaps/property.md
hadley: I like it!

t-kalinowski (approved):

Issue #573, opened 2025-10-14
Defining generic methods for suggested packages


jtlandis:
Hello everyone,

When building a package that is exporting a generic function, what is the proper way to export a method for a class within a "Suggests" package?

For example, if package `foo` exports the `foo()` generic, but lists package `bar` as a Suggests, what would be the proper way to register methods for classes within `bar`?

Would it be something to the effect of:

```r
# from the foo package
.onLoad <- function(lib, pkg) {

  if (isNamespaceLoaded("bar")) {
    method(foo, bar::class_bar) <- function(x) {cat("we dispatched on `bar::class_bar`")}
  } else {
    setHook(packageEvent("bar", "onLoad"), function(){
      method(foo, bar::class_bar) <- function(x) {cat("we dispatched on `bar::class_bar`")}
    })
  }
}

```

teunbrand:
I think the advice for external generics is written up here: https://rconsortium.github.io/S7/reference/new_external_generic.html

jtlandis:
FWIW, This issue was inspired by the pattern from `S7::methods_register` which also evaluates `setHook` in the FALSE branch:

```r
> S7::methods_register
function ()
{
    package <- packageName(parent.frame())
    ns <- topenv(parent.frame())
    tbl <- S7_methods_table(package)
    for (x in tbl) {
        register <- registrar(x$generic, x$signature, x$method,
            ns)
        if (isNamespaceLoaded(x$generic$package)) {
            register()
        }
        else {
            setHook(packageEvent(x$generic$package, "onLoad"),
                register)
        }
    }
    invisible()
}
```

hadley:
We should fix that.

PR #635, opened 2026-05-25
Unconditionally register a on-load hook

hadley:
This handles the (admittedly rare) case where a user unloads then re-loads a package.

Spotted in #573

### DIFF
--- a/R/external-generic.R
+++ b/R/external-generic.R
@@ -103,9 +103,8 @@ methods_register <- function() {
 
     if (isNamespaceLoaded(x$generic$package)) {
       register()
-    } else {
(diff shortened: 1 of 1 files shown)


t-kalinowski (commented):
I'm surprised we need this. If a package is reloading, isn't its `.onLoad()` hook running again already?

hadley:
Right, but we're currently only setting that if the package isn't loaded.  But I know we learned this lesson the hard way in rlang: https://github.com/r-lib/rlang/blob/b2e9594bdfc539c65370eb6be6208cc8637d356f/R/standalone-s3-register.R#L120

PR #639, opened 2026-05-25
Make it possible to unregister a method

### LINKED ISSUE #613: Provide way to unregister methods
e.g. @mmaechler's suggestion:

```R
method(<gen>, <signature>) <-  NULL
```

hadley:
I'm now more sceptical about this approach because S3 doesn't provide any way to unregister a method, and S4's `removeMethod()` explicitly advises against using it. ... But it is undeniably useful for testing.

Fixes #613


hadley:
@t-kalinowski more interested in whether you thing we should do this than the actual implementation.

t-kalinowski (approved):
After some thought, I think this is worth merging. It works cleanly for S7, and the S3 restriction is very reasonable.

It does give me pause that it's a feature that can be abused easily, but it's not like this addition is opening some monkey-patching Pandora's box. It's ergonomic syntax, worth it for the symmetry with `$<- NULL` semantics alone, and there are real-life scenarios where it is genuinely useful, like tests.

PR #640, opened 2026-05-25
Use `UnwindProtect` to clean up on getter error

### LINKED ISSUE #520: Getter and Setter are not called anymore once they crash.
I would like to implement an "initialization"-pattern with a S7 class. That is certain properties can only be accessed once the object is flagged as initialized (eg. all necessary resources have been allocated).

In the getter of the property I thus check if the object is initialized and raise an error if this is not the case. However, the getter seems to only run once and then afterwards the check is simply by-passed. _I expected the getter to be called everytime._

The property in question also needs to be settable as it represent a resource that got allocated on initialization and thus needs to be assignable. 

The following code highlights my problem:
```R
library(S7)
Test <- new_class(
  "Test",
  properties = list(
    init = class_logical,
    conn = new_property(
      class = class_numeric, # In reality a connection object.
      setter = \(self, value) {
        self@conn <- value 
        self
      },
      getter = \(self) {
        if (!self@init) stop("Uninitialized")
        return(self@conn)
      }
    )
  )
)

t <- Test(init = FALSE)
t
# <Test>
# Error in (function (self)  : Uninitialized
t
# <Test>
#  @ init: logi FALSE
#  @ conn: int(0)
t@conn <- 2
t@conn
# [1] 2
t@init
# [1] FALSE
```

I know that the end-user could also perform a check `if(!t@init)`. But then the check in the getter becomes pointless in the first place. I would like the class to self contained and guard against wrong usage of uninitialized properties.

hadley:
I also added a test for an erroring setter, but as claude points out, the setter path doesn't have the same bug because it shallow-duplicates the object before mutating `.setting_prop`, so the mutated copy is discarded on error.

Fixes #520

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,7 @@
 # S7 (development version)
 
 * `new_object()` no longer materialises ALTREP parent values (e.g. `seq_len()`), so constructing an S7 object that wraps a large compact integer sequence is now O(1) in memory instead of O(n) (@kschaubroeck, #607).
+* `prop()` (and `@`) no longer leaves an object in a broken state when a custom getter signals an error (#520).
 * Internal changes to support R-devel (4.6) (#592, #593, #598, #600).
 * `S7_error_method_not_found` now has a correct class vector without a duplicate `"error"` entry (@jjjermiah, #604)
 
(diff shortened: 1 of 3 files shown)


t-kalinowski:
Superceded by https://github.com/RConsortium/S7/pull/638

PR #641, opened 2026-05-25
Empower `S7_data()` for S3 objects then use it in `str()`

### LINKED ISSUE #380: Should S7 data return S3 objects?

### LINKED ISSUE #494: Dims in str for data.frame-inheriting object

hadley:
I've implemented these together because (a) they are simple but, more importantly, (b) the fix in `str()` directly inspires the new behaviour in `S7_data()`.

Fixes #380. Fixes #494.

### DIFF
--- a/tests/testthat/_snaps/class.md
+++ b/tests/testthat/_snaps/class.md
@@ -202,6 +202,16 @@
       List of 1
        $ : <text> chr "x"
 
+# S7 object / displays data.frame subclasses without error (#494)
+
+    Code
+      str(mydf(data.frame(a = 1:2, b = 1:2)))
+    Output
+      Classes 'mydf', 'S7_object' and 'data.frame':	2 obs. of  2 variables:
+      <mydf> 'data.frame':	2 obs. of  2 variables:
+       $ a: int  1 2
+       $ b: int  1 2
+
 # S7 object / displays list objects nicely
 
     Code
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,9 @@
 
 * `new_object()` no longer materialises ALTREP parent values (e.g. `seq_len()`), so constructing an S7 object that wraps a large compact integer sequence is now O(1) in memory instead of O(n) (@kschaubroeck, #607).
 * Internal changes to support R-devel (4.6) (#592, #593, #598, #600).
+* `S7_data()` now preserves the S3 class when the S7 class inherits from an S3 class, so e.g. `S7_data()` on a data.frame subclass now returns a data.frame (#380).
 * `S7_error_method_not_found` now has a correct class vector without a duplicate `"error"` entry (@jjjermiah, #604)
+* `str()` on S7 objects that inherit from data.frame (or other S3 classes whose underlying data has a `dim` attribute incompatible with the bare base type) no longer errors (#494).
 
 # S7 0.2.2
 
--- a/R/class.R
+++ b/R/class.R
@@ -322,15 +322,7 @@ str.S7_object <- function(object, ..., nest.lev = 0) {
       cat(" ")
     }
 
-    attrs <- attributes(object)
-    if (is.environment(object)) {
-      attributes(object) <- NULL
-    } else {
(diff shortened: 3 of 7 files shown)


jonthegeek (approved):
I confirmed that this fixes the underlying issue that caused me to submit #494.

This is wild timing. I submitted that 1.5 years ago with the intention of making it irrelevant to me "soon"... and intended to finally get to that "soon" step *today*. Thanks for fixing this so I don't have to work around it!

tests/testthat/_snaps/class.md
```diff
@@ -202,6 +202,16 @@
       List of 1
        $ : <text> chr "x"
 
+# S7 object / displays data.frame subclasses without error (#494)
+
+    Code
+      str(mydf(data.frame(a = 1:2, b = 1:2)))
+    Output
+      Classes 'mydf', 'S7_object' and 'data.frame':	2 obs. of  2 variables:
```
t-kalinowski, line 210: The duplication here of `mydf ... 'data.frame':	2 obs. of  2 variables:` was confusing to me, but after tracking it down I see it's not coming directly from S7 code. 

`str.data.frame` is invoked first (since `data.frame` appears before `S7_object` in the class list). Then that calls `NextMethod("str", ...)`, which invokes the `str.S7_object` method. Then we call `str(S7_data(object))`, which invokes `str.data.frame` again.

While unfortunate, after thinking about it a little, I think this is the least worst option. I briefly explored an alternative locally, avoiding re-invoking an already invoked `str` S3 method, but it wasn't significantly better output and I don't think the complexity is worth it.

```r
    data <- S7_data(object)
    classes <- attr(data, "class", TRUE)
    already_invoked <- vlapply(classes, \(cls) {
      !is.null(getS3method(
        "str",
        cls,
        optional = TRUE,
        envir = environment(str)
      ))
    })
    class(data) <- classes[!already_invoked]
    str(data, nest.lev = nest.lev)
```

t-kalinowski (approved):

tests/testthat/_snaps/class.md
hadley: Yeah, at some point we should do a deep dive on `str()` methods as they're surprisingly hard to get right.

PR #642, opened 2026-05-25
Fix typo in S4 method registration

hadley:


hadley:
Subsumed into bigger PR

Issue #455, opened 2024-10-03
Use `method<-` to register methods on S3 generics for base classes


t-kalinowski:
Currently, this raises an error:
``` r
library(S7)
an_s3_generic <- function(x) UseMethod("an_s3_generic")
method(an_s3_generic, S7::class_character) <- function(x) "foo"
#> Error: When registering methods for S3 generic an_s3_generic(), signature must be an S7 class, not a base type.
```

It would be nice if this worked and `method<-` was truly a single interface for registering any kind of method.

hadley:
I can't remember exactly why but I think I had a good reason for doing this.

t-kalinowski:
I think after #422, implementing this shouldn’t be too difficult, and we can quickly rediscover why it's not allowed 😆.

lawremi:
Agreed that this would need to happen if we really see S7 replacing the legacy S3 registration mechanism.

hadley:
Analysis from claude:

| Signature | S7 generic | S3 generic (`UseMethod`) | S3 Ops/internal (`+`, `[`, …) | S4 generic (`setGeneric`) |
|---|---|---|---|---|
| S7 class (`new_class("Foo")`) | ✓ | ✓ | ✓ | ✓ (S4 reg) |
| Base type (`class_character`, …) | ✓ | ✗ | ✓ | #642 |
| S3 class (`new_S3_class("foo")`) | ✓ | ✗ | ✓ | ✓ (S4 reg) |
| S7 union (`class_numeric`, `class_atomic`, …) | ✓ (flattened) | #636 | ✓ | ✗ |
| S4 class object | ✓ | ✗ | ✓ | ✓ |
| `class_any` | ✓ | ✗ | ✓ | ✗ |
| `class_missing` | ✓ | ✗ (only needed for multi-dispatch) | ✓ | ✗ |
| `NULL` | ✓ | ✗ | ✓ | ✗ |
| List signature (multi-dispatch) | ✓ (if generic has ≥2 dispatch args) | ✗ (S3 is single-dispatch) | ✓ (Ops dispatches on `e1, e2`) | ✓ |

**Notes:**

- **S3 Ops/internal column:** these *look* like S3 generics but `as_S3_generic()` rewrites them into entries of `base_ops` / `base_matrix_ops`, which are real S7 generics. So they accept everything the S7 column accepts.
[...]

PR #643, opened 2026-05-25
Complete S3 method registration

hadley:
Part of #455


R/method-register.R
```diff
@@ -99,22 +103,29 @@ register_S3_method <- function(
...
+    class_type(sig),
+    `NULL` = "NULL",
+    missing = stop(
+      "`class_missing` not supported for non-operator S3 generics.",
+      call. = FALSE
+    ),
+    any = "default",
+    S7_base = sig$class,
+    S7 = S7_class_name(sig),
+    S7_union = , # Handled above
```
t-kalinowski: This comment is confusing. Should this be a `stop("Unreachable")` for clarity?

tests/testthat/test-method-register.R
```diff
@@ -64,10 +64,60 @@ describe("method registration", {
...
   })
 
-  it("S3 registration requires a S7 class", {
-    foo <- new_class("foo")
+  it("can register S7 method for S3 generic with base type signature", {
+    s3_gen <- function(x) UseMethod("s3_gen")
+    method(s3_gen, class_character) <- function(x) "char"
+    method(s3_gen, class_integer) <- function(x) "int"
+
+    tbl <- environment(s3_gen)[[".__S3MethodsTable__."]]
```
t-kalinowski: I'm surprised by this, are we creating the binding? Do we really want to be doing this? I was under the impression that binding is only used by generics exported from a package namespace, not generally. 

Also, we might need to redesign this approach, (either by adding support, or adding guardrails and throwing errors when we can't satisfy the registration request), and expand the tests to actually excercise dispatch. E.g., this currently fails: 

```r
s3_gen <- local({
  function(x) UseMethod("s3_gen")
})

local({
  method(s3_gen, class_character) <- function(x) "char"
  method(s3_gen, class_integer) <- function(x) "int"
})

s3_gen("foo")
```

tests/testthat/test-method-register.R
```diff
@@ -64,10 +64,61 @@ describe("method registration", {
     }
   })
 
-  it("S3 registration requires a S7 class", {
-    foo <- new_class("foo")
+  it("can register S7 method for S3 generic with base type signature", {
+    local_s3_generic("s3_gen")
```
hadley, line 81: @t-kalinowski I think this looks better now — it uses the technique that I remember, i.e. you need to register the generic in the global namespace. This is a bit ugly but I don't think there's much that S7 can do about it.

t-kalinowski (commented):
Unfortunatly, the snippet I gave earlier still fails:

```r
s3_gen <- local({
  function(x) UseMethod("s3_gen")
})

local({
  method(s3_gen, class_character) <- function(x) "char"
  method(s3_gen, class_integer) <- function(x) "int"
})

s3_gen("foo")
#> Error in UseMethod("s3_gen") :
#>   no applicable method for 's3_gen' applied to an object of class "character"
```

I think this is a limitation of `registerS3method()`. It registers into `environment(generic)$.__S3MethodsTable__.`, but `UseMethod()` consults the table associated with `topenv(environment(generic))`. For normal top-level/package generics these are the same, but for local/generated generics they differ.

If we manually assign into the table that `UseMethod()` actually consults, it works:

```r
s3_gen <- local({
  function(x) UseMethod("s3_gen")
})

table_env <- topenv(environment(s3_gen))
table_env$.__S3MethodsTable__. <- new.env(hash = TRUE, parent = baseenv())

assign(
  "s3_gen.character",
  function(x) "char",
  envir = table_env$.__S3MethodsTable__.
)

s3_gen("foo")
#> [1] "char"
```

So I think the patch should keep using base `registerS3method()` generally, but branch for this special case:

```r
register_S3_method <- function(generic, signature, method, envir = parent.frame()) {
  class <- s3_signature_class(signature[[1]])

  generic_fun <- get(generic$name, mode = "function", envir = envir)
  generic_env <- environment(generic_fun)
  dispatch_env <- topenv(generic_env)

  if (!identical(generic_env, dispatch_env)) {
    table <- dispatch_env[[".__S3MethodsTable__."]]
    if (is.null(table)) {
      table <- new.env(hash = TRUE, parent = baseenv())
      dispatch_env[[".__S3MethodsTable__."]] <- table
    }

    assign(paste(generic$name, class, sep = "."), method, envir = table)
    return(invisible(method))
  }

  registerS3method(generic$name, class, method, envir = envir)
  invisible(method)
}
```

That lets the ordinary path continue to delegate to base R, while avoiding `registerS3method()` for the case that it does not handle ideally.

hadley:
Can you expand on your examples using `local()` are a use case that we should support? I don't think the goal here is to support every weird and wacky way that you might try to register S3 methods, but instead make this work for the typical use case (i.e. defining S3 methods in an package).

That said, your proposed fix isn't a lot of code, so maybe it's worth patching around the lack of comprehensive API in base R? It just makes me feel a little uncomfortable to  be providing tools for registering S3 methods that don't exist elsewhere.

t-kalinowski:
I'm thinking in general of the category of systems like https://github.com/klmr/box, https://github.com/rticulate/import, https://github.com/wahani/modules, (or my own https://t-kalinowski.github.io/envir/reference/import_from.html when used with an R script) where functions can be defined in one environment and then can be made available in another.

hadley:
I'm a little sceptical that this is important but it's also not that much code.

Issue #456, opened 2024-10-08
Mixing S4 and S7 inheritance hierarchies


lawremi:
It may be possible to have an S4 object inherit from S7_object. Some things that would need to happen:

- To enable an S4 class inheriting from an S7 class, `S4_register()` would need to pass a transient class to `S4Class=` on `setOldClass()` to establish the inherited slots on the S4 class.
- To enable S7 extending S4, we would need to remove the `stop()` and figure out things like how to construct the parent S4 instance, allow access to the S4 properties in `prop_set_()` and probably other things.
- To enable either direction, and support getters/setters on the S7 properties, `@()` would need to prioritize dispatch (instead of guarding against S4 objects). This would have a performance impact on S4 code that uses a lot of `@()`. We may have to sacrifice that feature for now.

hadley:
The first bullet is possibly related to #214

teunbrand:
Do you mean in the sense of `setClass("className", contains = <S7_class>)`? If so, is there any further guidance on this?

PR #644, opened 2026-05-25
Complete S4 method registration

hadley:
Part of #455

Someone who actually uses S4 really needs to take this for a test drive, because this PR revealed some longstanding bugs in the existing S4 support that no one has reported.


lawremi:
I'm going to work on #456 and as part of that will stress test this by porting something like SummarizedExperiment to S7, which will include extending S4 classes with S7 and using the S7 syntax to define methods on S4 generics. But let's go ahead and merge this as-is.

PR #648, opened 2026-05-26
Add `nameOfClass` method for S3 class objects

### LINKED ISSUE #521: add nameOfClass.S7_S3_class
Related to #432 , it would also be nice for `nameOfClass()` to support S3 class objects. If this makes sense, implementation should be as easy as:

```r
nameOfClass.S7_S3_class <- function(x) x$class[1L]
```

hadley:
Fixes #521


tests/testthat/test-S3.R
```diff
@@ -43,6 +43,14 @@ test_that("can construct data frame subclass", {
   expect_s3_class(df, "data.frame")
 })
 
+test_that("inherits() works with S7_S3_class", {
+  skip_if(getRversion() < "4.3")
```
hadley: Should use more specific skip helper.

t-kalinowski (approved):

PR #652, opened 2026-05-26
Expand `convert()` to handle base/S3 classes

### LINKED ISSUE #537: Default `convert()` method for upcasting throws error for classes whose parent is a `base_class`
The default implementation of `convert` for "upcasting"* attempts to preserve any properties that the subclass inherits and re-apply them to the parent class. However `base_class`'s don't have a `@properties` field, which throws an error when they're attempted to be read.

```r
my_logical <- S7::new_class("my_logical", parent = S7::class_logical)
S7::convert(TRUE, my_logical)
#> Error in .Primitive("@")(NULL, properties) :
#>  no applicable method for `@` applied to an object of class "NULL"

traceback()
#> 3: do.call(base::`@`, list(object, name)) at compatibility.R#20
#> 2: S7_class(from)@properties at convert.R#112
#> 1: S7::convert(TRUE, my_logical)
```

---

*sidenote: In the docs for `?convert`, I think the terms "upcasting" and "downcasting" are reversed.

hadley:
Fixes #537


R/convert.R
```diff
@@ -128,22 +129,33 @@ convert_down <- function(from, to, ...) {
...
-  from_prop_names <- names(from_props)
+  from_class <- S7_class(from)
+  user_args <- list(...)
 
-  # If `to` constructor has no `...`, only use properties that match its args
-  to_constructor_arg_names <- names(formals(to))
-  if (!"..." %in% to_constructor_arg_names) {
-    from_prop_names <- intersect(from_prop_names, to_constructor_arg_names)
-  }
+  if (is_class(from_class)) {
```
hadley: This diff is a bit hard to read, but it's just nesting the existing code inside the `is_class()` branch.

R/convert.R
```diff
@@ -128,22 +129,33 @@ convert_down <- function(from, to, ...) {
...
+
+    # Drop properties overridden by user-supplied arguments
+    from_prop_names <- setdiff(from_prop_names, names(user_args))
 
-  from_prop_values <- props(from, from_prop_names)
-  constructor_args <- c(from_prop_values, user_args)
+    from_prop_values <- props(from, from_prop_names)
+    constructor_args <- c(from_prop_values, user_args)
+  } else {
+    # `from` is a base or S3 object; pass it as `.data` to the constructor
```
hadley: This is the new code here

R/convert.R
```diff
@@ -128,22 +129,33 @@ convert_down <- function(from, to, ...) {
...
+    if (".data" %in% names(user_args)) {
+      constructor_args <- user_args
```
t-kalinowski: How would this branch be excercised from user code? I'm not quite following what kind of user call it makes sense to discard `from` with. 
I also don't see this branch excercised in tests.

R/convert.R
t-kalinowski, line 110: We should think about what behavior makes sense with `class_environment` here, before #622 lands. Statefully modifying attributes in `convert()` is probably not what we want. 

We might want to throw an error from the default `convert()` method if `from` is an environment.

hadley: Done in #622

R/convert.R
hadley: Yeah, good catch, it's not needed.

hadley:
@t-kalinowski way simpler now!

t-kalinowski (approved):

PR #654, opened 2026-05-27
Only validate parent properties once

### LINKED ISSUE #539: Parent properties are repeatedly re-validated for every descendent during construction
Similar to (but not the same as) the already-fixed #291, property-level validation is repeatedly run for every descendent of a class, even if the property has not been overridden:

``` r
library(S7)

parent = new_class("parent",
  properties = list(x = new_property(validator = \(value) cat("x validator\n"), default = 1)),
  validator = \(self) cat("parent validator\n")
)
child = new_class("child",
  parent = parent,
  validator = \(self) cat("child validator\n")
)
grandchild = new_class("grandchild",
  parent = child,
  validator = \(self) cat("grandchild validator\n")
)

parent()
#> x validator
#> parent validator
#> <parent>
#>  @ x: num 1
child()
#> x validator
#> parent validator
#> x validator
#> child validator
#> <child>
#>  @ x: num 1
grandchild()
#> x validator
#> parent validator
#> x validator
#> child validator
#> x validator
#> grandchild validator
#> <grandchild>
#>  @ x: num 1
```

<sup>Created on 2025-05-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I'd expect it only to validate properties once, at the level they are defined. Or perhaps more than that if a descendent overrides the same property (though ideally only if the overridden property changes the validator --- e.g. it would be nice if there were a mechanism for a child to change the default of a property but still have the validator on that property be triggered only once).

hadley:
Fixes #539


t-kalinowski (approved):
The tests are very clear!

hadley:
I'll pass on your compliments to claude 😆

PR #655, opened 2026-05-27
Consistent `...` handling

### LINKED ISSUE #423: `set_props()` fails when setting property named `object`
Fairly obvious. The rough equivalent `methods::initialize()` avoids this by naming its first parameter `.Object`, which is a lot less likely to collide.

### LINKED ISSUE #497: Constructors: Using `rlang::inject` with `new_object()`
Is there a way to use `rlang::inject` (or something similar) with `new_object`? It currently throws an error that `new_object()` must be called from within a constructor. Here's a reprex:

``` r
library(S7)
library(rlang)
library(purrr)
library(palmerpenguins)

# My real use case involves JSON data, so we'll transform into a list. 
my_penguin <- as.list(penguins[1,])

penguin <- new_class(
  "penguin",
  properties = list(
    species = class_factor,
    island = class_factor,
    bill_length_mm = class_double,
    bill_depth_mm = class_double,
    flipper_length_mm = class_integer,
    body_mass_g = class_integer,
    sex = class_factor,
    year = class_integer
  ),
  constructor = function(x) {
    # The `penguins` data frame contains all the possible properties for `penguin` objects.
    # For my real use case, these properties would be read from a JSON Schema file.
    penguins_properties <- set_names(colnames(penguins))
    # I want to pluck the property values for `my_penguin` as a named list that can be used
    # to construct the object.
    out <- map(penguins_properties, function(y) pluck(x, y))
    # This throws an error.
    inject(new_object(S7_object(), !!!out))
  }
)

penguin(my_penguin)
#> Error in new_object(S7_object(), species = structure(1L, levels = c("Adelie", : `new_object()` must be called from within a constructor
```

<sup>Created on 2024-11-18 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hadley:
* Allow `...` to be a single unnamed list. Fixes #497.
* Use `_` prefix to avoid accidental matches to existing properties. Fixes #423.


hadley:
For consistency we should do this in `set_props()` and `convert()` too.

hadley:
Use `stop2()` when available.

R/convert.R
```diff
@@ -127,12 +129,13 @@ is_down_cast <- function(x, class) {
...
 
-convert_down <- function(from, to, ...) {
+convert_down <- function(from, to, user_args = list()) {
   from_class <- S7_class(from)
 
   if (!is_class(from_class)) {
     # `from` is a base or S3 object; pass it as `.data` to the constructor
-    return(to(.data = from, ...))
+    user_args$.data <- from
+    return(do.call(to, user_args))
```
t-kalinowski, line 207: This approach now forces `...`, which will break any convert methods that want to introspect or delay evaluation of `...`.

R/utils.R
```diff
@@ -44,6 +44,28 @@ names2 <- function(x) {
...
+# Is `...` a single unnamed list argument? Used by functions that accept
+# either named arguments via `...` or a single list spliced in.
+is_single_list <- function(args) {
+  length(args) == 1L && !nzchar(names2(args)) && is.list(args[[1L]])
+}
+
+# Collect `...` into a named list. As a convenience, a single unnamed list is
+# spliced in so its elements become the values, making it easy to supply
+# values programmatically. All values must be named.
+splice_dots <- function(..., error_call = sys.call(-1)) {
```
t-kalinowski: I think this needs a better name. Maybe `collect_dots()`?

.claude/skills/property-dots/SKILL.md
```diff
@@ -0,0 +1,155 @@
...
+
+A function that returns a modified copy of an object, setting properties from
+`...`:
+
+```r
+#' @param _object The object to modify. It has an unusual name to avoid clashing
+#'   with property names supplied in `...`; see [new_property()] for details.
+#' @param ... Name-value pairs of properties to set. As a convenience, you can
+#'   supply a single unnamed list instead of individual name-value pairs.
+update_props <- function(`_object`, ...) {
```
t-kalinowski: Reading through this skill, I can't help but wonder, should `set_props()` simply take two arguments and we can avoid all this pain by simply removing the potential for name clashes? 

```r
set_props <- function(object, props) {} # where props is a named list
```

t-kalinowski (commented):
I think it's a fine idea to rename `.parent` to `_parent`, and similar.

I'm more ambivalent on `splice_dots()`, as implemented. It forces promises, and I think the name is misleading.

Also, my experience with `reticulate::tuple()` is that conveniently unpacking a list is sometimes quite a mental tax in practice, since, when writing against it, you're always guessing or trying to anticipate whether an argument may or may not be unpacked. Eventually, you come to the conclusion that the only reliable way to use it is with `tuple(list(...))`.

Another pattern I've seen sometimes, and kind of like, is to do something like this:

```r
new_object <- function(.object, ..., .props = list(...)) {}
```

edit: just saw that you've already suggested the same in https://github.com/RConsortium/S7/issues/497#issuecomment-4518925782

R/convert.R
hadley: That's true, but it's not a change: check out the old line 152 below. 

(I mean now we do force it for this very specific case, but that's only available in the dev version, so it's not a breaking change).

hadley:
In general, I am also in favour of more explicit unpacking, but this is an idiom that's used elsewhere in base R (e.g. `stop()`), and I don't think it's such a common need that it's worth adding an additional argument.

Similarly, in an ideal world, we wouldn't force the dots so early, but in all the existing use cases they were already forced at some point, and I don't think we've ever considered NSE for property setters, and I doubt we'd want to do the work to fully flesh that out.

hadley:
@lawremi do you have any opinions here? Are there patterns in base/methods that you think we should copy?

lawremi (approved):
This is a nice improvement. 

It would have been nice if `methods::initialize()` behaved like this, instead of treating every unnamed argument as a prototype. `new_object()` covers the common case with `_parent`. I don't recall ever passing multiple unnamed arguments to `initialize()`.

lawremi:
Reading some more of the discussion, as mentioned, there are examples of both explicit and implicit unpacking in R/Bioc. Since every argument to `new_object()` except `_parent` needs to be named, I think implicit unpacking is obvious enough. As has been said, it keeps the formals clean for what is a somewhat niche use case.

PR #657, opened 2026-05-27
Add `class_NULL` alias

### LINKED ISSUE #489: export `class_NULL`?
Should we add `class_NULL`? The primary benefit would be improved discoverability through autocompletions on `S7::class_*` and in the package documentation.

hadley:
Fixes #489


t-kalinowski (commented):
I thought the plan was to update the docs and sprinkle uses of `NULL` across the examples, rather than add and export a new symbol?

hadley:
IMO it's a simple enough alias, it puts `NULL` on the same footing as all other base types, and it's more discoverable.

t-kalinowski:
I'm neutral, leaning slightly against it. It doesn't satisfy any of the expectations about what other `class_*` objects look like: e.g., no `class` attribute, no print method, no constructor, and no validator.

It would help discoverability to add it, but there are other ways to aid discoverability by sprinkling `| NULL` throughout some examples and vignettes.

PR #659, opened 2026-05-27
Support S7 extension of S4 classes and vice-versa

### LINKED ISSUE #456: Mixing S4 and S7 inheritance hierarchies

lawremi:
Fixes #456. 

First iteration on enabling an S7 class to extend an S4 class and vice-versa.

The concrete, fail-fast goal is to enable SummarizedExperiment, a complex, central package in Bioconductor, to be rewritten in S7, based on S4 classes defined lower in the stack, like those from S4Vectors and IRanges, while not breaking compatibility with dependent packages, like SingleCellExperiment, via S4 shims extending the S7 classes. SingleCellExperiment is complex and deeply dependent on SummarizedExperiment, so if it continues working, it is a very good indicator.

What has been achieved outside of these improvements to S7:
* Fixed a bug in the methods package related to upcasting S4 objects derived old classes. Should be part of the next patch release, but it will cause release failures in the checks.
* Modified S4Vectors devel branch to use a `class1()` helper to be explicit when it wants the scalar name of the object's class and to ensure that it is robust to S3-style class vectors. Also improved the `setValidity2()` helper so that validity functions are enclosed in the S4Vectors namespace, enabling S4 class objects to remain identical across lazy loading. This is important since S4 class objects end up in the S7 class object, which is carried by S7 objects, and we want them to pass identity checks across packages.
* Rewrote SummarizedExperiment ([branch](https://github.com/lawremi/SummarizedExperiment/tree/s7-migration)) to use S7 classes, S7 generics and S7 syntax for defining methods on upstream S4 generics. Also defined S4 classes with the original names that derive from their corresponding S7 classes as a compatibility layer. It passes its tests, which were ported to S7 while preserving the original logic. 
* Ensured SingleCellExperiment passes tests _without any modification_.
* The path to Bioconductor adoption of S7 is now wide open.

For S7 classes extending S4 classes, the current approach is to instantiate S7 objects, meaning that the S4 bit is not set. The justification is that there are S4 object invariants that an S7 object cannot satisfy, and vice versa. The big one being the return value of `class()`. Instead we have an S7 shim that does a reasonable job of mimicking an S4 object, including slot access, validity checking, and initialization, but _not_ `class()` compatibility. See the evolving man page insert for more details.

When an S7 class is made to inherit from an S4 class, `S4_register()` is automatically called on the S7 class. This branch extends `S4_register()` to declare the S7 properties and `S7_class` attribute as S4 slots, which is helpful for making the S4 class more self-describing but critical in case the S7 class needs to be later extended by S4.

To extend an S7 class using `setClass()`, the S7 class first needs to be registered via `S4_register()` and `S4_contains()` should be called to obtain a suitable value for the `contains=` argument.  `S4_contains()` is an abstraction that currently just requires the S7 properties to be static, ie not using getters and setters, since S4 code using `slot()` would bypass those.  Any instances of an S7-derived S4 class are proper S4 objects, with the bit set and a scalar class vector.

Besides adding and changing hundreds of lines of logic around S4 classes, significant changes to support this in S7 include:
* Made S7 property storage compatible with S4 `NULL` slot sentinels. Until now, setting an S7 property to `NULL` removed the attribute, and S4 does not like it when its slots disappear from the attributes.
* The inheritance checks in `prop.c` need to be robust to scalar class vectors, using `@.S3Class` instead.
* Added S4 method registration for internal generics when S7 method signatures include classes with S4 ancestry. S4 methods on internal generics take precedence over S3 methods, so when S7 defines a method on an internal generic, it should define an S4 method if any of the classes in the signature are S4-derived classes, since otherwise it might be overridden by an S4 method defined for a parent class. S7 must also support multi-dispatch methods on internal generics where S4 supports it. The `dimnames<-()` generic is an example.
* `@<-.S7_object()` needs to support slots defined on S4 children that are not S7 properties, because S4 classes that extend `S7_object` will dispatch to the method, even when the S4 bit is set.

Resolving the two incompatible `class()` return values, scalar vs. non-scalar, is a larger issue. My proposal is that we export `methods:::.class1()` as `methods::class1()` and encourage all S4 code to use `class1()` to make the expectation of a scalar return value explicit. Practically, we could update the core of Bioconductor to adhere to that policy and thus enable new Bioconductor packages to extend the infrastructure with S7.


man/rmd/S4-compatibility.Rmd
```diff
@@ -0,0 +1,63 @@
+`new_class()` can use an S4 class as its parent. The S4 class can be supplied
```
hadley: I think this would probably be better off in `vignettes/compatibility.Rmd`. We're not super consistent on where we put these docs, but I think longer form stuff is just easier to read in vignettes, and then it's next to the S3 equivalent.

man/rmd/S4-compatibility.Rmd
```diff
@@ -0,0 +1,63 @@
...
+```r
+methods::setClass("Parent", slots = list(x = "numeric"))
+Child <- new_class("Child", methods::getClass("Parent"),
+  properties = list(y = class_character)
+)
+
+methods::is(Child(x = 1, y = "a"), "Parent")
+```
+
+Things that work reasonably:
```
hadley:
```suggestion
Things that work reasonably well:
```

man/rmd/S4-compatibility.Rmd
```diff
@@ -0,0 +1,63 @@
...
+* `methods::as(child, "Parent")` and `convert(child, to = Parent)` can produce
+  a real S4 object for the S4 parent class.
+
+Caveats:
+
+* An S7 object that directly extends S4 is currently represented as an S3
+  old-class object, not as an S4-bit object, so `isS4(child)` is `FALSE`.
+  This avoids advertising full S4 object invariants that the object does not
+  satisfy.
+* The S3 class vector is not scalar. S4 code that needs one primary class name
```
hadley: How common is this use of scalar `class()`?

man/rmd/S4-compatibility.Rmd
```diff
@@ -0,0 +1,63 @@
...
+  should use `class(x)[1L]`, or `methods::class1(x)` if available.
+* `methods::new(class(x)[1L], ...)` creates a fresh S4 object from the S4 class
+  definition. For an S7 class registered with S4, that object can inherit from
+  `S7_object` through the old-class graph while lacking the `S7_class` slot or
+  attribute that makes it an S7 object. Code that updates an existing object
+  should prefer `methods::initialize(x, ...)`.
+* S4 methods that call `methods::slot<-()` can bypass S7 property setters and
```
hadley: This is consistent with the design of S4 right, i.e. where you are generally responsible for validation after directly modifying a slot?

man/rmd/S4-registration.Rmd
```diff
@@ -0,0 +1,42 @@
...
+needs to dispatch on the class or use it in a slot.
+
+`S4_register()` invisibly returns the registered S4 class name. For an S7 class
+this is the S4 old-class name used for dispatch. For an S7 union this is the
+S4 class union name.
+
+`S4_register_contains()` is for the opposite direction: use it when an S4 class
+needs to extend an S7 class with `methods::setClass(contains = )`. It registers
+the S7 class if needed, then creates and returns an S4 shim class whose name
+ends in `::S4Slots`. The shim:
```
hadley:
```suggestion
`S4_register_contains()` is for inheritance: use it when an S4 class
should be allowed to extend an S7 class with `methods::setClass(contains = )`. It registers
the S7 class if needed, then creates and returns an S4 shim class whose name
ends in `::S4Slots`. The shim:
```

?

R/class-spec.R
```diff
@@ -192,9 +201,11 @@ class_constructor <- function(.x) {
 class_validate <- function(class, object) {
   validator <- switch(
     class_type(class),
-    S4 = function(object) {
-      check <- methods::validObject(object, test = TRUE)
-      if (isTRUE(check)) NULL else check
+    S4 = if (isS4(object)) {
```
hadley: Do we still need this check? Doesn't `isS4()` need to be true in order for `class_type()` to return S4?

R/class.R
```diff
@@ -124,16 +128,19 @@ new_class <- function(
     }
     if (
       abstract &&
-        (!is_class(parent) || !(parent@abstract || parent@name == "S7_object"))
+        !((is_class(parent) &&
+          (parent@abstract || parent@name == "S7_object")) ||
+          (is_S4_class(parent) && parent@virtual))
```
hadley, line 142: IMO now worth refactoring into a helper function with an informative name

R/class.R
```diff
@@ -436,4 +460,14 @@ check_prop_names <- function(properties, call = sys.call(-1L)) {
     )
     stop2(msg, call = call)
   }
+
+  reserved <- intersect("S7_class", names(properties))
```
hadley: Can we leave this for #677?

R/constructor.R
```diff
@@ -8,12 +8,26 @@ new_constructor <- function(
...
+    s4_data_part <- is_S4_class(parent) && ".Data" %in% names(parent@slots)
+    parent_call <- if (s4_data_part) {
+      bquote(
+        methods::getClass(.(as.character(parent@className)))@prototype@.Data
+      )
+    } else if (has_S7_symbols(envir, "S7_object")) {
+      quote(S7_object())
+    } else {
+      quote(S7::S7_object())
+    }
```
hadley, line 33:
```suggestion
    if (s4_data_part) {
      class <- as.character(parent@className)
      parent_call <- bquote(methods::getClass(.(class)@prototype@.Data)
    } else if (has_S7_symbols(envir, "S7_object")) {
      parent_call <- quote(S7_object())
    } else {
      parent_call <- quote(S7::S7_object())
    }
```

IMO that's a little easier to read

R/method-register.R
```diff
@@ -189,10 +189,18 @@ as_signature <- function(signature, generic, call = sys.call(-1L)) {
...
+    if (length(signature) == length(S4_signature)) {
+      n <- length(S4_signature)
+    }
+  }
+
   if (n == 1) {
     # Accept a bare list of length 1 too, for symmetry with multi-dispatch
     # generics where a list is required (#555).
-    if (is.list(signature) && !is.object(signature) && length(signature) == 1) {
+    if (is_multi_arg_signature(signature) && length(signature) == 1) {
```
hadley: I think the length check should go in the helper too?

R/method-register.R
```diff
@@ -208,6 +216,10 @@ as_signature <- function(signature, generic, call = sys.call(-1L)) {
   }
 }
 
+is_multi_arg_signature <- function(signature) {
```
hadley: Nice

R/S4.R
```diff
@@ -19,23 +27,494 @@
 #' method(S4_generic, Foo) <- function(x) "Hello"
 #'
 #' S4_generic(Foo())
+#'
+#' S4Foo <- new_class("S4Foo", properties = list(x = class_numeric), package = "S7")
+#' S4Foo_S4 <- S4_register_contains(S4Foo)
+#' methods::setClass("S4Child", contains = S4Foo_S4)
 S4_register <- function(class, env = parent.frame()) {
   if (is_class(class)) {
```
hadley: I think this would be a bit clearer if we moved all the early returns first?

tests/testthat/test-class.R
```diff
@@ -63,10 +63,10 @@ describe("S7 classes", {
     })
   })
 
-  it("can't inherit from S4 or class unions", {
+  it("can inherit from S4 but not class unions", {
     parentS4 <- methods::setClass("parentS4", slots = c(x = "numeric"))
     expect_snapshot(error = TRUE, {
-      new_class("test", parent = parentS4)
+      new_class("test", parent = parentS4, package = NULL)
```
hadley: I think this should now move out of the snapshot test to get something more strict.

tests/testthat/test-convert.R
```diff
@@ -127,6 +127,77 @@ describe("fallback convert", {
     expect_false(S7_inherits(obj))
     expect_equal(attr(obj, "x"), NULL)
   })
+
+  it("can convert_up() an S4-derived S7 object to an S4 object ", {
+    on.exit(S4_remove_classes(c("ParentS4", "ChildS7")))
```
hadley: Might be worth considering extening `local_S4_class()` to handle more cases so we could skip the explicit cleanup steps (which are easy to forget)?

hadley (commented):
I assume you know infinitely more about S4 than me 😄 so I focussed on docs and style.

Also need an update to `NEWS.md`?

t-kalinowski (commented):
@lawremi I took a cursory look and didn’t see any obvious issues. Could you please tag me again once you’ve had a chance to resolve Hadley’s comments? I’ll take a more thorough look then. That way I can avoid duplicating effort or leaving comments that might become stale.

man/rmd/S4-compatibility.Rmd
lawremi: In general, I think it depends on whether we're demonstrating something with literate programming or if it's just reference documentation, like this. People will naturally go to a man page to understand how something works. However, in this case, it looks like there is precedent, so I'll move it to the vignette and put a pointer in the man page.

man/rmd/S4-compatibility.Rmd
lawremi: Pretty common, unfortunately. I would guess most uses are in `show()` methods.

man/rmd/S4-compatibility.Rmd
lawremi: Yes, the slot assignment itself is checked, but validation needs to be explicit. The recommended practice in Bioconductor is to use `initialize()` to set slots, which sets them in batch and then validates.

R/method-register.R
lawremi: Perhaps this is a bit confusing. The length check is checking that there is only one arg. The "mutli-arg signature" is referring to the signature being a plain list. I'll rename that helper for clarity.

R/S4.R
lawremi: Agreed.

tests/testthat/test-convert.R
lawremi: `local_S4_class()` is convenient and robust, but it obfuscates the code, in my opinion. I think readers expect to see `setClass()`, `setOldClass()`, and `setClassUnion()`. I'd prefer something like `S4_remove_after(setClass(...))`. 

How about we defer this to a future change after merging this branch and discussing? Since the changes are so extensive, rebasing is becoming a chore. It might be better to iterate over multiple MRs.

tests/testthat/test-class.R
lawremi: Agreed, will make that stricter.

lawremi:
I think all of the major comments have been resolved, so @t-kalinowski can please give it a look.

t-kalinowski (commented):
See comments in https://github.com/RConsortium/S7/pull/720

lawremi:
Since @t-kalinowski had the good idea of using AI to help review this branch, I asked an agent to generate this review guide that walks a reviewer through the diff. The links are to code, not diffs, so I'm not sure how helpful they are, but they do point to cohesive bits of logic. Please let me know if this helps and/or if you'd like to schedule a call for me to walk you through this.

# Review Guide for PR #659: S4/S7 Inheritance

PR: <https://github.com/RConsortium/S7/pull/659>  
Full diff: <https://github.com/RConsortium/S7/pull/659/files>  
Branch compare: <https://github.com/RConsortium/S7/compare/main...issue-456-s4-extends-s7>

This PR is large because the feature crosses object representation, S4 class
registration, construction, initialization, validation, property access, and
method registration. This guide is intended to make the review tractable by
separating the design invariants from the files that implement them.

## What This PR Enables

The goal is two-way S4/S7 inheritance interop:

1. An S7 class can extend an S4 class.
   - S4 slots are represented as S7 properties.
   - S4 dispatch and `methods::is()` see the S4 parent.
   - The resulting S7 object remains an old-class/S3-style object, so
     `isS4()` is still `FALSE`.

2. An S4 class can extend an S7 class.
   - S7 properties are exposed as formal S4 slots on the reified old class.
   - S4 subclasses are real S4 objects.
   - S7 dispatch still sees the S7 class through `_S7_class`.
   - S4 validity can invoke S7 validation.

3. S7 method registration works when S4/internal generics are involved.
   - Internal generics with S4 dispatch, such as `dim()` or `dimnames<-`, can
     get S4 methods when an S7 class has an S4 ancestor.
   - External-generic sentinels preserve enough information to register full
     S4 signatures.

## Suggested Review Order

### 1. Start With The Public Contract

Read these first, before implementation details:

- `new_class()` S4 compatibility docs:  
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/class.R#L17-L69>
- `S4_register()` / `S4_contains()` docs and examples:  
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L1-L80>
- Compatibility vignette, S4/S7 section:  
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/vignettes/compatibility.Rmd#L180-L298>

Questions to answer at this level:

- Is the advertised behavior the right behavior?
- Are the documented caveats explicit enough?
- Is `S4_contains()` the right user-facing name for “this S7 class is suitable
  for `methods::setClass(contains = )`”?

### 2. Review The Core S4 Bridge

Most of the feature lives in `R/S4.R`:

- Public entry points and S4 class conversion:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L47-L154>
- S4 ancestry and S7-over-S4 detection:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L195-L218>
- Compatibility constraints for S4 subclasses of S7:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L236-L257>
- Reified old-class definition:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L259-L351>
- S4 validity integration:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L376-L412>
- S4 `initialize()` integration:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L414-L480>
- Mapping S4 classes and slots back to S7 classes/properties:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L484-L533>

Key invariants:

- `S4_register()` always creates a reified S4 old class for an S7 class.
  This avoids a dependent package being stuck with a non-reified old class if it
  later wants to extend the S7 class from S4.
- The S4-visible class carries a formal `_S7_class` slot whose value has S4
  class `"S7_class"`.
- `S4_contains()` does not register as a side effect; it asserts that the class
  is registered and that its properties can safely cross the S4 inheritance
  boundary.
- Properties with custom getters or setters cannot be exposed as S4 slots.
- S7 unions are not eagerly registered except when the user explicitly calls
  `S4_register()` on the union; when possible, existing S4 unions are reused.

### 3. Review Object Representation And Class Lookup

These are the representation-level changes reviewers should check carefully:

- `new_class()` automatically registers S7 classes with S4 when they extend S4:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/class.R#L148-L183>
- `S7_class()` handles S4 objects that carry `_S7_class`:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/class.R#L455-L468>
- `has_S7_class()` uses the C accessor, not direct attribute/slot assumptions:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/inherits.R#L49-L53>

Key invariants:

- S4 subclasses that inherit from S7 use a formal `_S7_class` slot
- R code should use `S7_class(object)` or `.Call(S7_class_, object)`, not
  direct `slot(object, "S7_class")` or `attr(object, "S7_class")`.

### 4. Review Property Access Across The Boundary

Property access is intentionally not perfectly encapsulated once S4 code is
allowed to see S7 properties as slots. Important pieces:

- `@<-` forwards S4-only slots to S4 and S7 properties to S7:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/property.R#L366-L375>
- C property get/set use `_S7_class` and the same storage path for S4/S7:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/src/prop.c#L506-L620>
- Upcasts drop stale `_S7_class`/legacy `S7_class` metadata before installing
  the target class:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/convert.R#L163-L198>

Known compromise:

- `methods::slot()` and `methods::slot<-()` can bypass S7 property setters.
  This is a consequence of exposing properties as S4 slots for compatibility.
  `methods::validObject()` should be used when S4 code mutates those slots.

### 5. Review Validation

The validation path is split between S7 and S4:

- S7 recursive validation stops at a supplied parent:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/valid.R#L90-L148>
- S4 validity for the reified old class calls S7 validation when the object
  matches the S7 class and skips duplicate validation for upcast parent views:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/S4.R#L376-L412>

Key questions:

- Does S4 validation call into S7 in the right places?
- Does it avoid recursively validating the same S7 hierarchy repeatedly?
- Are invalid S4/S7 mismatches reported as failures rather than silently
  accepted?

### 6. Review Dispatch And Method Registration Last

The dispatch-related change is smaller and mostly separable:

- S3/internal generic registration delegates to S4 registration when an S4
  ancestor is involved:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/method-register-S3.R#L25-L35>
- S4 method registration and signature mapping:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/method-register-S4.R#L1-L49>
- `as_signature()` allows full S4 signatures for internal generics:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/R/method-register.R#L228-L255>

Key invariant:

- Registering an S7 method on an internal generic should still use the S3 path
  for ordinary S3/S7 behavior, but additionally register an S4 method when the
  signature has an S4 ancestor and S4 will be the runtime dispatch path.

## Tests Worth Reading

It is probably worth _skimming_ some of the tests.

Most of the behavioral spec is in `tests/testthat/test-S4.R`.

Recommended test slices:

- Basic `S4_register()` / `S4_contains()` / union registration:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L6-L120>
- S4 classes extending S7 classes, including slot preservation and coercion:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L187-L260>
- Abstract S7 classes represented in S4:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L326-L378>
- S4 prototypes and NULL-slot sentinel behavior:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L380-L435>
- Rejection of custom getters/setters across the S4 inheritance boundary:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L449-L504>
- S4/S7 union mapping:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L506-L565>
- S7 classes extending S4, `initialize()`, `@<-`, and S3 data parts:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-S4.R#L650-L785>
- Internal generic and external sentinel registration:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-method-register-S3.R#L58-L166>
- Conversion tests involving S4:
  <https://github.com/RConsortium/S7/blob/issue-456-s4-extends-s7/tests/testthat/test-convert.R>

## Lower-Risk / Mechanical Files

These are useful to skim after the core review:

- `NAMESPACE`: exports/imports for the new API.
- `man/*.Rd`: generated from roxygen, except worth checking wording.
- `NEWS.md`: high-level feature note.
- `.lintr.R`: local lintr defaults.
- `.Rbuildignore`: vignette/review-support housekeeping.
- Snapshot changes under `tests/testthat/_snaps/`.

## Known Caveats To Confirm

These are intentional and documented:

- S7 objects that directly extend S4 are not S4 objects: `isS4(x)` is `FALSE`.
  S4 inheritance should be checked with `methods::is(x, "Parent")`.
- `class(x)` for S7 objects remains non-scalar to preserve S3/S7 dispatch
  semantics. S4 code that needs a primary name should use `class(x)[1L]`
  or `methods::class1(x)` when available.
- Exposing S7 properties as S4 slots means S4 code can bypass S7 setters with
  `slot<-`.
- Properties with custom getters/setters are rejected by `S4_contains()`.
- S7 unions are registered with S4 only when explicitly requested.
- `_S7_class` is the new storage name; `S7_class` is only a legacy read
  fallback for older serialized/lazy-loaded objects.

tests/testthat/test-convert.R
```diff
@@ -136,6 +136,49 @@ test_that("fallback convert can convert to base type", {
   expect_equal(attr(obj, "x"), NULL)
 })
 
+test_that("fallback convert can convert_up() an S4-derived S7 object to an S4 object", {
+  on.exit(S4_remove_classes(c("ParentS4", "ChildS7")))
+  setClass("ParentS4", slots = list(x = "numeric"))
+
```
hadley, line 142: Use `local_s4_class()` instead?

R/class-spec.R
```diff
@@ -190,12 +199,16 @@ class_constructor <- function(.x) {
 }
 
 class_validate <- function(class, object) {
+  if (is_S4_class(class)) {
+    if (isS4(object)) {
+      check <- methods::validObject(object, test = TRUE)
+      return(if (isTRUE(check)) NULL else check)
+    }
+    return(NULL)
```
hadley:
```suggestion
    if (isS4(object)) {
      check <- methods::validObject(object, test = TRUE)
      return(if (isTRUE(check)) NULL else check)
    } else {
      return(NULL)
    }
```

R/class.R
```diff
@@ -440,6 +475,15 @@ check_prop_names <- function(properties, call = sys.call(-1L)) {
...
   }
+
+  if ("S7_class" %in% nms) {
+    msg <- paste0(
+      "Property can't use S7 reserved name: ",
+      "S7_class",
+      "."
+    )
+    stop2(msg, call = call)
+  }
```
hadley: I think this is now outdated?

R/convert.R
```diff
@@ -185,9 +187,14 @@ convert_up <- function(from, to, call = sys.call(-1L)) {
       stop2(msg, call = call)
     }
 
-    from <- zap_attr(from, c(setdiff(from_props, to_props), "S7_class"))
+    from <- zap_attr(
+      from,
+      c(setdiff(from_props, to_props), "_S7_class", "S7_class")
+    )
```
hadley: I'm surprised to see this diff here. Isn't that already in main?

hadley (approved):
IMO it's worth doing a little more style polishing but really I it's fine to merge as is. It doesn't need to be perfect, it just needs to advance S4 support and if we get it in quickly we'll avoid a bunch of merge conflicts.

lawremi:
@t-kalinowski is it OK if we merge this and then address concerns from #720? There are some important things to fix there, but I'd rather do it on top of this.

hadley:
Tomasz is out on vacation so I’ll answer on his behalf 😀 yes it’s fine. 

I’ve also had conversations with him about my sense that codex tends to make code more technically correct at the expense of readability/understandability.

hadley:
🎉🎉🎉

Issue #609, opened 2026-05-06
Parent class constructor defaults are not safely scoped in child constructors


mitchelloharawild:
The default constructor's inheritance of parent constructor formals seems unsafe, since the evaluation environment might change. I encountered this when developing an S7 class in a package with a custom constructor, and then extending this class in a different package with the default constructor.

Perhaps passing parent properties to the parent constructor via `...` is a safer alternative? This would also fix #467 and #585.

Explanatory MRE:

``` r
library(S7)
```

The built-in constructor evaluates default arguments.

``` r
e <- new.env()
local({
  default <- 1L
  parent <- new_class(
    "parent",
    properties = list(
      x = new_property(class_integer, default = default)
    )
  )
}, envir = e)

# evaluated property defaults are used in formals
formals(e$parent)
#> $x
#> [1] 1
e$parent()
#> <parent>
#>  @ x: int 1
```

Suppose we prefer the constructor's formals to maintain the symbol defaults

```r
# Custom constructor for unevaluated defaults
local({
  default <- 1L
  parent <- new_class(
    "parent",
    properties = list(
      x = new_property(class_integer, default = default)
    ),
    constructor = function(x = default) {
      new_object(S7::S7_object(), x = x)
    }
  )
}, envir = e)

# Constructor with symbol property defaults used in formals
formals(e$parent)
#> $x
#> default
e$parent()
#> <parent>
#>  @ x: int 1
```

This works, the symbols are found from the correct environment when needed.

Then we inherit this class as a parent from a class in a different environment.

```r
child <- new_class(
  "child", 
  e$parent,
  properties = list(
    y = new_property(class_integer, default = 0L)
  )
)

# Parent constructor formals are copied to child constructor
# But the environment is now different, so the symbols cannot be found
formals(child)
#> $x
#> default
#> 
#> $y
#> [1] 0
child()
#> Error in `child()`:
#> ! object 'default' not found
```

The parent constructor's formals are merged with the child constructor's formals, and evaluated in the child constructor's environment -- so the symbol cannot be found.

If we instead pass through parent properties to the parent constructor via `...`, the appropriate arguments are evaluated in the appropriate environment.

```r

# Alternatively, passing ... to parent methods evaluates formals in the
# appropriate context/environment.
child <- new_class(
  "child", 
  e$parent,
  properties = list(
    y = new_property(class_integer, default = 0L)
  ),
  constructor = function(..., y = 0L) {
    new_object(e$parent(...), y = y)
[...]

PR #660, opened 2026-05-28
Implement external classes

### LINKED ISSUE #250: How to create a nestable R7 class?

### LINKED ISSUE #573: Defining generic methods for suggested packages

hadley:
I think the use of external is a little bit misleading here, because these are really "delayed" classes. But external matches external generic, which exist for a similar reason.

Fixes #250. Fixes #573.

Note this does not solve #317, since currently creating a subclass will force loading of the package via `class_constructor()`. But I think can fix that after #694 by using `...` to call constructors from other packages. That will decouple the superclass arguments from the subclass in a way that won't cause `R CMD check` to fail with a codoc mismatch if the parent args change.


hadley:
@t-kalinowski @lawremi I think this is now ready for your feedback.

lawremi (approved):
This is a very useful feature. Implementation looks good.

t-kalinowski (approved):
I pushed tests and fixes for some edge cases. Can you please take a look and see if the approach makes sense?

An open question is what behavior we want for property override checks when an external class is from a package that is not loaded yet. The current fix avoids treating an unresolved external class as matching every child type, but it's possible we want to take a different approach here (e.g., defer until the package is loaded, or load the package eagerly)

R/external-class.R
```diff
@@ -0,0 +1,209 @@
...
+find_external_class <- function(x) {
+  ns <- asNamespace(x$package)
+  if (exists(x$name, envir = ns, inherits = FALSE)) {
+    obj <- get(x$name, envir = ns, inherits = FALSE)
+    if (is_external_class_match(obj, x)) {
+      return(obj)
+    }
+  }
+
+  # Also consider cases where the constructor isn't named the same as the class
```
hadley: Are we sure that this is a case that we want to support? Why do we want to encourage people to have a mismatch between constructor and class name?

R/external-class.R
```diff
@@ -0,0 +1,209 @@
...
+#'
+#'   ```R
+#'   tree_stub <- new_external_class("mypkg", "tree")
+#'   new_class("tree", properties = list(child = NULL | tree_stub))
+#'   ```
+#'
+#' Make sure to call [S7_on_load()] in your package's `.onLoad()` so that
+#' deferred method registrations fire when the relevant package is loaded.
+#'
+#' External classes can not currently be used as parents in [new_class()].
```
hadley, line 31: Made this restriction more clear. Next step is to resolve #609, and #317.

t-kalinowski (commented):
I opened a PR with some proposed changes to this branch: https://github.com/RConsortium/S7/pull/710

Most of the changes address various edge cases raised by automated review.

One change I want to call out is that I'm proposing that external S7 classes must be exported by a package using the same symbol name as the class's `@name`. This makes resolution much simpler and faster. It also makes the code easier to reason about, both internally in S7 and in user code, and avoids introducing a bunch of complexity. In particular, we don't need to go on a fishing expedition looking for an exact match, which could potentially force a bunch of lazy bindings.

I also tightened up a few edge cases around this:

- external class resolution now has separate errors for missing packages, too-old package versions, and binding mismatches
- external classes work in union arms, including self-referential and mutually recursive class definitions
- child properties can narrow an external parent class to a loaded subclass
- subclassing against an unloaded external class now fails at the point where we need to resolve it
- method introspection and `method_explain()` now use the same required-resolution behavior
- unload cleanup handles the case where the external class package has already unloaded before the downstream package unloads

I also changed one of the tests from using `local_mocked_bindings()` to `local_package()`, to avoid needing to define `getNamespaceVersion` in the package namespace just to mock it during tests -- that seemed to me like test scaffolding leaking into production code.

PR #662, opened 2026-05-28
Strip external generics on build

### LINKED ISSUE #364: Embedding generics when defining methods
```R
library(S7)
foo1 <- new_class("foo1", properties = list(x = class_integer))
foo2 <- new_class("foo2", foo1, properties = list(y = class_double))

method(convert, list(foo1, class_integer)) <- function(from, to) {
  from@x
}
ls()
#> [1] "convert" "foo1"    "foo2"
```

I think this is going to be a problem because defining a method for a generic defined in another package will end up embedding a copy of that generic in your package. Not sure what we can do here given that this is a fairly deep seated behaviour of `<-` functions.

hadley:
Fixes #364


lawremi:
Might want to clarify in the docs that it should come after all method registration. The example in `zzz.R` shows that, but it might help to be explicit.

This also raises the idea of a combined function `S7::load()` that clears generics _and_ defines `.onLoad()` via side-effect. Probably not a very good idea, because confusion could result from another .onLoad() being defined. I wonder if the namespace system could support `.onLoad.<foo>` so that there can be multiple hooks.

tests/testthat/t4/R/t4.R
```diff
@@ -0,0 +1,16 @@
...
+
+# Register an S7 method for an S3 generic defined in another package (t3).
+S7::method(t3_s3, t4_class) <- function(x) "s3-dispatch"
+
+# Register an S7 method for an S4 generic defined in another package (t3).
+S7::S4_register(t4_class)
+S7::method(t3_s4, t4_class) <- function(x) "s4-dispatch"
+
+.onLoad <- function(libname, pkgname) {
+  S7::S4_register(t4_class)
```
hadley: @lawremi Claude discovered this while writing this test. Does that seem correct to you? (i.e. that `S4_register()`, which calls `setOldClass()` has to be called onLoad)

hadley:
Hmmm, I wonder if we can call `setHook(packageEvent(current_package, "onLoad"))` to register an `onLoad` hook in `S7_on_build()`?

...

Nope, that doesn't work.

tests/testthat/t4/R/t4.R
hadley: Or maybe this is related to the environment fixes in #643?

lawremi: Shouldn't be necessary to call `S4_register()` at runtime. All it does is call `setOldClass()` which of course works at build time. Strangely, Gemini suggested the same thing when converting a package to use S7.

hadley: Let's hope that #663 fixes this. I'll hold off on merging until that's confirmed.

hadley: And confirmed that it does 🎉

hadley:
@t-kalinowski the actual implementation is straightforward, but I added quite a bit of test infrastructure to reduce duplication. Let me know if you want that in a separate PR.

R/hooks.R
```diff
@@ -0,0 +1,69 @@
...
+  for (name in ls(ns, all.names = TRUE)) {
+    if (is_generic_sentinel(get0(name, envir = ns, inherits = FALSE))) {
+      rm(list = name, envir = ns)
+    }
+  }
+  invisible()
+}
+
+generic_sentinel <- function(generic) {
+  external <- as_external_generic(generic)
```
t-kalinowski, line 64: I think this makes it so that an external S3 generic is not callable after a method is registered.

I think it's fine, since it's rare to want to call a generic at pkg build time.

However, we can also maybe just do:

```r
attr(generic, "S7_external_generic_sentinel") <- TRUE
```

and keep the generic as-is.

tests/testthat/t3/NAMESPACE
```diff
@@ -0,0 +1,4 @@
+export(t3_s3)
+export(t3_s4)
+exportMethods(t3_s4)
```
t-kalinowski, line 3: It would be good to expand this test to also register a method for a base R generic (e.g., `sum`, `+`).

R/hooks.R
hadley: Well, it binds a non-function object to the name in the package environment. But I think you'll still be able to call it, you'll just be calling the original version, not the version copied into the package.

(I had claude attempt to keep it callable, but because of the overlap in the existing external generic pathway, the code ends up being quite complicated).

t-kalinowski (approved):

PR #663, opened 2026-05-28
fixes for S4 generics and classes defined across packages

lawremi:
This contains two fixes:
1. The current code assumes that `generic@package` indicates the home package of the generic. Unfortunately, that is not the case in general. The methods package uses `@package` to refer to the package defining the original function that the generic represents. Thus, `setGeneric("nrow", ...)` in a package will create a generic where `@package == "base"`. To make matters worse, `topenv(environment(generic))` is the namespace of `@package`, so the only way to find the package defining the generic is a brute force search through namespace imports.
2. Much easier, I noticed that the S4 method registration path was not passing along the calling environment to the `getClass()` call to resolve S7 old classes, even though that seemed to be the intent. Omitting `where` might problem if a parent package registers those classes with S4.

### DIFF
--- a/R/S4.R
+++ b/R/S4.R
@@ -150,7 +150,7 @@ S4_class_name <- function(x) {
 }
 
 S4_package_name <- function(f, env) {
-  if (methods::getPackageName(topenv(env)) == f@package) {
+  if (methods::getPackageName(topenv(env), create = FALSE) == f@package) {
     ## current ns might not be loaded yet, catch here
     return(f@package)
   }
--- a/R/S4.R
+++ b/R/S4.R
@@ -149,6 +149,46 @@ S4_class_name <- function(x) {
   }
 }
 
+S4_package_name <- function(f, env) {
+  if (methods::getPackageName(topenv(env)) == f@package) {
+    ## current ns might not be loaded yet, catch here
+    return(f@package)
+  }
+
+  name <- as.vector(f@generic)
+  generic_in_its_package <- methods::isGeneric(
+    name,
+    where = asNamespace(f@package)
+  )
+  if (generic_in_its_package) {
+    return(f@package)
+  }
+
+  # generic was defined for a function from a different package, like base
+  find_package_with_symbol(name, env, exclude = f@package) %||%
+    stop(
+      "Failed to find originating package for S4 generic '",
+      name,
+      "' in namespace imports.",
+      call. = FALSE
+    )
+}
+
+find_package_with_symbol <- function(name, env, exclude = NULL) {
+  imports <- getNamespaceImports(topenv(env))
+  pkgs <- setdiff(names(imports), exclude)
+  Find(
+    function(pkg) {
+      if (isTRUE(imports[[pkg]])) {
+        name %in% getNamespaceExports(pkg)
+      } else {
+        name %in% imports[[pkg]]
+      }
+    },
+    pkgs
+  )
+}
+
 S4_remove_classes <- function(classes, where = parent.frame()) {
   for (class in classes) {
     suppressWarnings(methods::removeClass(class, topenv(where)))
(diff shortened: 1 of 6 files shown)


hadley:
BTW I'm happy to review any of your PRs, but I've been waiting for you to explicitly request a review.

R/S4.R
```diff
@@ -148,6 +148,46 @@ S4_class_name <- function(x) {
   }
 }
 
+S4_package_name <- function(f, env) {
+  if (methods::getPackageName(topenv(env)) == f@package) {
```
hadley: Do we need `create = FALSE` here? (for safety)

R/S4.R
```diff
@@ -148,6 +148,46 @@ S4_class_name <- function(x) {
...
+      name,
+      "' in namespace imports.",
+      call. = FALSE
+    )
+}
+
+find_package_with_symbol <- function(name, env, exclude = NULL) {
+  imports <- getNamespaceImports(topenv(env))
+  pkgs <- setdiff(names(imports), exclude)
+  Find(
```
hadley: IMO I think this would be a little easier to understand if it used a for loop with an early return when the object was found, and then the error message from above could move into this function.

R/S4.R
```diff
@@ -148,6 +148,46 @@ S4_class_name <- function(x) {
...
+    name,
+    where = asNamespace(f@package)
+  )
+  if (generic_in_its_package) {
+    return(f@package)
+  }
+
+  # generic was defined for a function from a different package, like base
+  find_package_with_symbol(name, env, exclude = f@package) %||%
+    stop(
```
hadley: Probably worth switching to use `stop2()` and `sprintf()` for consistency with other errors in this package.

R/S4.R
```diff
@@ -148,6 +148,46 @@ S4_class_name <- function(x) {
   }
 }
 
+S4_package_name <- function(f, env) {
+  if (methods::getPackageName(topenv(env)) == f@package) {
+    ## current ns might not be loaded yet, catch here
+    return(f@package)
+  }
+
+  name <- as.vector(f@generic)
```
hadley: Should this be `as.character()`? But I don't know what `f@generic` would be otherwise.

lawremi: Used `as.vector()` to try to make it clear that we're stripping the package attribute. Obviously `as.character()` would do that also, but it also suggests that `f@generic` might be something other than a character vector. I'm ambivalent on this.

hadley: Ah, got it. Maybe 

```R
name <- f@generic
attr(name, "package") <- NULL
```

Maybe too explicit?

R/S4.R
lawremi: Good catch, added it.

R/S4.R
lawremi: I couldn't find `stop2()` anywhere, but I switched this to use `sprintf()`.

R/S4.R
lawremi: Switched to using `for()` instead of `Find()`. Agreed that some of those higher order functions can be detrimental to clarity. I left the error message in the calling function, because this function is scoped as a general utility, although one still waiting for a second caller.

R/S4.R
lawremi: Switched to use `as.character()`. It communicates the type and doesn't raise any eyebrows.

R/S4.R
```diff
@@ -149,6 +149,45 @@ S4_class_name <- function(x) {
...
+  if (generic_in_its_package) {
+    return(f@package)
+  }
+
+  # generic was defined for a function from a different package, like base
+  find_package_with_symbol(name, env, exclude = f@package) %||%
+    stop2(
+      sprintf(
+        "Failed to find originating package for S4 generic '%s' in imports.",
+        f@generic
```
hadley, line 172: I added a test to catch the missing value here.

hadley:
Merging because I'm hoping this will help with #662 and I don't want to force you to keep dealing with the merge conflicts I'm creating 😄

PR #664, opened 2026-05-29
Add `convert()` fallback to base `as.*` methods

### LINKED ISSUE #472: Default `convert()` methods for base classes.
Users may expect something like this to work:

```r
x <- 3
convert(x, class_character)
```
 
Concretely, I wonder if we should do:

```r
method(convert, list(class_any, class_character)) <- function(from, to, ...) as.character(from)
method(convert, list(class_any, class_integer)) <- function(from, to, ...) as.integer(from)
method(convert, list(class_any, class_logical)) <- function(from, to, ...) as.logical(from)
...
```

One thing to consider is that individual package authors may want different behavior. E.g., a strict whole-number check when casting double to integer
```r
method(convert, list(class_double, class_integer)) <- function(from, to, ...) {
  stopifnot("must be a whole number" = trunc(from) == from)
  as.integer(from)
}
```
We don't want different packages exporting different methods for this type of cast. We may want to dis-allow setting `convert(, to)` methods for base S3 classes for this reason.

hadley:
Fixes #472

### DIFF
--- a/tests/testthat/test-convert.R
+++ b/tests/testthat/test-convert.R
@@ -17,7 +17,12 @@ test_that("can register convert methods", {
   expect_equal(convert(obj, to = class_integer), "i")
 
   # Errors if none found
-  expect_snapshot(convert(obj, to = class_double), error = TRUE)
+  expect_snapshot(
+    convert(obj, to = class_double),
+    error = TRUE,
+    # for < 4.4.0
+    transform = \(x) gsub("'S4'", "'object'", x)
+  )
 })
 
 test_that("doesn't convert to subclass", {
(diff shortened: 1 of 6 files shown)


tests/testthat/test-convert.R
```diff
@@ -123,3 +123,30 @@ describe("fallback convert", {
...
+  method(convert, list(Txt, class_character)) <- function(from, to, ...) {
+    "custom"
+  }
+  expect_equal(convert(Txt("hi"), class_character), "custom")
+
+  # Upcasting to a base type strips the S7 wrapper rather than calling as.*()
+  method(convert, list(Txt, class_character)) <- NULL
+  obj <- convert(Txt("hi"), class_character)
+  expect_false(S7_inherits(obj))
+  expect_equal(attr(obj, "n"), NULL)
```
t-kalinowski:
```suggestion
  expect_null(attributes(obj))
```

t-kalinowski (approved):

PR #665, opened 2026-05-29
Use S7 namespace for base class constructors and validators

### LINKED ISSUE #553: should constructor, validator, getter, setter... be enclosed in baseenv()
This would make the S7 objects simpler, pointing only to named environment. My understanding is that a user fetching values from parent environments through lexical scoping would almost always be a mistake. An aspect I really like about S7 is that everything is contained in the object itself, the current enclosure system undermines this.

IIRC in R6 the enclosure is updated.

base classes currently use the enclosure could be adapted. For instance we can have : 

```r
class_logical <- list(
  class = "logical",
  constructor_name = "logical",
  constructor = (function(.data = logical(0)) {
    .data
  }) |>
    (`environment<-`)(baseenv()),
  validator = (function(object) {
    if (base_class(object) != "logical") {
      sprintf("Underlying data must be <logical> not <%s>", base_class(object))
    }
  }) |>
    (`environment<-`)(baseenv())
) |>
  structure(class = "S7_base_class")
```

Right now there is no code `class_logical` and friends are "different" in every session.

hadley:
This avoids capturing the local environment in the case of the validator, and the inconsistent use of the base environment in the case of the constructor. The validator and constructor for the S3 wrappers already used the S7 namespace.

Fixes #553

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # S7 (development version)
 
+* Base type wrappers like `class_integer` now define their constructor and validator in the S7 namespace. (#553).
 * `method<-` now accepts `NULL` to unregister an existing method, e.g. `method(foo, class_character) <- NULL` (#613).
 * `new_object()` no longer materialises ALTREP parent values (e.g. `seq_len()`), so constructing an S7 object that wraps a large compact integer sequence is now O(1) in memory instead of O(n) (@kschaubroeck, #607).
 * Internal changes to support R-devel (4.6) (#592, #593, #598, #600).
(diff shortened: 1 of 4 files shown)


t-kalinowski (approved):
I'm not sure I understand the motivation for the original issue, but this change seems harmless. I don't recall why we set the `baseenv()` parent here, though it's clear we went out of our way to do so. I think it might have been around the time we were working on the serialization/dput issues in base R, and this was somehow related.

In either case, this seems fine and simplifies the internals.

PR #667, opened 2026-05-29
Introduce `stop2()` helper

### LINKED ISSUE #646: Provide `stop_with_call()` helper
So that we can explicitly pass to helpers where we want to give a nice user facing call.

hadley:
With explicit `call` and `class` arguments. Then use it in lots of places. I didn't try to be exhaustive, but this is an overall improvement and we can chip away at the remaining issues over time.

Fixes #646


t-kalinowski (approved):
Much improved errors now!

PR #668, opened 2026-05-29
Final method registration refactoring

### LINKED ISSUE #455: Use `method<-` to register methods on S3 generics for base classes

hadley:
Fixes #455


R/method-register.R
```diff
@@ -21,7 +21,7 @@
 #'   or an [S4 generic][methods::setGeneric].
 #' @param signature A method signature.
 #'
-#'   For S7 generics that use single dispatch, this must be one of the
+#'   For single single dispatch generics, this must be one of the
```
t-kalinowski:
```suggestion
#'   For single-dispatch generics, this must be one of the
```

t-kalinowski (approved):

PR #670, opened 2026-05-30
Only use more specific convert methods

### LINKED ISSUE #429: convert() should be idempotent when `from` is an instance of `to`
If `from` is an instance of `to`, i.e., `identical(S7_class(from), to)` is `TRUE` (assuming `from` is an S7 object), then no dispatch should occur.

This is not (just) about efficiency. If there are `convert()` methods for downcasting to `to`, one of them will be called even if there is no need to downcast, potentially resulting in data being overwritten, or other weird things happening.

I'll also note that `methods::as()` is idempotent when `to` is `class_any`, but `class_any` is not allowed with `convert()`. Somewhat pedantic, but I thought it was worth mentioning.

hadley:
Fixes #429

@lawremi does this do what you expect?


lawremi (approved):
Thanks for fixing this, much appreciated.

R/convert.R
```diff
@@ -108,6 +117,29 @@ convert <- function(from, to, ...) {
...
+# inherits from `to`, only classes more specific than `to` are considered:
+# methods registered on `to` or its parents would downcast, potentially
+# overwriting data, when all that's wanted is an upcast (#429). Returns `NULL`
+# when `from` is already an instance of `to` signalling that conversion is
+# idempotent.
+convert_dispatch <- function(from, to) {
+  from_dispatch <- obj_dispatch(from)
+  to_class <- class_register(to)
+
+  for (i in seq_along(from_dispatch)) {
```
hadley, line 145: Actually is this correct? Shouldn't it only consider an exact match? (i.e. no inheritance at all?)

hadley: A little experimentation and iteration with claude has convinced me that this implementation is correct.

I've added more to the docs + examples to explain this.

PR #671, opened 2026-05-30
Correctly determine downcasts

### LINKED ISSUE #509: `convert()` automatically attempts incorrect sibling casts
Currently, `convert()` is described as automatically implementing downcasts to a descendant or upcasts to an ancestor class.* However, I was surprised to discover that it also attempts "sibling" casts; that is, casts to a class that is neither an ancestor nor a descendant, but which has a shared ancestor. Is this intended behavior?

Consider these classes, where `Base` has children `A` and `B`, and `B` has child `B_child`:

```r
Base = new_class("Base")

A = new_class(
  "A",
  parent = Base,
  properties = list(x = new_property(class_numeric, default = 1))
)

B = new_class("B", parent = Base)

B_child = new_class(
  "B_child",
  parent = B,
  properties = list(x = new_property(class_character, default = "hello"))
)
```

Per the documentation, I'd expect the following to work (and they do):

```r
convert(B_child(), Base)
## <Base>
convert(B(), B_child)
## <B_child>
##  @ x: chr "hello"
```

Surprisingly (to me), I can also automatically convert to a sibling; e.g. `B` to `A` even though neither is a subclass of the other:

```r
convert(B(), A)
## <A>
##  @ x: num 1
```

It will also attempt to convert `B_child` to `A`, though it will fail because their properties are not compatible:

```r
convert(B_child(), A)
## Error: <A> object properties are invalid:
## - @x must be <integer> or <double>, not <character>
```

Personally, I would consider either of these sibling casts to be an error, as it is not clear to me that one should expect sibling classes to share semantics under which an automatic conversion is likely to make sense. If one ever did want to do a sibling cast, it would be easy enough to do explicitly by going through the base class, a la `convert(convert(B_child(), Base), A)`.

I believe this is happening because `convert` is checking inheritance via `is_parent_instance()` here:

https://github.com/RConsortium/S7/blob/db5964265a88d114fb6de97361060979ad63b3d6/R/convert.R#L106

Which reports siblings as parents:

```r
S7:::is_parent_instance(B(), A)
## [1] TRUE
```

*Sidebar: I believe the documentation (and comments in source) of `convert()` currently uses "upcast" and "downcast" incorrectly: wherever I've encountered these, "upcast" indicates a cast going up the class tree (to an ancestor) and "downcast" indicates a cast going down the class tree (to a descendent); see e.g. [this wiki article](https://en.wikipedia.org/wiki/Downcasting).

hadley:
Fixes #509


R/convert.R
```diff
@@ -135,7 +135,7 @@ convert_up <- function(from, to, call = sys.call(-1L)) {
 }
 
 is_down_cast <- function(x, class) {
-  inherits(x, setdiff(class_dispatch(class), "S7_object"))
+  all(obj_dispatch(x) %in% class_dispatch(class))
```
t-kalinowski: I wonder if we should be even stricter here, checking length and order too, and not accommodating unrecognized additional parents, so we don’t false-positive on admittedly rare complex inheritance chains:

```r
x_classes <- obj_dispatch(x)
identical(x_classes, class_dispatch(class)[seq_along(x_classes)])
```

R/class-spec.R
```diff
@@ -343,5 +340,20 @@ obj_dispatch <- function(x) {
...
 # helpers -----------------------------------------------------------------
 
+# Does `child`'s dispatch extend `parent`'s? Subclassing only ever prepends
+# more specific classes, so `parent`'s classes must form the tail of `child`'s.
+# S7 wrappers of base/S3 types append "S7_object", which we ignore.
+class_dispatch_extends <- function(parent, child) {
+  parent <- drop_S7_object(parent)
+  child <- drop_S7_object(child)
+  n <- length(parent)
+  length(child) >= n && identical(child[length(child) - n + seq_len(n)], parent)
```
t-kalinowski, line 350:
```suggestion
  length(child) >= n && identical(child[seq.int(to = length(child), length.out = n)], parent)
```

t-kalinowski (approved):

R/class-spec.R
hadley: That's just as confusing to me 😆

PR #673, opened 2026-05-30
Automatically rename reserved properties

### LINKED ISSUE #579: Allow fallback method of storing properties for "forbidden" names

hadley:
Implemented in C for performance

Fixes #579

---

Benchmark from claude

## Ordinary-named properties (the common case)

| Operation                         | Before (min / median) | After (min / median) |
|-----------------------------------|-----------------------|----------------------|
| `Foo(x = 1, y = 2)` (construct)   | 66.7µs / 76.1µs       | 68.0µs / 77.7µs      |
| `f@x` (get)                       | 738ns / 902ns         | 779ns / 943ns        |
| `f@x <- 10` (set + validate)      | 38.0µs / 44.0µs       | 37.9µs / 44.0µs      |
| `prop<-(check = FALSE)` (raw set) | 328ns / 451ns         | 328ns / 410ns        |

All differences are within run-to-run noise.

## Special-named properties (after)

Class `Bar <- new_class("Bar", properties = list(names = class_character, y = class_numeric))`:

| Operation                             | min / median    |
|---------------------------------------|-----------------|
| `Bar(names = "a", y = 2)` (construct) | 62.6µs / 71.8µs |
| `b@names` (get)                       | 738ns / 902ns   |
| `prop<-(check = FALSE)` (raw set)     | 287ns / 410ns   |

The `"_"`-prefixed path is as fast as the ordinary path.


src/prop.c
```diff
@@ -102,6 +112,40 @@ SEXP extract_name(SEXP list, SEXP name_rchar) {
...
+  if (name_sym == R_DimSymbol)      return sym_u_dim;
+  if (name_sym == R_DimNamesSymbol) return sym_u_dimnames;
+  if (name_sym == R_ClassSymbol)    return sym_u_class;
+  if (name_sym == R_TspSymbol)      return sym_u_tsp;
+  if (name_sym == sym_comment)      return sym_u_comment;
+  if (name_sym == R_RowNamesSymbol) return sym_u_row_names;
+  return name_sym;
+}
+
+SEXP prop_storage_rename_(SEXP names) {
```
hadley, line 129: Implemented in C to avoid having to encode special names in C and R.

R/class.R
```diff
@@ -406,23 +408,21 @@ S7_class <- function(object) {
...
+    stop2("Properties can't be named \"...\".", call = call)
+  }
+
+  # Names with special C handlers in base R (names, class, ...) are stored under
+  # a "_"-prefixed attribute, see prop_storage_rename(). "_"-prefixed names are
+  # therefore reserved to avoid colliding with that storage.
+  reserved <- startsWith(nms, "_")
+  if (any(reserved)) {
+    msg <- sprintf(
+      "Properties can't start with \"_\": %s.",
```
t-kalinowski: This new restriction deserves a mention in the NEWS.md

t-kalinowski: A blanket restriction on a `_` prefix seems unnecessarily restrictive. I think it’s common enough in the wild to see JSON objects with a `_` prefix that this would be painful. For example, consider a serialized Python object with dunder fields. I think we might as well pick a longer, more unique prefix now. Maybe something like `_S7:::`?

src/prop.c
```diff
@@ -102,6 +112,40 @@ SEXP extract_name(SEXP list, SEXP name_rchar) {
...
+
+  R_xlen_t n = Rf_xlength(names);
+  SEXP out = PROTECT(Rf_allocVector(STRSXP, n));
+  for (R_xlen_t i = 0; i < n; i++) {
+    SEXP name_rchar = STRING_ELT(names, i);
+    if (name_rchar == NA_STRING) {
+      SET_STRING_ELT(out, i, name_rchar);
+    } else {
+      SEXP storage_sym = prop_storage_sym(Rf_installTrChar(name_rchar));
+      SET_STRING_ELT(out, i, PRINTNAME(storage_sym));
```
t-kalinowski, line 143: Microoptimization: seems wasteful to round-trip string -> sym -> string, to then just go back to `sym` again in `names<-`. I would probably add a C entry point for `set_prop_names` to avoid this.

hadley: I think we're always working with a named vector on the R side, not directly manipulating the attributes?

R/class.R
hadley: I think we could just drop the check and only document it. I don't think we want to make a more complicated prefix because I'd like to keep it aligned with #655.

t-kalinowski: I’m still concerned that restricting names like `__class__`, `_comment`, `_names`, and `_dim` will be a pain point. These are common enough in the wild when serializing JSON-ish data that it seems worth using something a little more unique.

Since this isn’t a user-facing feature, I think we have more flexibility in the internal naming scheme. And #655 is only a restriction on an ergonomic entry point, which can be worked around by calling `prop<-` directly. Enforcing the restriction here feels different, since it would more materially limit what S7 can be used for.

hadley: Right, I'm saying we could just drop the enforcement here. We don't have any reports of the more common names causing problems (except when creating new S7 classes), so I'm not that worried about this very speculative case.

t-kalinowski: We can narrow the enforcement to just the symbols we're using, but we'll still have an opportunity for a clash. I'm merely suggesting that instead of using symbols like `_names` as the attribute name, we use `_S7::names`, to reduce (but not eliminate) the opportunity, with essentially no change in code complexity.

The motivation being that if someone tries to implement a json marshalling/unmarshalling library ontop of S7, then we'll minimize the opportunity for unrepresentable data. E.g., though relatively rare, these object names aren't unheard of:

https://github.com/search?q=%22_comment%22%3A+language%3Ajson&type=code
https://github.com/search?q=%22_names%22%3A+language%3Ajson&type=code
https://github.com/search?q=%22_dim%22%3A+language%3Ajson&type=code

hadley: Yeah, I get that, I just don't buy that it's worth it. But I think we've discussed this enough and we're not going to come to consensus, so I'm going to proceed over your objections and deal with the consequences when they happen 😄

t-kalinowski (approved):
Other than the one comment, everything else looks good!

PR #675, opened 2026-06-01
Run `class_validate()` on properties

hadley:
* Fix bug in `class_POSIXct` thus revealed
* Fix bug in S4 validation thus revealed


t-kalinowski (approved):
Nice catch!

hadley:
All credit goes to claude 😆

PR #677, opened 2026-06-02
S7_class becomes a reserved attribute

lawremi:
This is hopefully an obvious change where the `S7_class` attribute is no longer allowed to be declared as a property. Just noticed this while I was working on the S4 support.

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -436,4 +436,14 @@ check_prop_names <- function(properties, call = sys.call(-1L)) {
     )
     stop2(msg, call = call)
   }
+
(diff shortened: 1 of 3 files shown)


hadley:
Maybe do after #673?

hadley:
Or maybe we should consider renaming this to `_S7_class` or `_class` so we put it in the "namespace" that we've declared to be internal?

lawremi:
No strong feelings, but if we're going to rename it, the sooner the better, to minimize the breakage of serialized instances.

PR #678, opened 2026-06-02
Misc S4 fixes

lawremi:
Place to compile S4 fixes not directly related to the S4 inheritance work. Will request review when there's reasonable confidence that there won't be another fix shortly forthcoming.


hadley:
Are these still relevant or can we close this now?

lawremi:
I have no specific reason to believe that they have become obsolete. Will resolve conflicts and look into it.

lawremi:
Actually, I remember that I had rebased the other branch on top of this, so this is totally redundant.

PR #681, opened 2026-06-04
Impement `S7_user_frame()` and `S7_generic_call()`

hadley:
These are not exactly symmetric, but I think they solve the common cases, and the implementation is general enough that we can expand in the future if we discover the need.

Fixes #596


t-kalinowski (changes requested):
I pushed a failing regression test for this case: `same-generic nested calls are not super redispatches`.

Right now we can false positive in detecting `super()` redispatch for an ordinary nested call to the same generic, e.g. a method calling `foo("inner")`, currently gets treated like `foo(super(...))` because the caller’s parent is the same generic. As a result, `S7_generic_call()` reports the outer call and `S7_user_frame()` returns the outer caller frame, instead of the inner generic call and the method frame that invoked it.

hadley:
Ooh, good catch!

R/method-call.R
```diff
@@ -0,0 +1,97 @@
...
+  # Walk past super() re-dispatches
+  while (is_super_dispatch(frame)) {
+    method <- parents[frame]
+    frame <- parents[method]
+  }
+
+  frame
+}
+
+# S7_dispatch() marks the generic's frame with `_dispatched_super` when it
```
hadley, line 142: I'm not sure I love this approach, but it feels safer than either trying to inspect the generic call for calls or `super()`, or evaluating each of the arguments to the generic to see if they have class `S7_super`.

t-kalinowski (dismissed):
Defining a sentinel value in the generic frame seems fine for now, until a more elegant solution emerges. However, with the current implementation, we walk past all super frames, even those for another method. I pushed a test (currently failing) covering the case of a method calling another method with super.

hadley:
@t-kalinowski do you want to just finish this off? It seems a bit pointless for you to make a failing test and then for me to tell claude to fix it 😆

t-kalinowski:
Thanks! Happy to. I pushed a change. Github won't let me request a review from the original author (you). Can you please take a look?

hadley:
I'm worried that the complexity here is a sign that the idea to go all the way back to the originating generic call is a bad idea. But it still seems useful, so what do you think about gating it behind an argument? So by default these functions return the immediate generic frame, but you can opt-in to navigating all the way back past any super calls?

PR #691, opened 2026-06-09
Check that overriden properties narrow the type specification

### LINKED ISSUE #352: Extending a parent class with inconsistent properties
Currently it's possible to define a class who's properties are inconsistent with its parent:

``` r
library(S7)

foo1 <- new_class("foo1", properties = list(x = class_integer))
foo2 <- new_class("foo2", parent = foo1, properties = list(x = class_character))
```

This doesn't error when you create the class, but it's not possible to construct it because there's no value that will satisfy the validator of both `foo1` and `foo2`:

``` r
foo2(x = 1L)
#> Error: <foo2> object properties are invalid:
#> - @x must be <character>, not <integer>
foo2(x = "x")
#> Error: <foo1> object properties are invalid:
#> - @x must be <integer>, not <character>
```

Should we make this a define-time error? In principle, I think it's ok for a child class to make a property more restrictive than the parent class, but there's no way to test for test for this in general (since the restriction may occur at the validator level).

hadley:
Fixes #352


t-kalinowski (approved):
I pushed some changes fixing some edge cases around unions with `NULL`. You may want to look over before merging:

https://diffshub.com/RConsortium/S7/compare/634c2fe366c7ad2b344159bedbb3101420af683f...8b14997a4764099b77d9822761a4a632a8ce939f

hadley:
LGTM. Thanks!

PR #692, opened 2026-06-10
Improve handling of abstract S3 classes

### LINKED ISSUE #686: ggplot2 fails to install with S7 main
With the current main branch installed (and with R trunk), I'm getting this when trying to install ggplot2 from CRAN:

```
** byte-compile and prepare package for lazy loading
Error in S7::new_object(x) : 
  `.parent` must be an instance of S3<gg>, not <list>.
Error: unable to load R code in package ‘ggplot2’
Execution halted
ERROR: lazy loading failed for package ‘ggplot2’
```

Perhaps we now have stricter checks and have found a bug in ggplot2? Or it's just something in my installation...

hadley:
* No longer error when constructing a class that inherits from an abstract S3 class
* Errors when upcasting to an abstract S3 class
* New test to confirm that `super()` works for an abstract S3 class.

No news bullet for the motivating issue since this was a bug introduced in the dev version.

Fixes #686

### DIFF
--- a/R/class.R
+++ b/R/class.R
@@ -267,6 +267,13 @@ check_can_inherit <- function(
 
 is_class <- function(x) inherits(x, "S7_class")
 
+# A class you can't supply an instance of: an abstract S7 class, or an S3 class
+# registered without a constructor (e.g. a marker class like "gg" or "POSIXt").
+class_is_abstract <- function(class) {
+  (is_class(class) && class@abstract) ||
+    (is_S3_class(class) && class$abstract)
+}
+
 check_parent <- function(parent, class, call = sys.call(-1L)) {
   parent_class <- class@parent
   if (is.null(parent_class)) {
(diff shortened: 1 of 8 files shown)


R/class.R
```diff
@@ -267,6 +267,13 @@ check_can_inherit <- function(
 
 is_class <- function(x) inherits(x, "S7_class")
 
+# A class you can't supply an instance of: an abstract S7 class, or an S3 class
+# registered without a constructor (e.g. a marker class like "gg" or "POSIXt").
+class_is_abstract <- function(class) {
+  (is_class(class) && class@abstract) ||
+    (is_S3_class(class) && class$abstract)
```
t-kalinowski: Should we use `isTRUE(class$abstract)` here, in case the class is missing the `$abstract` field? This could happen if people are using the new S7 with S7 objects that were constructed with older S7 (i.e., built packages or serialized class defs generated with current release S7).

hadley: I think that's pretty low likelihood given the relative rarity of abstract S3 classes. Although I should double check that this fixes the ggplot2 issue.

hadley: `isTRUE()` doesn't actually help because we need to work with S3 classes created by older S7. And while these are obviously rare in the scheme of things they do occur in ggplot2, hence this issue. 

I've added a fallback and verified that this now does actually fix the ggplot2 loading issue.

t-kalinowski (approved):

PR #694, opened 2026-06-10
Improve construction of subclasses

### LINKED ISSUE #467: Default value introduced in subclass does not propagate to constructor

### LINKED ISSUE #585: Child class property setters that mask parent class properties are unused at construction
Given a set of child-parent classes, where the child masks a parent property and adds a custom setter such as

```r
parent <- new_class(
  "parent",
  properties = list(
    "x" = class_integer,
    "y" = class_character,
    "z" = class_any
  )
)

child <- new_class(
  "child",
  parent = parent,
  properties = list(
    "z" = new_property(
      class_double,
        stop("hey! don't set that!")
        self
      }
    )
  )
)
```

Calling the child constructor does not ever hit the child property setter.

```r
child(z = 1)
```
```
<child>
 @ x: int(0)
 @ y: chr(0)
 @ z: num 1
```

I would expect that calling `child(z = 1)` would use the child property setter and throw the error, but it seems that the parent constructor is used then the value is type checked against the child property class without ever using the child setter.

hadley:
* `constructor_args()` no longer drops child properties that shadow parent properties, so they become constructor args with the child's default. Fixes #467.
* The generated body no longer passes declared properties to the parent constructor; they go to `new_object()` instead, which runs the child's setter and class check. Fixes #585.
* Read-only properties are removed from the constructor.
* Constructors for subclasses of abstract classes now include parent properties.


R/class.R
```diff
@@ -130,21 +130,20 @@ new_class <- function(
     }
   }
 
-  # Combine properties from parent, overriding as needed
```
hadley, line 134: This manipulation now happens in `new_constructor()` so it has both the parent and overridden properties.

tests/testthat/_snaps/constructor.md
```diff
@@ -55,18 +55,20 @@
       foo1 <- new_class("foo1", abstract = TRUE, properties = list(x = class_double))
       new_constructor(foo1, list())
     Output
-      function () 
+      function (x = numeric(0))
```
hadley, line 58: This seems like a clear bug that we never noticed before.

R/constructor.R
```diff
@@ -29,30 +37,43 @@ new_constructor <- function(
...
   # ensure default value for `...` is empty
   if ("..." %in% names(args)) {
     args[names(args) == "..."] <- list(quote(expr = ))
   }
 
-  parent_args <- as_names(names(arg_info$parent), named = TRUE)
+  # Declared properties (including overrides of parent properties) are set by
+  # new_object() so that their defaults and setters are respected;
+  # only the remaining arguments are passed to the parent constructor.
+  parent_arg_nms <- setdiff(names(arg_info$parent), names2(properties))
```
t-kalinowski: This implicitly requires that overridden properties are optional in the parent. 
E.g., with this approach, this fails:

``` r
  devtools::load_all()
#> ℹ Loading S7
  foo <- new_class(
    "foo",
    properties = list(
      x = new_property(class_numeric, default = quote(stop("need x")))
    )
  )
  foo2 <- new_class(
    "foo2",
    parent = foo,
    properties = list(
      x = new_property(class_numeric, default = 2)
    )
  )
  
  `attributes<-`(foo2, NULL)
#> function (x = 2) 
#> S7::new_object(foo(), x = x)
#> <environment: 0x7698baa50>
  
  foo2()
#> Error in `foo()`:
#> ! need x
  foo2(x = 3)
#> Error in `foo()`:
#> ! need x
```

Is this what we want?

hadley: I agree that doesn't seem correct. OTOH I'm not sure if we fully support required properties — my mental mode of S7 was that you can always construct an object with no arguments, because that's such a useful way to generate a prototype. But it's possible that thinking is centred in vctrs and doesn't make sense in general.

If we do want to support required properties I don't see how to square that with respecting setters. For required properties to work you need to pass the property to the parent and to respect the child setter, you need to pass it to the child. Maybe we have to pass the property to both the child and the parent? 

Maybe you could pass the property to both parent and child, but then you'd call the setter twice, which might be confusing (although it does mean that the child definitely can't weaken any checks that the parent does).

t-kalinowski: I think the best approach is to pass it to both. This will have a performance impact unfortunately, but hopefully it's minor, and the benefits we get in correctness, enforcement, and simplicity are worth it.

hadley: Seems to work!

t-kalinowski (approved):
I pushed a few commits that try to make overridden parent properties behave more consistently in default constructors.

The implementation is a little hard to follow because it has to decide separately which arguments belong in the child constructor, which should be forwarded to the parent constructor, and which should be skipped. The tests are probably the clearest description of what I was aiming for here; each added test failed without the corresponding patch.

In short, the tests cover:

- subclass defaults replacing parent defaults
- mandatory parent properties being satisfiable from subclass defaults
- overridden properties still being forwarded through parent `...`
- subclass setters running during construction
- parent validators rerunning when inherited properties are reset
- read-only overrides being removed from the constructor without removing unrelated parent args
- incompatible dynamic settable overrides not being forwarded to the parent

Please take a close look. I’m not certain this is the simplest implementation, but I think the behavior in the tests is worth considering.

https://diffshub.com/RConsortium/S7/compare/c254152d88316466ccf0a5979d316c5309bb05da...b64d8e6548e369c4c0b56383593fd48ba76da9d5

R/constructor.R
```diff
@@ -29,30 +37,71 @@ new_constructor <- function(
...
+    args <- args[setdiff(names2(args), read_only_override_nms)]
+  }
+
   # ensure default value for `...` is empty
   if ("..." %in% names(args)) {
     args[names(args) == "..."] <- list(quote(expr = ))
   }
 
-  parent_args <- as_names(names(arg_info$parent), named = TRUE)
+  # Overridden properties are passed to both the parent and child.
```
hadley: This comment has now drifted from it's related implementation, and I think your new code below would benefit from a brief explanatory comment or too (and maybe some helpers?).

t-kalinowski: Done. ✅ I moved much of the logic out into a helper with informative names for each stage.

PR #696, opened 2026-06-11
Warn if property default is not as expected

### LINKED ISSUE #541: Codoc mismatch issue, revisited
I am having an issue similar to https://github.com/RConsortium/S7/issues/535. I've defined a class like this:

```
#' `cohort_element` Class Definition
#'
#' @description Represents the `cohort_element` class.
#'
#' @param visit_occurrences Visit occurrences by type. Can be 1 or 2 (default: `c(any = 1L)`).
#' @param visit_gaps Gaps between visits by type (default: `c(any = 0L)`).
#'
#' @section Validator:
#' Ensures visit_occurrence and visit_gaps are mutually consistent, and all inputs are allowed values
#'
#' @examples
#' \dontrun{
#' cohort_element()
#' }
#'
#' @export
cohort_element <- S7::new_class(
  name = "cohort_element",
  properties = list(
    visit_occurrences = S7::new_property(
      S7::class_integer,
      default = c(any = 1L),
      validator = function(value) {
        if (!all(names(value) %in% c("any", "inpatient", "outpatient", "emergency"))) {
          "must be named integers with valid visit types"
        } else if (!all(value %in% 1:2)) {
          "must be 1 or 2 for each element"
        }
      }
    ),
    visit_gaps = S7::new_property(
      S7::class_integer,
      default = c(any = 0L),
      validator = function(value) {
        if (!all(names(value) %in% c("any", "inpatient", "outpatient", "emergency"))) {
          "must be named integers with valid gap types"
        }
      }
  )
  ),
  validator = function(self) {
    if (length(self@visit_occurrences) != length(self@visit_gaps) || !all(sort(names(self@visit_occurrences)) == sort(names(self@visit_gaps)))) {
      "@visit_occurrences and @visit_gaps must have the same length and names"
    }
  }
)

```

When I run `devtools::check()`, it gives this warning:
```
  Codoc mismatches from Rd file 'cohort_element.Rd':
  cohort_element
    Code: function(visit_occurrences = c(any = 1L), visit_gaps = c(any = 0L))
    Docs: function(visit_occurrences = c(any = 1L), visit_gaps = c(any = 0L))
```

It only appears to occur when I pass a vector as the default values.  When I just make it an unnamed scalar I don't get the warning. Any advice?

hadley:
Fixes #541


t-kalinowski (approved):
This is a pragmatic solution. On a longer timescale, it would be great if we could fix this in `R CMD check` instead...

hadley:
@t-kalinowski I dropped the withr dep on the assumption that was an accident.

t-kalinowski:
Oh, I assumed it was safe to use since it's as a dependency of testthat.

PR #698, opened 2026-06-12
Implement `S7_on_unload()`

### LINKED ISSUE #316: Unregister methods on unload
`methods_register()` also needs add a hook on package unload to remove all methods, and reset any unused hooks.

hadley:
Perform the opposite role as `S7_on_load()`. Fixes #316.

NB: for a full fix, this requires pkgload 1.5.3 which now actually calls the unload hooks.


hadley:
@t-kalinowski this doesn't need a full review, but I've gotten stuck in the weeds of refactoring and I think it would be useful to get some high level feedback before I go too far.

t-kalinowski (commented):
It looks good to me so far, I like this approach.

hadley:
I'll do another round of refactoring as part of #660, but the tests are much easier to understand now.

t-kalinowski (approved):
I reviewed this and think the PR is good to go. I pushed a few additions while looking through unload cleanup. The main change is that `S7_on_unload()` now only unregisters an external S7 method if the unloading package still owns the active method.

There are a few supporting changes for that: package-registered S7 methods now carry owner metadata, unload skips external generics when the upstream package is unavailable or too old for the declared minimum version, and hook cleanup still removes stale hooks after an external method has been deleted. I added regression tests for those cases and updated NEWS/docs to match.

Please take a close look at the additions. If you think some of this is more edge-case handling than we want in this PR, I’m fine with dropping it. 

https://diffshub.com/RConsortium/S7/compare/09dfb7aabce6b4975d0b3238282b9aa3bb1bf170...d6567f1632d0e89774b41731ee501ddd0dad6313

hadley:
Yeah, I'm not sure I would have gone that far, but it makes sense, and no need to undo it 😄

PR #701, opened 2026-06-15
Use `test_that()` everywhere

### LINKED ISSUE #629: Use `test_that()` consistently

hadley:
Fixes #629


hadley:
@t-kalinowski just an FYI that I've moved to `test_that()` everywhere. I didn't think it was worth having you review this, but I asked a clean claude code to verify that every `it()` test on main was preserved.

PR #702, opened 2026-06-15
Use `local_s4_class()` + `:=` in more places

### LINKED ISSUE #700: Boost `local_s4_class()`

hadley:
Since it automatically cleans up the class. This revealed a small bug in `:=`, which I think is worth fixing (and it's an easy fix if we assume that `:=` eventually ends up in base R).

Fixes #700


t-kalinowski (approved):
LGTM! I pushed one minor fix to one of the tests.

hadley:
Doh, thanks. Merged: the build failures look like a transient bioc problem.

PR #703, opened 2026-06-15
Implement `convert_lazy()`

### LINKED ISSUE #428: non-strict convert()
It would be nice to have the equivalent of `methods::as(strict = FALSE)`, which ensures that `from` inherits from `to` but does not force `from` to be an _instance of_ `to` (stripping any extra properties). 

Maybe this is a different function, like `convert_lazy()` or `ensure_inherits()`.

hadley:
Fixes #428


lawremi (approved):
Looks good.

PR #704, opened 2026-06-15
Add intersect example

### LINKED ISSUE #419: Consider example of multiple dispatch with `intersect(x, y)`
Could be added to this section
https://rconsortium.github.io/S7/articles/generics-methods.html#multiple-dispatch

See also this thread
https://fosstodon.org/@jonocarroll/112913640978442903

`intersect(x, y)` is not a generic function in base R. We introduced the {generics} R package to turn it into a generic, so we could write methods on them in classes throughout the tidyverse - but the generic only dispatches on `x` and then it is up to the method to ensure that `y` is the right class.

It might be interesting to implement `intersect(x, y)` using multiple dispatch as a motivating example for that feature. I imagine that 99.99% of the time, you will end up only implementing a method for the case where `x` and `y` are the same type, i.e. for when they are both `Date`. But being able to express this in the type system rather than having the method confirm that `y` is the right type seems nice.

hadley:
Fixes #419

### DIFF
--- a/vignettes/generics-methods.Rmd
+++ b/vignettes/generics-methods.Rmd
@@ -294,3 +294,52 @@ There are two special classes that become particularly useful with multiple disp
 
 -   `class_any()` will match any class
 -   `class_missing()` will match a missing argument (i.e. not `NA`, but an argument that was not supplied)
+
+### A real example
+
+Let's explore the idea with a real function, `intersect()`, which returns the elements common to two vectors.
+Base R's `intersect()` isn't a generic, so it doesn't know how to combine vectors with custom classes.
+It would be nice if it were a double-dispatch generic, because most of the time you only want to intersect two vectors of the same type, so you only need to implement methods where `x` and `y` have the same class:
+
+```{r}
+intersect := new_generic(c("x", "y"))
+
+method(intersect, list(class_Date, class_Date)) <- function(x, y) {
+  .Date(base::intersect(unclass(x), unclass(y)))
+}
+method(intersect, list(class_POSIXct, class_POSIXct)) <- function(x, y) {
+  .POSIXct(base::intersect(unclass(x), unclass(y)), tz = attr(x, "tzone"))
+}
+```
+
+This implementation means we get the behavior we want when an intersect method exists:
+
+```{r}
+weekdays <- as.Date(c("2026-06-15", "2026-06-16", "2026-06-17"))
+holidays <- as.Date(c("2026-06-17", "2026-07-04"))
+intersect(weekdays, holidays)
+```
+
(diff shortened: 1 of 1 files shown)


vignettes/generics-methods.Rmd
```diff
@@ -294,3 +294,52 @@ There are two special classes that become particularly useful with multiple disp
...
+```{r, error = TRUE}
+intersect(weekdays, 1:3)
+```
+
+Occasionally you _do_ want to handle a mix of types, and this is also easy with a double-dispatch generic.
+For example, we could intersect a date with a date-time by promoting the date to a date-time and reusing the method above:
+
+```{r}
+method(intersect, list(class_Date, class_POSIXct)) <- function(x, y) {
+  intersect(as.POSIXct(x, tz = attr(y, "tzone")), y)
```
hadley, line 334: Hmmm, this now seems wrong since we're wandering into convert territory where you probably don't want to apply superclass methods in this way.

DavisVaughan: Somehow I would feel better if you had extracted

```r
posixct_intersect <- function(x, y) {
  .POSIXct(base::intersect(unclass(x), unclass(y)), tz = attr(x, "tzone"))
}
```

and used that in all 3 places (and had an equivalent `date_intersect()`).

That would at least avoid the second level of dispatch in favor of a static function.

vignettes/generics-methods.Rmd
```diff
@@ -294,3 +294,52 @@ There are two special classes that become particularly useful with multiple disp
...
+```
+
+And a nice error when it doesn't:
+
+```{r, error = TRUE}
+intersect(weekdays, 1:3)
+```
+
+Occasionally you _do_ want to handle a mix of types, and this is also easy with a double-dispatch generic.
+For example, we could intersect a date with a date-time by promoting the date to a date-time and reusing the method above:
```
DavisVaughan, line 330: I would go a little farther and say that you should be pretty careful about adding mixed type methods, and you should only do it for types that are very closely tied together (like integer/double). Honestly even date and date-time is a stretch imo (because date has no implied time zone).

To me this is exactly the same as adding `ptype2` methods for your class - i.e. it's performing automatic conversion under the hood.

We've learned over the years that you really shouldn't be adding `ptype2` methods beyond self-self most of the time.

DavisVaughan (approved):

hadley:
@DavisVaughan overall, this makes me feel like `intersect()` isn't a great example for double-dispatch, it just adds too many edge cases. They're important, but I don't think they should be introduced here.

DavisVaughan:
Yea, makes sense to me

PR #706, opened 2026-06-17
Agents update

hadley:
@t-kalinowski this mostly comes from upstream but if you spot anything you think could be improved, I can apply in usethis.


t-kalinowski (approved):
Nice!

PR #709, opened 2026-06-19
Implement wrappers for condition classes

### LINKED ISSUE #586: Feature Request: add `class_error`, `class_condition`, `class_warning` base classes
It would be great if these were available, so it would be easier to use S7 to implement custom error classes.

hadley:
Fixes #586


sebffischer:
I like it. Although this is rather minimal and does not provide a lot of functionality itself, I think condition objects are fundamental to the language, so I personally think they do deserve base classes. 

I also like what you said in the issue:

> OTOH maybe this could be a place where we could provide more details for people who do want to extend condition classes for their own purposes?

For completeness sake, `condition_message` should maybe also be included here.

hadley:
@t-kalinowski worth reading the last comment in #586 since I'm pretty ambivalent about this PR.

t-kalinowski (approved):
I think this PR strikes a nice balance: it builds a foundation users can extend without adding much complexity or surface area to S7. However, I’m not fully seeing the motivating use case yet, so overall, I’m ambivalent about this change.

hadley:
I'm still pretty ambivalent about this change, so I think that means it's not worth it. It's not a big lift, but I also don't think it's the right time for S7 to start providing wrappers for all base S7 objects, and the current implementation of these classes doesn't provide much in the way of new functionality. But if it's useful to you, @sebffischer, you should feel free to copy this code into your package, and we'll reconsider in the future, when the foundations of S7 are more fixed and stable.

PR #713, opened 2026-06-22
Require R 4.2.0

hadley:
Thus supporting five version of R (current + four previous)

@lawremi @t-kalinowski any concerns?


t-kalinowski (approved):
None from me

lawremi (approved):

Issue #143, opened 2022-01-31
Evolution vignette


hadley:
Need a vignette that systematically works through ways that classes and generics might change:

* Generic adds an argument
* Generic removes an argument
* Generic changes package
* Class adds a property (incl. potential to clash with subclasses)
* Class removes a property (incl. deprecation with active property)

This will be useful for future users, and will force us to ensure that we've fully thought through the key cases.

lawremi:
Let's also include some guidance around how to design classes and generics so as to mitigate the impact of any changes. For example, limiting the number of arguments to the generic and using composition to isolate property changes in helper classes. 

Based on past experience, these situations arise rarely, and when they do, the best solution is just to work with downstream dependencies to update them. The most important thing to get right is detecting conflicts and inconsistencies when they occur.

hadley:
@lawremi I suspect there's a different between BioC and CRAN here — when evolving an interface of a CRAN package, it's really nice to be able to avoid requiring simultaneous package releases. So while most of the time you do end up working with the downstream maintainers to fix the problem, it's nice to not have R CMD check fail while the packages are slightly out of sync.

hadley:
One last note: when you change a `generic()` it's nice to be able to still successfully source the whole package so that you can fix methods one-by-one, rather than having to update them all before your package re-loads successfully.

PR #717, opened 2026-06-24
Pass `...` to parent constructor

### LINKED ISSUE #317: Extending a class from another package

### LINKED ISSUE #609: Parent class constructor defaults are not safely scoped in child constructors

hadley:
There are two cases where it's not safe to re-use the deafults from the parent in a subclass constructor:

* When the parent comes from another package, because the specification might change between class construction and object instantaition. Fixes #317.
* When the parent constructor is custom, because we don't know anything about its environment. Fixes #609.

This does not make external parent classes entirely safe because it's possible the parent can make backward incompatible changes leading to run-time errors. But I think we believe that's outside of the scope of S7, and something that you'd discover when running your revdep checks prior to CRAN submission.


hadley:
The evolution lab I'm building for #143 turned up a gap here: making an upstream class abstract now breaks **construction of downstream subclasses**, because the forwarding constructor calls the parent constructor at run time.

```r
# pkgA
Foo <- new_class("Foo", package = "pkgA",
  properties = list(size = class_double), abstract = TRUE)

# pkgB
AFoo <- new_external_class("pkgA", "Foo")
Baz <- new_class("Baz", package = "pkgB", parent = AFoo,
  properties = list(y = class_double))

Baz(size = 1, y = 2)
#> Error in new_object(S7_object(), size = size) :
#>   Can't construct an object from abstract class <Foo>.
```

The inline path never calls the parent constructor, so local subclasses of abstract classes are fine. But with `constructor_forward()`, an abstract external parent can never be subclassed at all — and an upstream package making an existing class abstract becomes an immediate hard break for downstream subclasses (verified in the `class-make-abstract` scenario of the lab).

Maybe an abstract class's constructor needs to be callable when it's building the parent part of a subclass (e.g. an internal flag that `new_object()` checks), or the forwarding constructor needs a different strategy for abstract parents?

R/constructor.R
```diff
@@ -116,6 +135,76 @@ new_constructor <- function(
...
+constructor_forward <- function(parent, properties, envir, package) {
+  # Work out how to refer to and call the parent's constructor:
+  # * `ref`: passed to `new_call()` to build the parent call (a name, or a
+  #   `package`/`name` pair for an external class in another package).
+  # * `name`/`fun`: the constructor to embed in the child's environment (NULL
+  #   for external classes, which are resolved dynamically via `pkg::name`).
+  if (is_external_class(parent)) {
+    resolved <- resolve_external_class_req(parent, package)
+    if (identical(package, parent$package)) {
+      ref <- parent$name
```
t-kalinowski, line 159: I spent some time 🤔 on this branch and whether it’s necessary. I ended up adding a test to make sure it’s exercised. I still can’t really come up with a realistic user need this addresses, but it seems harmless to leave in.

t-kalinowski (approved):
This is a tricky PR, but covers all the edge-cases I can think of well. Nice!

R/constructor.R
hadley: Yeah, agreed.

PR #718, opened 2026-06-24
Keep S7's := ahead of rlang and data.table

### LINKED ISSUE #697: Fix rlang/data.table conflicts
```
library(ellmer)
#> Warning: replacing previous import 'S7:::=' by 'rlang:::=' when loading
#> 'ellmer'
```

t-kalinowski:
Fixes #697.

S7, rlang, and data.table all export `:=`. Before this change, the version found on the search path depended on the order in which the packages were attached, and R could print a masking message.

This PR makes S7's `:=` win in either order. R no longer prints an attach-time message saying that one package's `:=` masks another's. Messages about other masked objects are still shown.

When rlang or data.table is already attached, S7 changes only the `:=` binding that the package added to the search path. It does not change the package namespace, so explicit access with `` rlang::`:=` `` or `` data.table::`:=` `` is unchanged. The original search-path binding is restored if S7 is detached.

The package vignette also shows how to import `:=` from S7 when a package imports rlang or data.table.

## Testing

- The full test suite passes.
- The pkgdown check passes.
- R CMD check completes with no errors or warnings; the two existing notes remain.
- Both attachment orders were tested on R 4.2.3 and R 4.3.3.

### DIFF
--- a/R/compatibility.R
+++ b/R/compatibility.R
@@ -44,7 +44,7 @@ activate_bind_compatibility <- function() {
     return(invisible())
   }
 
-  for (package in packages_exporting_bind()) {
+  for (package in c("data.table", "rlang")) {
     rule <- conflictRules(package)
     conflictRules(
       package,
@@ -54,22 +54,6 @@ activate_bind_compatibility <- function() {
   }
 
   invisible()
-}
-
-packages_exporting_bind <- function(lib.loc = .libPaths()) {
-  paths <- unlist(
-    lapply(lib.loc, list.dirs, recursive = FALSE, full.names = TRUE),
-    use.names = FALSE
-  )
-  paths <- paths[file.exists(file.path(paths, "Meta", "nsInfo.rds"))]
-
-  exports_bind <- vapply(
-    file.path(paths, "Meta", "nsInfo.rds"),
-    function(path) ":=" %in% readRDS(path)$exports,
-    logical(1)
-  )
-
-  setdiff(unique(basename(paths)[exports_bind]), "S7")
 }
 
 #' @aliases @
--- a/R/compatibility.R
+++ b/R/compatibility.R
@@ -7,6 +7,71 @@ activate_backward_compatiblility <- function() {
   invisible()
 }
 
+activate_attach_compatibility <- function(pkgname) {
+  if (getRversion() >= "4.3.0" && !search_has_bind_conflict(pkgname)) {
+    return(invisible())
+  }
+
+  env <- as.environment(paste0("package:", pkgname))
+  env[[".conflicts.OK"]] <- TRUE
+  invisible()
+}
+
+search_has_bind_conflict <- function(pkgname) {
+  pkg <- paste0("package:", pkgname)
+  env <- as.environment(pkg)
+  bind <- env[[":="]]
+  where <- setdiff(search(), c(pkg, "Autoloads", "CheckExEnv"))
+
+  for (pos in where) {
+    other <- as.environment(pos)
+    if (!exists(":=", envir = other, inherits = FALSE)) {
+      next
+    }
+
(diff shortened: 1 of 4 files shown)


R/compatibility.R
```diff
@@ -7,6 +7,55 @@ activate_backward_compatiblility <- function() {
   invisible()
 }
 
+activate_attach_compatibility <- function(pkgname) {
+  if (getRversion() >= "4.3.0" && !search_has_bind_conflict(pkgname)) {
+    return(invisible())
+  }
+
+  env <- as.environment(paste0("package:", pkgname))
+  env[[".conflicts.OK"]] <- TRUE
```
hadley: I think we should only ever apply this to S7 itself, and I think we can move it up to `.onAttach()` so we use it unconditionally when `conflictRules()` is not available.

R/compatibility.R
```diff
@@ -7,6 +7,55 @@ activate_backward_compatiblility <- function() {
...
+  }
+
+  FALSE
+}
+
+activate_bind_compatibility <- function() {
+  conflictRules <- get0("conflictRules", envir = baseenv(), inherits = FALSE)
+  if (is.null(conflictRules)) {
+    return(invisible())
+  }
```
hadley: I think it would be clearer to use an explicit version here

R/compatibility.R
```diff
@@ -7,13 +7,71 @@ activate_backward_compatiblility <- function() {
...
-  if (getRversion() >= "4.3.0" && !search_has_bind_conflict(pkgname)) {
+# The masking S7 performs deliberately: `@` over base (on R < 4.3.0, where
+# S7 exports its own `@`), and `:=` over rlang and data.table.
+s7_expected_masks <- list(
+  base = "@",
+  rlang = ":=",
+  data.table = ":="
+)
+
+# Re-emit the conflict report that library() would have produced (see
```
hadley: Given we can switch to `conflictRules()`, I think we could drop this.

t-kalinowski: `conflictRules()` only solves the case when rlang or data.table are loaded *after* S7. We also need a solution for when they're loaded before S7.

The `.conflicts.OK` binding is the only lever within our control there, and unfortunately, it's quite coarse: it only lets us opt into disabling *all* warnings. In which case, if there are other conflict warnings that we don't want to suppress (meaning, not `:=`), then it's our responsibility to report them.

R/zzz.R
```diff
@@ -132,12 +132,19 @@ methods::setOldClass(c("S7_method", "function", "S7_object"))
...
 
 .onAttach <- function(libname, pkgname) {
-  env <- as.environment(paste0("package:", pkgname))
+  activate_bind_compatibility()
+
   if (getRversion() < "4.3.0") {
-    env[[".conflicts.OK"]] <- TRUE
+    # S7_at already supplies @ without conflicting with base::@.
+    env <- as.environment(paste0("package:", pkgname))
+    rm(list = "@", envir = env)
```
hadley, line 140: Cunning!

vignettes/packages.Rmd
```diff
@@ -26,6 +26,16 @@ library(S7)
 
 First, add `S7` to the `Imports` field of your `DESCRIPTION`. We then recommend importing all S7 functions into your package `NAMESPACE` with `import(S7)`, or, if you're using roxygen2, `@import S7`.
 
+If your package also imports rlang or data.table, import `:=` from only S7. Otherwise, R warns that one import replaces the other when it loads your package. For example, to import all of rlang except `:=` with roxygen2:
+
+``` r
+#' @import S7
+#' @rawNamespace import(rlang, except = ":=")
```
hadley:
```suggestion
#' @import rlang, except = ":="
```

hadley (approved):

PR #719, opened 2026-06-24
Tag S7-generated constructors with `new_S7_constructor()`

hadley:
This allows us to elminate the fragile `is_S3_stub_constructor()`  and is useful infrastructure for #717


R/constructor.R
```diff
@@ -118,6 +123,22 @@ constructor_args <- function(
 
 # helpers -----------------------------------------------------------------
 
+# Was this constructor generated by S7 or supplied by the user?
+is_default_constructor <- function(constructor) {
+  inherits(constructor, "S7_constructor")
+}
+
+#' @export
+print.S7_constructor <- function(x, ...) {
```
t-kalinowski, line 132: I'm surprised there are no snapshot diffs as part of this PR. I would expect to see some change in snapshots anytime we change printing.

hadley: The point of this change is to make sure that there aren't any diffs 😄

PR #720, opened 2026-06-25
Polish S4-backed S7 edge cases

t-kalinowski:
## Summary

I am far from an S4 expert, and this branch is heavily LLM-assisted. It addresses correctness issues flagged by automated/agent-assisted code review of the original branch for #659.

One failure mode of that review process is that it can pull in extra complexity: fixes for edge cases that are technically defensible, but may not be worth supporting in practice. Without an explicit decision in the code or tests about what S7 should and should not support, those fixes can look like scope creep. I am too far removed from the S4 ecosystem to make those judgment calls confidently.

@lawremi, could you look over this branch, especially the tests, and help decide which cases matter and which ones we should drop?

The branch currently contains changes for cases where:
- S4 data-part classes lose attributes or `.S3Class` metadata when used as S7 parents.
- S7 subclasses of S4 classes lose property metadata during construction, validation, conversion, or upcasting.
- S4 slot definitions, including renamed slots and package-qualified slot classes, are not checked against the corresponding S7 properties.
- S4 unions and virtual classes are not matched or initialized with the expected package-qualified class identity.
- S4 parent classes with implicit data parts or conflicting slot accessors are accepted.
- S3/S4 method registration misses dispatch markers for S4-backed S7 classes.
- Installed S4 classes expose stale registration metadata when reused by S7 classes.

It also adjusts S4 registration and validation helpers around metadata refresh and validity proxy reuse, and adds regression tests for construction, inheritance, conversion, validation, data-part handling, method registration, and installed-package compatibility.

One thing worth noting is that even though there are already a lot of changes on this branch that build on top of the already large #659, automated code review is still flagging issues and has not yet come back clean. It might be better to split #659 (and some portions of this PR we deem are worth keeping) into a series of smaller, more incremental stacked PRs. That would make each step easier to reason about and review, both for humans and agents.


lawremi:
Thanks for doing this @t-kalinowski. It's very helpful. 

What this shows is that fully supporting every edge case would likely double the size and complexity of S7, which probably wouldn't be worth it. The goal is to provide a migration path from S4 to S7, rather than to achieve 100% compatibility between two different ecosystems, one of which we hope to eventually deprecate.

I agree with you that we should take an incremental approach here. Since nothing here seems to invalidate the approach taken in #659, let's merge that, and then we can address these issues in some prioritized way.

PR #721, opened 2026-07-07
high priorty S4 compatibility fixes

lawremi:
This PR addresses several high-priority compatibility issues found in #720:

- Registers S4 methods in the caller/package environment when S7 methods are registered for S4-derived classes on base/internal generics, fixing cases like `unlist()`.
- Allows generated constructors for S4 parents to route parent slot arguments through `methods::initialize()`. Parent slots were being dropped during construction.
  - Side fix: `new_object()` correctly detects its evaluation frame even when lazily evaluated as part of another call.
- Makes the `initialize()` method on S4-derived classes delegate to any non-default parent `initialize()` method, so as not to assume that the custom method has the same interface and semantics as the default method. When S4 is not extending S7, we could just not register our `initialize()` method when the default is overridden by a parent, but this seems like the simplest solution.
- Rejects S7 properties with custom getters or setters when they are overriding an S4 slot, since the slot layer will violate any encapsulation.
- Teaches `class_extends()` to correctly handle S4/S7 inheritance cases.

### DIFF
--- a/R/class-spec.R
+++ b/R/class-spec.R
@@ -412,6 +412,8 @@ class_extends_S4_name <- function(class) {
     class@className
   } else if (is_class(class)) {
     S7_class_name(class)
+  } else {
+     NULL
   }
 }
 
--- a/R/class-spec.R
+++ b/R/class-spec.R
@@ -413,7 +413,7 @@ class_extends_S4_name <- function(class) {
   } else if (is_class(class)) {
     S7_class_name(class)
   } else {
-     NULL
+    NULL
   }
 }
 
--- a/R/class-spec.R
+++ b/R/class-spec.R
@@ -396,12 +396,22 @@ class_extends <- function(child, parent) {
     parent <- resolve_external_class_req(parent)
     class_extends(child, parent)
   } else if (is_S4_class(child) || is_S4_class(parent)) {
-    is_S4_class(child) &&
-      is_S4_class(parent) &&
-      methods::extends(child@className, parent@className)
+    child <- class_extends_S4_name(child)
+    parent <- class_extends_S4_name(parent)
+    !is.null(child) &&
+      !is.null(parent) &&
+      methods::extends(child, parent)
   } else {
     # handle S7, S3, and base types.
     class_dispatch_extends(class_dispatch(parent), class_dispatch(child))
+  }
+}
+
+class_extends_S4_name <- function(class) {
+  if (is_S4_class(class)) {
+    class@className
+  } else if (is_class(class)) {
+    S7_class_name(class)
   }
 }
 
--- a/R/class.R
+++ b/R/class.R
@@ -347,7 +347,7 @@ check_parent <- function(parent, class, call = sys.call(-1L)) {
 #' @rdname new_class
 #' @export
 new_object <- function(`_parent`, ...) {
-  class <- sys.function(-1)
+  class <- sys.function(sys.parent())
   if (!inherits(class, "S7_class")) {
     stop2("`new_object()` must be called from within a constructor.")
   }
--- a/R/class.R
+++ b/R/class.R
@@ -485,10 +485,7 @@ check_prop_overrides <- function(
   call = sys.call(-1L)
 ) {
   overridden <- intersect(names(child_props), names(parent_props))
-  parent_S4 <- if (is_S4_class(parent)) parent else S4_ancestor(parent)
-  if (!is.null(parent_S4)) {
-    check_S4_slot_overrides(child_props, parent_S4, call = call)
-  }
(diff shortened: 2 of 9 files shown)


R/class-spec.R
```diff
@@ -396,15 +396,25 @@ class_extends <- function(child, parent) {
...
     class_dispatch_extends(class_dispatch(parent), class_dispatch(child))
   }
 }
 
+class_extends_S4_name <- function(class) {
+  if (is_S4_class(class)) {
+    class@className
+  } else if (is_class(class)) {
+    S7_class_name(class)
+  }
```
hadley, line 417:
```suggestion
  } else {
     NULL
  }
```

Just to be explicit?

R/class.R
```diff
@@ -347,7 +347,7 @@ check_parent <- function(parent, class, call = sys.call(-1L)) {
 #' @rdname new_class
 #' @export
 new_object <- function(`_parent`, ...) {
-  class <- sys.function(-1)
+  class <- sys.function(sys.parent())
```
hadley, line 350: Can you explain why this is needed/what it does?

R/class.R
```diff
@@ -485,6 +485,10 @@ check_prop_overrides <- function(
   call = sys.call(-1L)
 ) {
   overridden <- intersect(names(child_props), names(parent_props))
+  parent_S4 <- if (is_S4_class(parent)) parent else S4_ancestor(parent)
+  if (!is.null(parent_S4)) {
+    check_S4_slot_overrides(child_props, parent_S4, call = call)
```
hadley: Given that `parent_S4` is only used by `check_S4_slot_overrides()`, maybe pass `parent` and then early return if it's an S4 class?

R/constructor.R
hadley, line 34: I think it would be worth refactoring this if statement to make it clear that there are now 3 cases?

R/method-register-S3.R
hadley: Would it be clearer to rename this `envir` instead of creating a `S4_envir`?

R/property.R
```diff
@@ -608,3 +608,7 @@ prop_is_read_only <- function(prop) {
 prop_has_setter <- function(prop) is.function(prop$setter)
 
 prop_is_dynamic <- function(prop) is.function(prop$getter)
+
+prop_is_encapsulated <- function(prop) {
```
hadley, line 612: Nice

tests/testthat/test-class.R
```diff
@@ -116,6 +116,47 @@ test_that("inheritance lets child properties narrow with S4 inheritance", {
...
+
+test_that("inheritance lets S4 children narrow S7 parent properties", {
+  defer(S4_remove_classes(c(
+    "S4PropertyS7Parent",
+    "S4PropertyS7Child"
+  )))
+
+  S4PropertyS7Parent := new_class(package = NULL)
+  S4_register(S4PropertyS7Parent)
+  setClass(
```
hadley, line 142: Use `local_S7_class()` here and elsewhere?

tests/testthat/test-S4.R
```diff
@@ -702,6 +702,104 @@ test_that("S7 classes can extend S4 classes", {
...
+      package = NULL
+    ),
+    "custom getter or setter"
+  )
+
+  S4regAccessorMiddle := new_class(
+    parent = getClass("S4regAccessorParent"),
+    package = NULL
+  )
+  expect_error(
```
hadley, line 787: Might be useful to have at least one snapshot test of this error so you can see the whole thing in one place?

hadley (approved):
Here are a bunch of minor style suggestions; feel free to take as many or as few as you like.

R/class.R
lawremi: The case that exposed this problem is `initialize(new_object())`, which causes `new_object()` to be evaluated when forced by `initialize()`. As you know, this results in a confusing stack trace, where `new_object()` is nested within `initialize()`. That makes `sys.function(-1)` see `initialize()`, while `sys.parent()` returns the actual index of the frame in which `new_object()` is evaluated.

To quote the admittedly vague man page:
```
The parent frame of a function evaluation is the environment in
which the function was called.  It is not necessarily numbered one
less than the frame number of the current evaluation,
```

I guess this is partly why rlang was created :)

hadley: Ah, makes sense!

R/constructor.R
lawremi: Except that the third case depends on the first two cases being resolved, since `new_object_call` is being wrapped.

tests/testthat/test-class.R
lawremi: I'll make a separate PR addressing the S4 class clean up in tests.

PR #722, opened 2026-07-09
alternative approach to S4 test clean up

lawremi:
Instead of calling `local_S4_class()` for each individual class, we could instead just make a single call at the top of a test block that has the side effect of removing all S4 classes assigned in the block. 

```r
{
  local_S4_classes()
  setClass(...)
  setClass(...)
  setOldClass(...)
}
```

Benefits of this approach:

- Allows for conventional S4 syntax for defining classes, including `setClass()`, `setClassUnion()`, and `setOldClass()`, separating the clean up concern from the test logic
- Handles cases like `S4_register()` and the implicit registration that happens with `new_class(parent=S4_class)`.

Downside:

- Far reaching side effect; it will not be obvious from the call site that a class will later be removed.

But for the testing use case, that is unlikely to be problematic.

Thoughts?

### DIFF
--- a/R/S4.R
+++ b/R/S4.R
@@ -656,15 +656,6 @@ find_package_with_symbol <- function(name, env, exclude = NULL) {
   }
 }
 
-S4_remove_classes <- function(classes, where = parent.frame()) {
-  where <- topenv(where)
-  for (class in classes) {
-    if (methods::isClass(class, where = where)) {
-      methods::removeClass(class, where)
-    }
-  }
-}
-
 globalVariables(c(
   ".Data",
   "className",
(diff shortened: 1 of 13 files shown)


hadley (approved):
Ah much better!

(One random thought: should we hack in special support for `setClass()` in `:=`?)

lawremi:
No need for `:=` to support `setClass()`. Very little S4 code assigns the return value of `setClass()`, since class generators were added late in the development cycle.

hadley:
Very little S4 code in the wild might assign, but a high proportion of the S4 code in S7 does 😄

Issue #723, opened 2026-07-10
Constructor performance


t-kalinowski:
While chatting at useR in Warsaw, @bnaras mentioned that the performance of the S7 constructor could be improved for performance-sensitive applications such as CVXR. This also matches our experience in r-xla/anvl.

It would be good for us to optimize the generated constructor and/or `new_object()`. Some big performance gains seem achievable with a little caching. @bnaras was kind enough to share a script (generated with AI assistance) demonstrating this:

```r
## Self-contained reproducible example:
## The `paste` overhead in S7's object constructor, and how CVXR's
## `.fast_new` sidesteps it.
##
## THE PROBLEM
## -----------
## Every `new_object()` call rebuilds the class-dispatch vector via S7's
## internal `class_dispatch()`, which walks the parent chain and calls
##     S7_class_name(x)  ==  paste(c(x@package, x@name), collapse = "::")
## once per level.  For a class N levels deep that is N `paste()` calls on
## EVERY construction -- yet the resulting vector
##     c("pkg::Var", "pkg::Leaf", "pkg::Expr", ..., "S7_object")
## is a CONSTANT of the class.  CVXR builds it once and caches it.
##
## Writeup: notes/s7_paste_overhead.md .  Canonical `.fast_new`:
## CVXR/rsrc_tree/zzz_R_specific/utility.R (ADR D_PERF.1).
##
## Run:  Rscript scripts/s7_paste_overhead.R

stopifnot(requireNamespace("S7", quietly = TRUE),
          requireNamespace("bench", quietly = TRUE))
library(S7)

## ---------------------------------------------------------------------------
## 1. A 5-level hierarchy, mimicking CVXR's Expression -> Leaf -> Variable
## ---------------------------------------------------------------------------
Expr  <- new_class("Expr",  package = "demo",
                   properties = list(id = class_integer))
Node  <- new_class("Node",  parent = Expr,  package = "demo")
Atom  <- new_class("Atom",  parent = Node,  package = "demo")
Leaf  <- new_class("Leaf",  parent = Atom,  package = "demo")
Var   <- new_class("Var",   parent = Leaf,  package = "demo",
                   properties = list(nm = class_character),
                   ## a normal S7 constructor: the hot path we are measuring
                   constructor = function(id, nm) {
                     new_object(S7_object(), id = id, nm = nm)
                   })

## ---------------------------------------------------------------------------
## 2. CVXR-style .fast_new: cache the dispatch vector once per class,
[...]

PR #724, opened 2026-07-10
Consider constructor hot paths

hadley:
* Catch the computed class name and dispatch vector in new `S7_class` and `S7_dispatch` attributes
* Re-implement `class_type()` in C since it's called a lot, and means we can ensure that the first and last types get the same performance (old: 600-3000ns, new 200-250ns)

Part of #723 

This gives a meaningful performance to construction: 
 
| Depth |       main |  Current | Speedup | Reduction |
|------:|-----------:|---------:|:-------:|:---------:|
|     0 |    61.8 µs |  45.6 µs |  1.36×  |    26%    |
|     5 |   457.7 µs | 241.4 µs |  1.90×  |    47%    |
|    10 | 1,085.4 µs | 446.1 µs |  2.43×  |    59%    |
 
Depth 0 is one class directly extending `S7_object`; depths 5 and 10 add that many parent links.

Also helps with `S7_inherits()` and `super()`:

`S7_inherits()` and `super()` were benchmarked using an instance of the deepest class and the root user class as the target.
 
| Depth | Operation       |       main |   Current | Speedup | Reduction |
|------:|-----------------|-----------:|----------:|:-------:|:---------:|
|     0 | `S7_inherits()` |  11.726 µs |  7.872 µs |  1.49×  |    33%    |
|     0 | `super()`       |  26.855 µs | 12.751 µs |  2.11×  |    53%    |


hadley:
@t-kalinowski this doesn't solve all the performance issues, but it's a meaningful improvement.

t-kalinowski (approved):
Very nice!

PR #735, opened 2026-07-28
Use base `attr()` for internal property access

t-kalinowski:
## Summary

S7 now reads class metadata directly with `attr()` in construction and
validation hot paths, avoiding the S3 dispatch overhead of `@`. In the measured
cases, this makes `new_object()` and `validate()` 25-30% faster.

On R 4.3 and later, S7 also no longer defines its compatibility `@` binding, so
remaining internal `@` expressions resolve to base R's generic. On older R
versions, S7 continues to define and export its compatibility implementation,
preserving qualified access to the exported `@` and existing property access
behavior.

## User-facing changes

- Speeds up `new_object()` and `validate()` without changing their APIs.
- Keeps property access behavior unchanged on supported R versions, including
  the exported compatibility `@` on R 4.2 and earlier.

## Internal changes

- Reads known class metadata attributes directly in performance-sensitive
  construction and validation paths.
- Uses exact attribute lookup for `.should_validate`.
- Keeps the conditional namespace export for older R versions and avoids an S7
  `@` binding on R 4.3 and later.
- Adds regression coverage for `validate()` on bare S7 objects and class
  objects.
- Ignores `.claude/worktrees` so local worktrees are not accidentally tracked.

Related to #723.


NEWS.md
```diff
@@ -27,6 +27,8 @@
...
 * `new_class()` generates a working default constructor for a subclass of a class with a custom constructor: the subclass takes `...` and forwards it to the parent constructor (followed by a named argument for each property the subclass adds), so the parent's argument defaults are matched and evaluated by the parent itself. This is a breaking change: such a subclass constructor no longer exposes the parent's properties as named or positional arguments, only via `...` (#609, #317).
 * `new_class()`'s default constructor now respects properties overridden in a subclass: the subclass's default is used (#467) and its setter is run during construction (#585). Values for overridden properties are passed to both the parent constructor and the new object, so a subclass can override a parent property whose default is mandatory.
+* `new_external_class()` creates a delayed reference to an S7 class in another package (or your own package, but not yet defined). It is useful for registering methods on classes from suggested packages (#573) and for creating self-referential or mutually recursive classes (#250).
```
hadley: Oops, shouldn't have tried to do merge resolution on github 😬

hadley:
Looks like this needs one more step to avoid the documentation of `@`

R/class-spec.R
```diff
@@ -212,7 +212,7 @@ class_validate <- function(class, object) {
 
   validator <- switch(
     class_type(class),
-    S7 = class@validator,
+    S7 = attr(class, "validator"), # runs on every construction
```
t-kalinowski: As a general policy, I think in production code like this we should always call `attr()` with `exact=TRUE` (no opinion on if positional or named; I'd pick positional since it's more compact).

R/class.R
```diff
@@ -421,13 +433,15 @@ new_object <- function(`_parent`, ...) {
 
   # Don't need to validate the parent class if it's already validated and none
   # of its properties were reset by this call.
-  parent_validated <- inherits(class@parent, "S7_object") &&
-    !class@parent@abstract
+  parent_validated <- inherits(class_parent, "S7_object") &&
```
t-kalinowski:
```suggestion
  parent_validated <- 
    inherits(class_parent, "S7_object", TRUE) &&
```

t-kalinowski (commented):
Looks great! I agree with the direction.

R/class-spec.R
hadley: This is specifically for attributes on a class that we control, and I added a note to the definition to ensure that it stays safe in the future.

t-kalinowski: Even with that knowledge, I would still add a `, TRUE` to every call.

hadley: Ok, in that case you should finish this PR off and merge it 😄

PR #737, opened 2026-07-29
Only validate base properties once

hadley:
Part of #723. Rebased onto current `main`.

For a property restricted to a base type, `prop_validate()` called `class_inherits()` to check the underlying type and then called the base class validator, which checked exactly the same thing again. This is a reasonable hot path since we expect that many properties will be base classes.

## Benchmarks

I benchmarked current `main` (`91f41404`) and this PR locally back-to-back using `bench/constructor.R`, with all comparable expressions in a single randomized `bench::mark()` call.

| Case | Before | After | Change |
|---|---:|---:|---:|
| 2 base-type properties | 40 µs | 34 µs | 1.2x faster |
| 10 base-type properties | 100 µs | 70 µs | 1.4x faster |
| 50 base-type properties | 400 µs | 250 µs | 1.6x faster |

Hierarchy construction, zero-property construction, and standalone validation were within the ~10% between-session drift threshold.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,6 +83,7 @@ There are three possible ways to run code, listed in rough order of desirability
 - Put all expressions being compared in a single `bench::mark()` call. `bench::mark()` interleaves and randomizes them so that thermal drift, background load, and garbage collection affect the expressions roughly equally.
 - Absolute timings can drift by ~10% between R sessions. So don't pursue changes smaller than this threshold.
 - Always benchmark both pre- and post-timings locally; don't assume timings posted in a GitHub issue are applicable to the current machine.
+- When reporting benchmark results, always include both the absolute pre- and post-timings and the relative change.
 - We care at most about microsecond scale. In the rare cases where they are important, batch 1000s of ns operations to get to around 1µs. On macOS, the timer used by `bench::mark()` has ~40 ns granularity.
 - Use `lobstr::obj_sizes()` to measure marginal memory usage when objects share references. Pass two equivalent objects to `obj_sizes()`: the first result includes the shared graph, while the second includes only the new object's contribution.
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -83,7 +83,7 @@ There are three possible ways to run code, listed in rough order of desirability
 - Put all expressions being compared in a single `bench::mark()` call. `bench::mark()` interleaves and randomizes them so that thermal drift, background load, and garbage collection affect the expressions roughly equally.
 - Absolute timings can drift by ~10% between R sessions. So don't pursue changes smaller than this threshold.
 - Always benchmark both pre- and post-timings locally; don't assume timings posted in a GitHub issue are applicable to the current machine.
-- When reporting benchmark results, always include both the absolute pre- and post-timings and the relative change.
+- When reporting benchmark results, always include both the absolute pre- and post-timings and the relative change as "x faster", not as a percentage reduction in time. Round timings to two significant digits (e.g. 100 µs, not 100.2 µs; 70 µs, not 69.4 µs).
 - We care at most about microsecond scale. In the rare cases where they are important, batch 1000s of ns operations to get to around 1µs. On macOS, the timer used by `bench::mark()` has ~40 ns granularity.
 - Use `lobstr::obj_sizes()` to measure marginal memory usage when objects share references. Pass two equivalent objects to `obj_sizes()`: the first result includes the shared graph, while the second includes only the new object's contribution.
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -57,6 +57,7 @@
 * `set_props()` now names its first argument `_object` to minimise the chances of a clash with a property (#423). It also accepts a single unnamed named list as a shortcut for splicing property values, making it easier to set properties programmatically (#497).
 * `str()` on S7 objects that inherit from data.frame (or other S3 classes whose underlying data has a `dim` attribute incompatible with the bare base type) no longer errors (#494).
 * `super()` now works with S3 and S4 objects, not just S7 objects (#500).
+* `validate()` now checks property types substantially faster, because a property restricted to a base type (e.g. `class_double`) no longer has its underlying type checked twice. Constructing an object with 50 base type properties is 1.6x faster, and with 10 base type properties 1.4x faster (#723).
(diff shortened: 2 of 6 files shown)


t-kalinowski (approved):
I thought it was odd that we added a fast-path in two locations rather than just one, but after some investegation I think it makes sense. 

Some clanker findings:

> <head></head><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Checked out PR 737 at<span class="Apple-converted-space"> </span><code>d8ff8d3d</code><span class="Apple-converted-space"> </span>in detached-HEAD state. The worktree is clean.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Both paths are exercised:</p><div style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><div>
> Operation | validate_properties() | prop_validate()
> -- | -- | --
> Simple constructor | 1 | 0
> Constructor with property validator | 1 | 1
> Explicit validate() | 1 | 0
> Single @<- replacement | 0 | 1
> Batched props<- replacement | 1 | 0
> 
> </div></div><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">The distinction is:</p><ul style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;"><li>[`validate_properties()` (line 154)](/Users/tomasz/.codex/worktrees/ef1d/S7/R/valid.R:154) handles constructors, explicit validation, and batched replacements. Its shortcut avoids calling<span class="Apple-converted-space"> </span><code>prop_validate()</code><span class="Apple-converted-space"> </span>at all for the overwhelmingly common valid base property without a custom validator.</li><li>[`prop_validate()` (line 306)](/Users/tomasz/.codex/worktrees/ef1d/S7/R/property.R:306) is also called directly from the C implementation of single-property replacement in [src/prop.c (line 609)](/Users/tomasz/.codex/worktrees/ef1d/S7/src/prop.c:609). That path subsequently calls<span class="Apple-converted-space"> </span><code>validate(..., properties = FALSE)</code>, so it never reaches<span class="Apple-converted-space"> </span><code>validate_properties()</code>.</li><li>When a base property has its own validator, the outer shortcut cannot apply. Both functions run, but the inner shortcut avoids repeating the base-type check before running the custom validator.</li></ul><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">Your intuition is right semantically: there is one rule implemented twice. Keeping it only in<span class="Apple-converted-space"> </span><code>prop_validate()</code><span class="Apple-converted-space"> </span>would remain correct, but constructors would still pay an R function call per property. Keeping it only in<span class="Apple-converted-space"> </span><code>validate_properties()</code><span class="Apple-converted-space"> </span>would miss single replacements and properties with custom validators. A same-session probe showed both shortcuts were material.</p><p style="caret-color: rgb(255, 255, 255); color: rgb(255, 255, 255); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-line: none; text-decoration-thickness: auto; text-decoration-style: solid;">I’d retain both, but explicitly label the outer one as an inlined fast path kept consistent with<span class="Apple-converted-space"> </span><code>prop_validate()</code>. Focused<span class="Apple-converted-space"> </span><code>valid</code>and<span class="Apple-converted-space"> </span><code>property</code><span class="Apple-converted-space"> </span>tests pass: 153 assertions.</p>

hadley:
I also benchmarked the simpler implementation that always calls `prop_validate()`, then uses `is.null(err)` + `next` to avoid appending `NULL` results.

| Properties | Existing fast path | `is.null()` + `next` | Existing fast path is |
|---|---:|---:|---:|
| 2 | 35 µs | 36 µs | 1.05x faster |
| 10 | 70 µs | 77 µs | 1.1x faster |
| 50 | 260 µs | 290 µs | 1.1x faster |

This recovers much of the cost of simply dropping the special case: at 50 properties the regression falls from roughly 20% to 10%. The 10- and 50-property differences are around the project’s ~10% noise threshold. Given the simpler single source of truth, I’ve switched to the `is.null()` + `next` workflow. Focused valid and property tests pass: 153 assertions.

Issue #607, opened 2026-04-21
S7 objects materialise ALTREP objects


mitchelloharawild:
MRE: 

``` r
library(S7)
myint <- new_class(
  "myint",
  parent = class_integer
)
pryr::object_size(seq_len(1e6))
#> 680 B
pryr::object_size(myint(seq_len(1e6)))
#> 4.01 MB
```

<sup>Created on 2026-04-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I wasn't able to track down exactly why the ALTREP sequences are materialised, but it seems occur when the "S7_object" class attribute is set in the `S7::new_object()` constructor. If I hack together an S7 object using `structure()` rather than `S7::new_object()`, the result is ALTREP safe - although some background process also materialises it (presumably the IDE, since it doesn't happen in R cli).

``` r
library(S7)
myint <- new_class(
  "myint",
  parent = class_integer,
  constructor = function(x) {
    # Hacky simplification of S7::new_object() to 
    # troubleshoot ALTREP materialisation
    if (FALSE) S7::new_object()
    structure(
      x,
      class = c("myint", "integer", "S7_object"),
      S7_class = sys.function(-1L)
    )
  }
)
pryr::object_size(z <- myint(seq_len(1e6)))
#> 1.28 kB
pryr::object_size(z)
#> 1.28 kB (4.00MB in Positron IDE)
```

<sup>Created on 2026-04-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.5.2 (2025-10-31)
#>  os       KDE neon User Edition
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language (EN)
#>  collate  en_US.UTF-8
#>  ctype    en_US.UTF-8
#>  tz       Australia/Melbourne
#>  date     2026-04-21
#>  pandoc   3.8.3 @ /usr/share/positron/resources/app/quarto/bin/tools/x86_64/ (via rmarkdown)
#>  quarto   1.8.26 @ /opt/quarto/bin/quarto
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
[...]

PR #738, opened 2026-07-29
Cut fixed bookkeeping cost in `new_object()`

hadley:
Part of #723. Rebased onto current `main`. The `@` → `attr()` metadata reads this originally carried have been dropped — those now belong to #735. What remains are reads of the *new* cache attributes this PR introduces (`S7_setter_props`, `S7_dispatch`), which have no `@` equivalent.

## What

`new_object()` runs on every construction, once per level of inheritance, so its fixed costs are paid over and over.

* **Cache the names of properties with custom setters** in a new `S7_setter_props` attribute, replacing `vlapply(class@properties[names(args)], prop_has_setter)` on every construction (2.54 µs → 0.37 µs). Always a character vector, so a missing attribute unambiguously means "class built by an older S7", and there's a fallback for that case with a test.

* **Fast paths for the two common shapes**: no properties supplied at all, and no supplied property having a custom setter. Both skip the setter split and the setter loop; the former also skips the storage rename.

* **Reuse the class's cached `S7_dispatch` vector** instead of rebuilding it on every construction.

* **A fast path in `check_parent()`** for what generated constructors always produce: an S7 object whose class is exactly the parent class (4.10 µs → 1.3 µs).

* **`collect_dots()` returns early** when `...` is empty (1.07 µs → 0.25 µs).

## Numbers

| | before | after |
|---|---|---|
| no properties | 37 µs | **27 µs** (1.4x) |
| depth1 | 54 µs | **45 µs** (1.2x) |
| custom constructor | 46 µs | 38 µs |
| depth10 | 785 µs | 668 µs |

## A landmine worth knowing about

`check_parent()` **cannot call `class()`**. Its `class` argument is bound to an S7 class object, which is itself a function, so R's function-position lookup finds the argument and `class(parent)` recurses forever — blowing the C stack *during package load*. This uses `oldClass()`; there's a comment saying so.

## Worth a look during review

The `check_parent()` fast path is the change I'd most want a second pair of eyes on. It forces `_parent` marginally earlier than before, ahead of the "class has no parent" and abstract-parent early returns. It can only ever return early on an *exact* class match, and I confirmed a forged `structure(1, class = c("Base", "S7_object"))` is still rejected (no `_S7_class` attribute), so it doesn't loosen the check.

## Not done, deliberately

`sys.function()` is not worth replacing for speed, but it is worth fixing for **memory**: `R_sysfunction()` returns `duplicate(cptr->callfun)` (`src/main/context.c:518`, complete with R Core's own `/***** do we need to DUP? */`), and `duplicate()` deep-copies attributes. Since S7 stores class metadata as closure attributes, every construction deep-copies the class and its whole ancestor chain — so a depth-5 instance costs ~5400 bytes against ~40 for a plain equivalent. That's the subject of the draft #740, not this PR.

`local_constructing()` is the largest remaining per-level item at 2.1 µs, 1.2 µs of which is `defer()`'s `do.call(on.exit, ..., envir = frame)`. Making it conditional needs a cached "has an abstract ancestor" flag, which isn't sound for external classes resolved at run time, so I left it.

## Correctness

`devtools::test()`: `FAIL 3 | PASS 1240` against `main`'s `FAIL 3 | PASS 1239` — same pre-existing environmental failures (fixture package installs; `R CMD check` reports `checking tests ... OK`). No snapshot changes. The #607 ALTREP test (which byte-compiles `new_object` via `cmpfun`) passes; no local binding of `_parent` was introduced.

Also checked by hand: reserved-name properties still stored as `_names`/`_dim`, inherited setters still run, duplicate names in a spliced list still resolve first-wins, abstract classes still error directly and still work as parents, read-only and dynamic properties fine, and a genuine subclass instance as `_parent` is still accepted.

## Relationship to the other PRs

#736, #737 and #739 are the siblings; all four are rebased onto `main` and independent. This PR and #739 both add an attribute in `new_class()`'s attribute block, so whichever merges second needs a one-line rebase and to add both names to the `prop_names` synchronisation test. Note also that #735 documents a constraint on that block: no metadata attribute name may be a prefix of another, since the hot-path reads there omit `exact = TRUE`. `S7_setter_props` and `S7_dispatch` both satisfy it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hadley:
This seems pretty marginal so I'm not going to implement it.

PR #739, opened 2026-07-29
Cache each class's hierarchy of validators

hadley:
Part of #723. Rebased onto current `main`. The `@` → `attr()` reads of `validator` and `parent` this originally carried have been dropped — those now belong to #735. What remains are reads of the *new* `S7_validators` cache attribute, which has no `@` equivalent.

Of the four sibling PRs this is the one with the most design in it, and worth the closest review. It's the big win for deep hierarchies.

## What

Validating an object walked up the class hierarchy calling `class_validate()` at every level, paying for `is_S4_class()` and a `class_type()` switch each time, even though almost no hierarchy has any validators at all.

Each class now caches, in a new `S7_validators` attribute, every validator that must run for one of its instances — ordered from the class itself up to the root, named with the owning class's description. It's built incrementally from the parent's cache, so maintaining it costs nothing.

The cache is `NULL`, meaning "walk the hierarchy", when:

* the ancestry contains an S4 or external class, whose validators can only be resolved at validation time; or
* the class was built by an older version of S7 (no attribute).

It's only consulted when `parent` is `NULL`, i.e. when the whole hierarchy is being validated. That's deliberate: the cache describes the *full* ancestry, not the truncated walk that `validate_from(parent = )` performs, so `recursive = FALSE`, the S4-ancestor path, and constructors with an already-validated parent all still walk.

## Numbers

I benchmarked current `main` (`91f41404`) and this PR locally back-to-back using `bench/constructor.R`, with all comparable expressions in a single randomized `bench::mark()` call.

| Case | Before | After | Change |
|---|---:|---:|---:|
| `validate()` on a depth-five object | 31 µs | 21 µs | 1.5x faster |
| One-shot depth-five construction | 32 µs | 20 µs | 1.6x faster |

Generated constructor chains, wide classes, and the other benchmark cases were within the ~10% between-session drift threshold. This is expected: the cache is used when validating the full hierarchy, while generated constructors validate incrementally against an already-validated parent.

## Why this caches closures, not classes

My first version cached the *class objects* that own validators. It worked and sped up the deep cases, but it **regressed** `wide10` (113 → 120 µs) and `wide50` (418 → 438 µs).

The reason is that `R_sysfunction()` returns `duplicate(cptr->callfun)` (`src/main/context.c:518`, with R Core's own `/***** do we need to DUP? */` comment), and `duplicate()` deep-copies attributes. Since S7 keeps class metadata in closure attributes, every construction deep-copies the class and its whole ancestor chain — so adding an attribute that referenced ancestors made that copy O(depth²).

Caching the validator closures instead removed the regression, because duplicating a closure is cheap: formals, body and environment are shared, and a bare validator has no attributes. (The underlying duplication is addressed separately in draft #740.)

## Also in this PR

If a validator returns something other than `NULL` or a character vector, the error now names the class that owns it. On `main` you get the useless `<S7_class> validator must return NULL or a character, not <double>.`; now it's `<Foo> validator must ...`. Same text from both the cached and walked paths. No existing snapshot depended on the old message.

## Correctness

The cache reads the parent's cache at `new_class()` time, so a class object mutated after creation (e.g. `attr(Foo, "validator") <- ...`) would go stale — already unsupported, and the old walk had the same exposure through `@parent`.

I did *not* try to cheapen the `identical()` in the fallback walk. Comparing `S7_class_name` first would be faster but changes behaviour when two classes in one chain share a fully-qualified name.

New tests cover: every ancestor's validator running for a subclass defined *after* its ancestors (including a base-type root); the invalid-return message from both the cached and the walked path; and an *external* parent's validator still running, which guards the `NULL`-cache fallback. Verified `R CMD INSTALL` works — the `S7_object` bootstrap at build time, where `class_type()`'s DLL isn't available yet, is guarded by the `is.null(parent)` branch.

`devtools::test()`: `FAIL 0 | WARN 0 | SKIP 0 | PASS 1271`. Snapshots: additions only, no existing snapshot changed.

## Relationship to the other PRs

#736, #737 and #738 are the siblings; all four are rebased onto `main` and independent. This PR and #738 both add an attribute in `new_class()`'s attribute block, so whichever merges second needs a one-line rebase and to add both names to the `prop_names` synchronisation test. Note #735 documents a constraint on that block: no metadata attribute name may be a prefix of another, since the hot-path reads there omit `exact = TRUE`. `S7_validators` satisfies it — but it is *close* to the existing `validator`, so it's worth keeping in mind that the reverse direction (a new attribute named `validator...` shorter than an existing one) is the hazard.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -57,7 +57,9 @@
 * `set_props()` now names its first argument `_object` to minimise the chances of a clash with a property (#423). It also accepts a single unnamed named list as a shortcut for splicing property values, making it easier to set properties programmatically (#497).
 * `str()` on S7 objects that inherit from data.frame (or other S3 classes whose underlying data has a `dim` attribute incompatible with the bare base type) no longer errors (#494).
 * `super()` now works with S3 and S4 objects, not just S7 objects (#500).
+* `validate()` is 1.7x faster because each class now caches the validators of its whole hierarchy, so validating an object no longer walks it. Constructing an object of a class with 10 ancestors is 1.5x faster (#723).
 * `validate()` now signals validation errors with class `S7_error_validation_failed`, so they can be caught with `tryCatch()` (#602, #605).
+* `validate()` now names the class that owns the validator when a validator returns something other than `NULL` or a character vector, instead of the uninformative `<S7_class>` (#723).
 
 # S7 0.2.2
 
--- a/R/class.R
+++ b/R/class.R
@@ -191,6 +191,11 @@ new_class <- function(
   class_name <- paste(c(package, name), collapse = "::")
   attr(object, "S7_class_name") <- class_name
   attr(object, "S7_dispatch") <- S7_class_dispatch(class_name, parent_resolved)
+  attr(object, "S7_validators") <- S7_class_validators(
+    validator,
+    class_name,
+    parent
+  )
   class(object) <- c("S7_class", "S7_object")
 
   if (S7_extends_S4(object)) {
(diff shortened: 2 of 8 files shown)


hadley:
This no longer seems worth it.

PR #743, opened 2026-07-29
Reduce copies of S7 class objects

### LINKED ISSUE #742: Think about how to store the class object
Unfortunately `sys.function()` makes a deep copy of the class, meaning that we make a lot of unnecessary duplicates with both increases memory consumption and makes various `identical()` comparisons slower.

```R
library(S7)
library(lobstr)

compare_addresses <- function(...) {
  expressions <- as.list(substitute(list(...)))[-1]
  x <- list(...)
  addresses <- obj_addrs(x)

  data.frame(
    expression = vapply(expressions, deparse1, character(1)),
    address = addresses,
    same_as_first = addresses == addresses[[1]],
    row.names = NULL
  )
}

# Constructing instances ---------------------------------------------------

Foo := new_class(properties = list(x = class_double), package = NULL)
x <- Foo(x = 1)
y <- Foo(x = 2)

# On main, each instance contains a duplicate of Foo.
compare_addresses(
  Foo,
  S7_class(x),
  S7_class(y)
)

# Serializing a class ------------------------------------------------------

class_path <- tempfile(fileext = ".rds")
saveRDS(Foo, class_path)
Foo_rds <- readRDS(class_path)
x_rds <- Foo_rds(x = 1)
y_rds <- Foo_rds(x = 2)

compare_addresses(
  Foo_rds,
  S7_class(x_rds),
  S7_class(y_rds)
)

# The restored generator has a new address, as expected.
compare_addresses(Foo, Foo_rds)

# Serializing instances ---------------------------------------------------

instances_path <- tempfile(fileext = ".rds")
saveRDS(list(x, y), instances_path)
xy_rds <- readRDS(instances_path)

# Restored instances contain separate copies of the class.
# R's serializer preserves sharing for environments, but not closures.
compare_addresses(
  S7_class(xy_rds[[1]]),
  S7_class(xy_rds[[2]])
)

# Installing and loading a package ----------------------------------------

pkg <- file.path(tempdir(), "s7address")
dir.create(file.path(pkg, "R"), recursive = TRUE)

writeLines(
  c(
    "Package: s7address",
    "Version: 0.0.0.9000",
    "Title: S7 Address Reprex",
    "Description: A temporary package used to inspect S7 class addresses.",
    "License: MIT",
    paste0(
      "Authors@R: person(\"A\", \"Person\", email = \"a@example.com\", ",
      "role = c(\"aut\", \"cre\"))"
    ),
    "Imports: S7"
  ),
  file.path(pkg, "DESCRIPTION")
)
writeLines(
  c(
    "export(Foo)",
    "importFrom(S7,\":=\")",
    "importFrom(S7,class_double)",
    "importFrom(S7,new_class)"
  ),
  file.path(pkg, "NAMESPACE")
)
writeLines(
  "Foo := new_class(properties = list(x = class_double))",
  file.path(pkg, "R", "classes.R")
)

lib <- file.path(tempdir(), "s7address-lib")
dir.create(lib)
install.packages(
  pkg,
  lib = lib,
  repos = NULL,
  type = "source",
  quiet = TRUE,
  INSTALL_opts = c("--no-byte-compile", "--no-test-load")
)

invisible(loadNamespace("s7address", lib.loc = lib))
pkg_Foo <- getExportedValue("s7address", "Foo")
pkg_x <- pkg_Foo(x = 1)
pkg_y <- pkg_Foo(x = 2)

compare_addresses(
  pkg_Foo,
  S7_class(pkg_x),
  S7_class(pkg_y)
)
```

hadley:
Fixes #742.

This prototype stores an environment-backed class reference on ordinary S7 instances instead of the class closure itself. This avoids the deep copy from `sys.function()` and ensures `saveRDS()` saves a single copy of the object.

The tests use a small internal C `obj_addr()` helper to assert reference identity directly.


hadley:
Todo:

* Extract out helpers for both creating and retrieving so we can document in one place
* Benchmark the performance impact of the changes
* Benchmark the size impact of the changes

hadley:
I benchmarked this locally against `main` (R 4.5.3, aarch64 macOS; bench 1.1.4, lobstr 1.2.1). I ran `bench/constructor.R` twice in alternating order and averaged the two reported medians. I also corrected the memory case locally so it creates one class and passes two instances of that class to `lobstr::obj_sizes()`; previously it measured two separately created class definitions.

Marginal retained memory per instance:

| hierarchy depth | main | PR | reduction |
|---:|---:|---:|---:|
| 1 | 3,048 B | 168 B | 94.5% |
| 5 | 7,208 B | 168 B | 97.7% |
| 10 | 12,720 B | 168 B | 98.7% |
| 20 | 24,120 B | 168 B | 99.3% |

So the PR achieves the intended result: per-instance class storage becomes a flat 168 B rather than growing with the class hierarchy.

Construction timings (average of the two median runs):

| case | main | PR | change |
|---|---:|---:|---:|
| depth1 | 24.6 µs | 27.0 µs | +10% |
| depth5 | 119.7 µs | 135.1 µs | +13% |
| depth10 | 238.2 µs | 283.1 µs | +19% |
| oneshot5 | 31.9 µs | 36.6 µs | +15% |
| wide0 | 23.7 µs | 26.7 µs | +13% |
| wide2 | 40.7 µs | 43.0 µs | +6% |
| wide10 | 102.2 µs | 102.2 µs | 0% |
| wide50 | 413.4 µs | 402.6 µs | -3% |
| setter | 41.9 µs | 44.4 µs | +6% |
| custom | 33.1 µs | 35.5 µs | +7% |

`validate5` was effectively unchanged (+1%). The timing cost is most visible for lightweight/deep constructors, where the extra class-reference lookup is a larger share of total work; it disappears as property work dominates.

hadley:
Follow-up class-operation benchmarks, now in a separate `classes` benchmark section (depth-5 class; same local setup as above):

| case | main | PR | change |
|---|---:|---:|---:|
| retrieve with `S7_class()` | 1.4 µs | 1.6 µs | +14% |
| `identical()` on classes from two instances | 17.5 µs | 0.5 µs | -97% |
| `class_extends()` on classes from two instances | 16.5 µs | 0.7 µs | -96% |

This confirms the expected downstream benefit of shared class identity: operations whose first step is an identity comparison become roughly 24–35× faster. Direct class retrieval is slightly slower because the C accessor must dereference the environment-backed class reference.

hadley:
I also prototyped having S7-generated constructors pass the class reference explicitly to `new_object()`, bypassing the dynamic `get0()` lookup. I ran three alternating baseline/prototype sessions and averaged the reported medians:

| case | dynamic lookup | explicit reference | change |
|---|---:|---:|---:|
| depth 1 | 27.3 µs | 26.7 µs | -2.2% |
| depth 5 | 138.1 µs | 134.7 µs | -2.4% |
| depth 10 | 286.0 µs | 280.1 µs | -2.1% |
| one-shot depth 5 | 36.3 µs | 35.4 µs | -2.6% |
| custom | 36.0 µs | 35.9 µs | -0.3% |

So explicit passing saves about 0.6 µs per `new_object()` call, consistent with the isolated lookup benchmark, but it does not materially move whole-constructor performance (~2%, below our 10% threshold). I reverted the prototype rather than complicating generated calls and the `new_object()` interface.

R/class.R
```diff
@@ -514,6 +530,33 @@ S7_class <- function(object) {
...
+# Class objects are closures, which leads to two problems:
+# * `sys.function()` does deep copies
+# * `serialize()`/`saveRDS()` only de-dups environments
+# We solve both problems with an environment-backed class reference. The
+# reference is bound as `.S7_class_ref` in the constructor's environment and
+# points back to the completed class through `$class`. Ordinary S7 objects store
+# the reference instead of the closure, avoiding `sys.function()` and ensuring
+# that objects serialized together share a single copy of their class.
+new_class_ref <- function() {
+  ref <- new.env(parent = emptyenv())
```
t-kalinowski, line 546: My guess is that if we set `hash = FALSE` here we'll see slightly better performance. (probably above, in the `constructor_env`, too)

tests/testthat/test-class.R
```diff
@@ -363,6 +363,87 @@ test_that("new_object() gives useful error if called directly", {
   expect_snapshot(new_object(), error = TRUE)
 })
 
+test_that("new_object() stores a shared class reference (#742)", {
+  Foo := new_class(package = NULL)
+  x <- Foo()
+  y <- Foo()
+
+  x_ref <- attr(x, "_S7_class", exact = TRUE)
```
t-kalinowski, line 371: Do we have a principled policy for when to use the underscore prefix and when to use the dot prefix?

t-kalinowski (approved):

hadley:
I set `hash = FALSE` on both `new.env()` calls introduced by this branch and benchmarked the change back-to-back.

- Class definition: 73 µs → 72 µs, effectively unchanged.
- Allocation per class definition: 8.41 KB → 7.86 KB, about 1.07x less.
- Constructor, class lookup, and instance memory benchmarks were unchanged within noise; all differences were ≤3%.

Focused `class` and `constructor` tests pass: 163 assertions.

t-kalinowski (approved):

PR #744, opened 2026-07-29
Add constructor benchmarks

hadley:
Adds a reusable benchmark script for constructor performance and per-instance memory usage.

The script:
- compares hierarchy depth, property width, setters, custom constructors, validation, dispatch, inheritance, and property access in one interleaved `bench::mark()` call
- measures marginal instance memory with `lobstr::obj_sizes()` so shared class graphs are counted only once
- supports selecting benchmark groups, saving results, and comparing back-to-back runs

It also documents the project benchmark methodology in `AGENTS.md` and excludes `bench/` from package builds.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -24,3 +24,4 @@
 ^\.git-blame-ignore-revs$
 ^AGENTS\.md$
 ^\.lintr\.R$
+^bench$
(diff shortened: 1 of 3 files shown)


hadley:
@t-kalinowski This is heavily AI generated, so no need to review too closely. But I think it's useful to establish some best practices, since claude/codex doesn't seem to have great intuition here.

t-kalinowski (approved):

PR #753, opened 2026-08-26
Restore compatibility with release ggplot2

### LINKED ISSUE #747: Dev version of S7 doesn't work with release version of ggplot2 today
```
> library(ggplot2)
Error: package or namespace load failed for ‘ggplot2’:
 .onLoad failed in loadNamespace() for 'ggplot2', details:
  call: S7::new_object(elements, complete = complete, validate = validate)
  error: `_parent` must be an instance of S3<gg>, not <list>.
```

t-kalinowski:
Fixes #747.

Release ggplot2 packages can again load with the development version of S7. Their stored `S3<gg>` class definitions were created before S7 recorded explicit abstract metadata or tagged generated constructors, so S7 treated the abstract parent as concrete and rejected the list used to construct a theme.

Internally, this restores the legacy stub-constructor check only for S3 class definitions that lack explicit abstract metadata. Current definitions continue to use the metadata and constructor tag, and concrete S3 parents still receive the usual parent validation. The regression fixture reproduces the S7 0.2.2 class shape.

## Testing

- `devtools::test()` on R 4.4.3: 1,288 tests passed.
- `devtools::check()` on R 4.4.3: 0 errors, 0 warnings, and 4 unrelated notes.
- Loaded ggplot2 4.0.0 on R 4.4.3 and ggplot2 4.0.3 on R 4.6.1, then constructed `theme()` with each.


R/class-spec.R
```diff
@@ -431,6 +433,29 @@ obj_dispatch <- function(x) {
 
 # helpers -----------------------------------------------------------------
 
+# Does `child`'s S3 dispatch inherit from `parent`'s? S3 systems may prepend
```
hadley: I'm surprised that this doesn't explicitly mention the backwards compatibility issue.

R/S3.R
```diff
@@ -148,6 +148,21 @@ is_S3_class <- function(x) {
   inherits(x, "S7_S3_class")
 }
 
+# Detect the stub constructor that `new_S3_class()` inserted before abstract
+# S3 classes had explicit metadata. Needed for class definitions serialized by
+# older versions of S7.
+is_S3_stub_constructor <- function(constructor) {
```
hadley, line 156: I wonder if it would be better to add an explicit `_version` attribute to classes so we could make it clear when we're adding backward compatibility shims?

hadley (commented):
Since this is at least the second time we've accidentally broken ggplot2, maybe it's worth a new GitHub action that tests ggplot2 explicitly? (I think it's fine to do just on CI, since a test will be fiddly/expensive and required ggplot2 as a suggested dep)

R/class-spec.R
t-kalinowski: Added a comment mentioning the backwards compatibility motivation. This should make it easier to remove in the future.

R/S3.R
t-kalinowski: Added a `_version` attribute, starting at `1L`. I used an integer representation version rather than the S7 package version because it only needs to change when the representation changes. Older class definitions remain unversioned, and the backwards compatibility code only runs for those definitions.

.github/workflows/R-CMD-check.yaml
```diff
@@ -55,3 +55,45 @@ jobs:
         with:
           upload-snapshots: true
           build_args: 'c("--no-manual","--compact-vignettes=gs+qpdf")'
+
+  ggplot2:
```
hadley: I'd put this in a completely separate job so it doesn't get clobbered when you run `use_tidy_github_actions()`

t-kalinowski: Moved the ggplot2 compatibility check to its own `test-ggplot2.yaml` workflow, so `use_tidy_github_actions()` can update `R-CMD-check.yaml` without removing it.

R/class.R
```diff
@@ -335,7 +335,9 @@ class_is_abstract <- function(class) {
   if (is_class(class)) {
     attr(class, "abstract", TRUE) # called on construction
   } else if (is_S3_class(class)) {
-    class$abstract %||% is_default_constructor(class$constructor)
+    class$abstract %||%
+      (is.null(attr(class, "_version", exact = TRUE)) &&
```
hadley: I wonder if the fallback should live in `is_default_constructor()` for clarity?

R/class-spec.R
```diff
@@ -326,10 +326,10 @@ class_inherits <- function(x, what) {
     missing = FALSE,
     any = TRUE,
     S4 = methods::is(x, what),
-    S7 = has_S7_class(x) && inherits(x, S7_class_name(what)),
+    S7 = inherits(x, "S7_object") && inherits(x, S7_class_name(what)),
```
hadley, line 330: Could this fallback live in `has_S7_class()` or is that too expensive?

R/class-spec.R
```diff
@@ -431,6 +433,34 @@ obj_dispatch <- function(x) {
...
+#
+# ggplot2 4.0.x relies on the S7 0.2.2 behavior where an S3 class
+# specification can match before shared trailing classes (#747), e.g. `"Coord"`
+# in c("CoordCartesian", "Coord", "ggproto", "gg"). Preserve that behavior for
+# backward compatibility, while requiring multi-class specifications to be
+# contiguous and ordered. `class_dispatch_extends()` remains tail-only for
+# downcasts.
+# S7 wrappers of base/S3 types append "S7_object", which we ignore.
+class_dispatch_inherits <- function(parent, child) {
+  parent <- drop_S7_object(parent)
```
hadley, line 459: Only use this path if `version` is NULL?

hadley (approved):

